PR #9, opened 2011-11-28
Reverted setup.py to not using metadata.py, Added code to set/get Selenium's implicit wait functionality

emanlove:
These commits apply two changes, reverting the setup.py and adding the implicitly_wait functionality found in Selenium 2 (3c91af5d).


j1z0:
I agree with removing the third party libs and the Implicit wait change is the same thing I've got on my system so I'm going to go ahead and merge this request.  Ryan you can undo if you have issues with it.

PR #16, opened 2011-12-12
Temporary fix for missing VERSION string after removing metadata.py.

emanlove:
This should be the last fix in moving away from static libraries and refactoring setup.py.  Realized that the version string was needed elsewhere and removing metadata.py broke the package.


emanlove:
Forgot to mention, I called this "temporary" in that to fix this issue of now missing VERSION number I simply added it to the file.  But this means that it is located in two places.  I have no problem with this but others might see differently and want a better solution (like the one in the [decorator package](http://code.google.com/p/micheles/source/browse/decorator/setup.py)).

PR #27, opened 2011-12-28
Textfield value should be -keyword fails, doesn't find the element

laurilarjo:
"Textfield Value Should Be" -keyword did not find the textfield, but just failed with:

FAIL Value of text field 'edit-mail' should have been 'acceptance-user@example.com' but was 'None'

This small change fixes the issue, but is probably not the optimal solution.

### DIFF
--- a/src/Selenium2Library/keywords/_formelement.py
+++ b/src/Selenium2Library/keywords/_formelement.py
@@ -243,7 +243,7 @@ class _FormElementKeywords(KeywordGroup):
         Key attributes for text fields are `id` and `name`. See `introduction`
         for details about locating elements.
         """
-        element = self._element_find(locator, True, False, 'text field')
+        element = self._element_find(locator, True, False, 'input')
         if element is None: element = self._element_find(locator, True, False, 'file upload')
         actual = element.get_attribute('value') if element is not None else None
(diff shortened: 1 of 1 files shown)


emanlove:
Lauri, although your test passed with this change I am not sure this change is a good one or even valid.  The tag parameter which you changed is passed along to the [find method](https://github.com/rtomac/robotframework-selenium2library/blob/master/src/Selenium2Library/locators/elementfinder.py#L16) within the ElementFinder class. And the find method in turn passes this value to [ElementFinder._get_tag_and_constraints()](https://github.com/rtomac/robotframework-selenium2library/blob/master/src/Selenium2Library/locators/elementfinder.py#L106).  This method then compares the "tag" value to several constants of which 'input' is not one of the valid choices but 'text area' is. I suspect that when you changed this value to 'input' the locator relaxed it's constraints and thus "found" the element you were looking for. Probably the parameter name "tag" is a mislabeled and the _get_tag_and_constraints method should error if the "tag" is not one listed in the elif structure but as far as I can tell this change is incorrect.

Do you see this differently?

j1z0:
Lauri,

You probably want the function element_text_should_be.

PR #28, opened 2012-01-04
Implicit wait

j1z0:
OK this is my stab at the implicit_wait issue.  Here is my thinking about it.
1. implicit_wait will make selenium failures take longer so by default it should be turned off:  implicit_wait=0.0
2. When s2l is initially created an implicit_wait value can be set that will then be used by each new browser that is created.  This is largely a convenience function.
3.  When creating a new browser i.e. calling open_browser there is now an optional argument `implicit_wait` if this is specified then new browser will be created with whatever implicit wait value is passed.  If not the browser will use the s2l default implicit_wait time set during library creation.
4.  set_selenium_implicit_wait will now update the default implicit_wait value and it will overwrite the implicit wait value for all currently open browsers.
5. get_selenium_implicit_wait will return the default implicit_wait value.  There is a potential issue here in that if you call open_browser pass in an implicit_wait time and then call get_selenium_implicit_wait the 2 values might not match.  The only work around I can see to this is to extend WebDriver it self to store / return the currently set implicit wait time but I don't see a real need for it at this point so I'm leaving it alone.


emanlove:
> Here is my thinking about it.
> 
>   1) implicit_wait will make selenium failures take longer so by default it should be turned off: implicit_wait=0.0

This should default to zero as this is also a functionality change which could catch people by surprise so I totaly agree here.

>   2) When s2l is initially created an implicit_wait value can be set that will then be used by each new browser that is created. This is largely a convenience function.

Sounds good.

>   3) When creating a new browser i.e. calling open_browser there is now an optional argument implicit_wait if this is specified then new browser will be created with whatever implicit wait value is passed. If not the browser will use the s2l default implicit_wait time set during library creation.

Will people get confused or forgetful as to the value of implicit wait if they don't specifically set the value at the point of calling open_browser? Can the value of implicit wait be set in one location and call open_browser in another location within a test suite thus causing hard to track down "bugs" for people?

>   4) set_selenium_implicit_wait will now update the default implicit_wait value and it will overwrite the implicit wait value for all currently open browsers.

The idea that updating the value will change all open browsers is a little worrisome. That is I could see someone setting specific implicit waits on some browsers and then use set_selenium_implicit_wait "accidentaly" or not realizing they modified certain browsers. It's just one of those things I could see tripping up people if they don't carefully read the documentation.

The other issue here is that one can not reset the implicit_wait on an individual already opened browser which seems like a loss of functionality.

>   5) get_selenium_implicit_wait will return the default implicit_wait value. There is a potential issue here in that if you call open_browser pass in an implicit_wait time and then call get_selenium_implicit_wait the 2 values might not match. The only work around I can see to this is to extend WebDriver it self to store / return the currently set implicit wait time but I don't see a real need for it at this point so I'm leaving it alone.

Agree.

I wonder with 3 and 4 if we are giving people too much room for getting into trouble. I also question whether we should have two functions for setting implicit wait; one for the current browser and a second for all browsers. Or just one that sets the value for the current browser and allows uses to use FOR loops within their tests to set on multiple browsers (although I don't see alot of keywords in order to get information on multiple browsers, ie get_number_of_browsers, get_browser_list, etc which would make this possible)

j1z0:
Ed,
Thanks for the feedback. I see two options to avoid confusion. 

1 set implicit wait function only affects the current browser and does not affect the default implicit wait. I.e.  default cAn only ever be set at library creation All other changes in the branch remain the same. This I think removes the confusion u are referring to for point 3.  Alternatively we can get ride of default altogether. 
1.  Two functions like u said is simple enough to implement and would provide a more "batteries included" approach for those that need it so I'm happy to do that. 

What say u?

Cheers,
Jeremy

emanlove:
Jeremy,

I think a second function for setting the browser's implicit wait would work.  Something like

``` python
    def set_browser_implicit_wait(self, seconds):
        """Sets current browser's implicit wait in seconds.

        From selenium 2 function 'Sets a sticky timeout to implicitly 
            wait for an element to be found, or a command to complete.
            This method only needs to be called one time per session.'

        Example:
        | Set Browser Implicit Wait | 10 seconds |

        See also `Set Selenium Implicit Wait`.
        """
        _implicit_wait_in_secs = robot.utils.timestr_to_secs(seconds)
        self._current_browser().implicitly_wait(_implicit_wait_in_secs)
```

Just as you mentioned, I'm avoiding the complexity of adding a member variable along with get/set functions to the BrowserCache class.  So you can only set the implicit wait on a browser and not retrieve the set value.

Also with this change we should change the doc on set_selenium_implicit_wait to indicate it sets the value on all (open?) browsers.

Ed

emanlove:
Jeremy,

The code in my previous commit was just an example.  I will let you add in a "set_browser_implicit_wait" function if you wish into your pull request.

Ed

PR #31, opened 2012-01-09
Use Selenium's select class for index-based option selecting. This seems...

hhakkala:
... to help with performance issues with selecting elements from large singe-select lists.


j1z0:
I agree with you that the select class will speed things up, but your change breaks the acceptance tests.... to properly use the Select class you probably need to refactor the entire _selectelement class.  Otherwise it will be quite confusing as to why some functions use the Select class and others don't, plus you won't really get the speed advantages either.

So if anybody wants to volunteer to do this.. that would be awesome. :)

j1z0:
Heikki,

This issue has been fixed by Ed's latest commit.  So I'm closing this request.  

Thanks,
Jeremy

PR #37, opened 2012-03-04
Fix select an item in a single selection list

emanlove:
This is a quick fix for the failing acceptance.keywords.list test, 'Select From Single Selection List'. While selecting an item on a single selection list, Selenium 2 no longer accepts clicking on the option.  I am using the Select utility class from selenium to select the item by index.

### DIFF
--- a/src/Selenium2Library/keywords/_selectelement.py
+++ b/src/Selenium2Library/keywords/_selectelement.py
@@ -1,4 +1,5 @@
 from selenium.webdriver.remote.webelement import WebElement
+from selenium.webdriver.support.ui import Select
 from keywordgroup import KeywordGroup
 
 class _SelectElementKeywords(KeywordGroup):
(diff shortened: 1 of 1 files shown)


emanlove:
Sorry about placing this in the original repository; I menat to have it in mine fork.  I will let either of you merge this in as a good code review process.  - Ed

j1z0:
Up to you Ed, if you just want to put in this quick fix to fix the failing test then I'm OK with it  If you do have time to redo the rest of the _selectelement.py to use Select that would also be cool :)

Jeremy

emanlove:
Sorry, should have checked my email, first.  I'm for putting in the 
quick fix so we can get out the release.  Then setup a more complete 
change over to the Select class for another point release.

Ed

j1z0:
I agree man.  You can merge the pull request now right?  Or do you need me to do it?

Cheers,
Jeremy

emanlove:
Merged it this morning (last night Malaysia time, I think).  One down 
and five more to go!

Ed

PR #38, opened 2012-03-05
Fix issue with Acceptance.Windows test

emanlove:
Using Suite Teardown within Acceptance.Open and Close test suite, which comes right before the Acceptance.Windows test, and calling "Close All Browsers" we can resolve the failing windows test. In investigating the cause of failure, the tests passed if ran isolated (using the command "$ python run_tests.py python *ff --suite='Acceptance.Windows'"). Further investigation found the commenting out all the "Close Browser" keywords within the Open and Close test would also fix the issue (but expectantly cause some of those test to fail). It is unclear why calling "Close All Browsers" works while the less inclusive "Close Browser" fails or at least cause an issue latter on.

Although this fixes the problem I am troubled by the test isolation issue and I am still unsure why exactly this fixes the problem.

### DIFF
--- a/test/acceptance/open_and_close.txt
+++ b/test/acceptance/open_and_close.txt
@@ -1,5 +1,6 @@
 *** Settings ***
 Resource        resource.txt
+Suite Teardown  Close All Browsers
 
 *** Test Cases ***
 Browser Should Open And Close


j1z0:
This is what I know so far:

If you call the test without the Close All browsers Suite Teardown.

in Open and Close Browser Suite

browser is:
http://127.0.0.1:52422/hub/session/5b81cdcc-4126-654a-a486-5e7e8d956a4b/url

in windows suite setup

open browser is
 http://127.0.0.1:52451/hub/session/f0b72245-f6fa-8e4d-877c-6a5fa35ddc25/url

 but by the time we set the timeout the browser is:
 http://127.0.0.1:52422/hub/session/5b81cdcc-4126-654a-a486-5e7e8d956a4b/url

 so somehow the code has gotten confused and tried to pick the browser from Open and Close Browser Suite
 which is allready closed.

 if we look at the code for close browser versus close all browsers:

```
 def close(self):
    if self.current:
        browser = self.current
        browser.quit()
        self.current = self._no_current
        self.current_index = None
        self._closed.add(browser)

def close_all(self):
    for browser in self._connections:
        if browser not in self._closed:
            browser.quit()
    self.empty_cache()
    return self.current
```

we can see that close_all actually empties the cache where close leaves the browser in the closed cache.
(as an aside note the Close All actually is a function defined in the ConnectionCache class so we could have and probably should use that one instead of rewriting it here)

Ryan apparently knew this based upon the comment he wrote at the top of the windows suite:

These tests must open own browser because windows opened by earlier tests would otherwise be visible to Get Window XXX keywords even if those windows were closed.

It appears to me that perhaps the ConnectionCache (from robot framework) should hav an unregister funtion (which it currently doesn't) to unregister a closed connection.

In the meantime, the pull request is a good idea and it fixes the issue so I vote to go ahead and merge it.  But we may want to flag this as a separate issue, possible with the robot framework guys?  What does everybody else think?

emanlove:
On 03/06/2012 03:25 AM, Jeremy Johnson wrote:

> ...
>   if we look at the code for close browser versus close all browsers:
> 
> ```
>   def close(self):
>      if self.current:
>          browser = self.current
>          browser.quit()
>          self.current = self._no_current
>          self.current_index = None
>          self._closed.add(browser)
> 
>  def close_all(self):
>      for browser in self._connections:
>          if browser not in self._closed:
>              browser.quit()
>      self.empty_cache()
>      return self.current
> ```
> 
> we can see that close_all actually empties the cache where close leaves the browser in the closed cache.
> 
> Ryan apparently knew this based upon the comment he wrote at the top of the windows suite:
> 
> These tests must open own browser because windows opened by earlier tests would otherwise be visible to Get Window XXX keywords even if those windows were closed.
> 
> It appears to me that perhaps the ConnectionCache (from robot framework) should hav an unregister funtion (which it currently doesn't) to unregister a closed connection.

Some good insight here.

> In the meantime, the pull request is a good idea and it fixes the issue so I'll go ahead and merge it.  But we may want to flag this as a separate issue, possible with the robot framework guys?  What does everybody else think?

I agree. Let's merge and report the issue to robotframework.

Ed

PR #41, opened 2012-03-15
added iframe support, by allowing the element finder to handle filters w...

j1z0:
ok here is my attempt at getting frames to work in painless manner.  Basically I changed the element_finder functionality so that the filter could handled a list of tags.  ie. ['frame','iframe'] as opposed to a single tag.  This also required some changes to the xpath generation for find by default stuff.

Also I updated all the appropriate functions that have to do with frames to look for both frames and frames.  Finally I added a couple of tests.

All the test pass.  But I'm would be interested in feedback, especially if there is a way to make the code cleaner... cooler... nicer :)

### DIFF
--- a/src/Selenium2Library/keywords/_browsermanagement.py
+++ b/src/Selenium2Library/keywords/_browsermanagement.py
@@ -173,7 +173,7 @@ class _BrowserManagementKeywords(KeywordGroup):
         details about locating elements.
         """
         self._info("Selecting frame '%s'." % locator)
-        element = self._element_find(locator, True, True, tag='frame')
+        element = self._element_find(locator, True, True, tag=['frame','iframe'])
         self._current_browser().switch_to_frame(element)
 
     def select_window(self, locator=None):
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -508,7 +508,8 @@ return !element.dispatchEvent(evt);
 
     def _frame_contains(self, locator, text):
         browser = self._current_browser()
-        element = self._element_find(locator, True, True, 'frame')
+
+        element = self._element_find(locator, True, True, ['frame','iframe'])
         browser.switch_to_frame(element)
         self._info("Searching for text from frame '%s'." % locator)
         found = self._is_text_present(text)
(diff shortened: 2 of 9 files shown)


emanlove:
One of the problems I saw with several of the suggested fixes is that they confuse what element_finder means by "tag". element_finder understands "tag" to be both "element type" and  what most people think of as "html tag", see _get_tag_and_constraints.

I see the point of having multiple tags when it comes to frames but I am not sure it makes sense to have multiple tags for anything else as one would be searching for one tag or element type. I'm also not sure about falling though to "find by default" and in particular the key attrs search as frames is more of a element type then an attribute of a specific type.

emanlove:
> I'm also not sure about falling though to "find by default" and in particular the key attrs search as frames is more of a element type then an attribute of a specific type. - emanlove

I'm getting myself all confused here as to the definition of a frame; whether it is an element or an attribute. According to the HTML spec both frame and iframe are elements while the xhtml spec states frame to be an attribute of table. I am still working through some fundamental concepts before I finalize me thoughts on any piece of code.

One thing that I am wondering is whether user will want to differentiate between and frame element and an iframe element.  I can see that some might not care but others might want to only match one or the other. Ultimately I could see a Frame utility class similar to Selenium's Select class.

adwu73:
j1z0, I got your point, I am willing to test your fix on my real apps once done.

From my point view, there are no need to differentiate frame and iframe in test cases, this is an implementation details that tester don't need to care.

j1z0:
@emanlove Frame and iFrame are element types they are not attributes.  Conceptually they are a way to insert one html page inside of another html page.

Regarding the fall through to find by default, that has always been there.  However I get your point, perhaps we can create a special strategy in the element_finder _find_frames or something like that.  I guess this is probably the same idea as the Frame utility class you mentioned.  The advantage there is it would keep the rest of the element_finder.py code cleaner.

Also as it stands now, a tester doesn't need to worry about the differences between frame and iframe, but they can always use the locator argument to specifically limit to a particular tag.  something like css=iFrame[#this_one]  or whatever. I think that much should remain, i.e. by default you don't have to care, but you can pick if you really need to.

adwu73:
j1z0, have this iframe support issue been fixed in selenium2lib 1.0? tks! Adam

j1z0:
the fix is still in the iframe branch, so you could manually pull that down, but it didn't make it into 1.0.  I'm still waiting for acceptance on the merge request from the other devs.

Cheers,
Jeremy

adwu73:
tks, Jeremy for the update

emanlove:
On 03/23/2012 09:51 PM, adwu73 wrote:

> tks, Jeremy for the update
> 
> ---
> 
> https://github.com/rtomac/robotframework-selenium2library/pull/41#issuecomment-4670928

Adam,

A fix for the iframe issue has been merged into rtomac/master.

Ed

PR #44, opened 2012-04-03
Updating INSTALL.rst to promote usage of pip in addition to easy_install

jfilipe:
Shouldn't encourage people to use easy_install when pip is now the standard.

### DIFF
--- a/INSTALL.rst
+++ b/INSTALL.rst
@@ -17,7 +17,11 @@ Installing from PyPI
 --------------------
 
 Selenium2Library is available in the Python Package Index (PyPI_). To install,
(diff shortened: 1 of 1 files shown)


markhu:
A similar change was made at https://github.com/rtomac/robotframework-selenium2library/commit/2daee1e0f

link to latest: https://github.com/rtomac/robotframework-selenium2library/blob/master/INSTALL.rst

emanlove:
As noted by @markhu a similar change was made in 1.0.1 and more explicit wording noting the usage of pip over easy_install for the reasons of installing dependencies under Windows was added in with pull request #113.

Issue #45, opened 2012-04-09
robotframework-selenium2library-1.0.0 dependency issue


apravdin:
robotframework-selenium2library-1.0.0 is supported by pip, while robotframework 2.6.3 is a dependency, but is not supported by pip
This is problematic

j1z0:
Your correct. Let me work on updating the installers. 

Cheers,
Jeremy

jussimalinen:
I changed the acceptance test status checker and doc generation to work with robot 2.7.0. Both will print an error if user tries to run them with robot < 2.7.

I have however changed the setup to use robot >=2.6.0 as dependency, because most users wont run the acceptance tests or generate the docs.

After this we should be able to support pip finally.

See the pull request.

PR #50, opened 2012-05-04
Update to support Robot Framework 2.7 and above.

jussimalinen:
This should fix issue #45

See the comments there.


jussimalinen:
So, any status update on handling this pull request? We really could use the possibility to install with pip...

If there are problems are questions about this request, I am happy to help.

emanlove:
@jussimalinen: is there a reason to limit tests/run_tests to robotframework 2.7? I don't see a change here nor within robotframework between  2.6.1 and  2.7.  I know I tested the selenium2library using 2.6.1.

emanlove:
I've done a quick review of the restrictions on test/run_tests.py to robotframework versions greater than 2.7 and have a few more questions. Is there not functionality within robotframework core for which we can import and which performs the same functionality found in robotframework-selenium2library/test/resources/statuschecker.py?

My issue is that if we relax the requirement to 2.6.0 or greater (which I am strongly in favor of) then the internal tests should be able to check rf-s2l again those same versions of rf.  If I run test/run_tests.py under rf version 2.6.3 with the above changes then only output.xml is created and the warning message is given.  And if I try commenting out the check for 2.7 still with all other change intact I get the expected ImportError exception but I also get the log.html and report.html files which is extremely helpful for debugging purposes.

Is is possible to move this check to test/resources/statuschecker.py so tests can be run on both 2.6.x and 2.7.x?

rtomac:
Pull request looks great.

Ed's question is my only question as well. I also think it's important we can analyze output for these tests against Robot 2.6 since we are choosing to continue supporting it.

Assuming there is no functionality within Robot that can be imported, I'd suggest we include both versions of statuschecker.py (e.g. statuschecker.2.6.py and statuschecker.2.7.py) and conditionally run the correct version of the tool. I'm moving forward with that change now.

emanlove:
On 05/31/2012 09:16 AM, Ryan Tomac wrote:

> Pull request looks great.
> 
> Ed's question is my only question as well. I also think it's important we can analyze output for these tests against Robot 2.6 since we are choosing to continue supporting it.
> 
> Assuming there is no functionality within Robot that can be imported, I'd suggest we include both versions of statuschecker.py (e.g. statuschecker.2.6.py and statuschecker.2.7.py) and conditionally run the correct version of the tool. I'm moving forward with that change now.
> 
> ---
> 
> https://github.com/rtomac/robotframework-selenium2library/pull/50#issuecomment-6035448

Ryan,

You might try something like

``` python

try:
     from robot.result import ExecutionResult

     def process_output(inpath, outpath=None):
         result = ExecutionResult(inpath)
         _process_suite(result.suite)
         result.save(outpath)
         return result.return_code
else:
     from robot.output import TestSuite

     def process_output(inpath, outpath=None):
         suite = TestSuite(inpath)
         _process_suite(suite)
         suite.write_to_file(outpath)
         return suite.critical_stats.failed

```

  although using much better python syntax

Ed

rtomac:
Thanks, good thought.

I just realized the problem is that the new version of statuschecker changes the way status messages are indexed...see Jussi's commit [here](https://github.com/jussimalinen/robotframework-selenium2library/commit/88884a2aa4e75ce5c88047d40df054c1093f34bb). This probably explains Jussi's strict 2.7 check.

I'd suggest we just forgo the status message checking for now when running under Robot 2.6. At least the tests will run through to completion and HTML output files will be generated, and we can get this pull request merged. Then we can reach out to Pekka to see if he has any ideas for dealing w/ this given his knowledge of the tool. Sound good?

emanlove:
Ryan,

Will the HTML files be generated?  That's what my diff was trying to do. Here is slightly better code and I tested it with 2.6.3 and 2.7.1 (although it is still an ugly hack)

``` diff
diff --git a/test/resources/statuschecker.py b/test/resources/statuschecker.py
index 20b6a84..d20088a 100755
--- a/test/resources/statuschecker.py
+++ b/test/resources/statuschecker.py
@@ -40,14 +40,22 @@ detail in the tool documentation.

 import re

-from robot.result import ExecutionResult
-
-
-def process_output(inpath, outpath=None):
-    result = ExecutionResult(inpath)
-    _process_suite(result.suite)
-    result.save(outpath)
-    return result.return_code
+try:
+   from robot.result import ExecutionResult
+   
+   def process_output(inpath, outpath=None):
+       result = ExecutionResult(inpath)
+       _process_suite(result.suite)
+       result.save(outpath)
+       return result.return_code
+except ImportError:
+   from robot.output import TestSuite
+                           
+   def process_output(inpath, outpath=None):
+       suite = TestSuite(inpath)
+       _process_suite(suite)
+       suite.write_to_file(outpath)
+       return suite.critical_stats.failed

 def _process_suite(suite):
     for subsuite in suite.suites:
diff --git a/test/run_tests.py b/test/run_tests.py
index 8b88ee3..e1391d6 100755
--- a/test/run_tests.py
+++ b/test/run_tests.py
@@ -57,7 +57,7 @@ def stop_http_server():

 def process_output():
     print
-    _check_robot_27()
+    #_check_robot_27()
     call(['python', os.path.join(env.RESOURCES_DIR, 'statuschecker.py'),
          os.path.join(env.RESULTS_DIR, 'output.xml')])
     rebot = 'rebot' if os.sep == '/' else 'rebot.bat'
```

rtomac:
Oh, gotcha. I didn't realize your code was meant for statuschecker.py.

Probably stating the obvious, but all statuschecker.py is doing for us is verifying log messages that we're writing to log file while tests are running. It's cool that we can do that, and it's important, but not critical. Without running statuschecker.py at all, the tests still run through properly and all HTML files get generated.

That's why I was suggesting that we just forgo running statuschecker.py for now when running under 2.6. That will allow us to get the pull request merged and a 1.0.1 release built today, and then we can pursue the fix to statuschecker.py that you're suggesting so we can reinstate it. Sound good?

emanlove:
So you're talking of just doing something like this

``` diff
diff --git a/test/run_tests.py b/test/run_tests.py
index 8b88ee3..2a852f1 100755
--- a/test/run_tests.py
+++ b/test/run_tests.py
@@ -56,10 +56,9 @@ def stop_http_server():
     call(['python', env.HTTP_SERVER_FILE, 'stop'])

 def process_output():
-    print
-    _check_robot_27()
-    call(['python', os.path.join(env.RESOURCES_DIR, 'statuschecker.py'),
-         os.path.join(env.RESULTS_DIR, 'output.xml')])
+    if _check_robot_27():
+        call(['python', os.path.join(env.RESOURCES_DIR, 'statuschecker.py'),
+             os.path.join(env.RESULTS_DIR, 'output.xml')])
     rebot = 'rebot' if os.sep == '/' else 'rebot.bat'
     rebot_cmd = [rebot] + [ arg % ARG_VALUES for arg in REBOT_ARGS ] + \
                 [os.path.join(ARG_VALUES['outdir'], 'output.xml') ]
@@ -73,9 +72,10 @@ def process_output():
 def _check_robot_27():
     try:
         from robot.result import ExecutionResult
-    except:
-        print "Analyzing the test results requires Robot Framework 2.7.0 or later."
-        _exit(255)
+    except ImportError:
+        return False
+    else:
+        return True

 def _exit(rc):
     sys.exit(rc)
```

Yeah that sounds good. I'm looking forward to a release today!

jussimalinen:
Sorry for not replying earlier, I was out of office for a few days!

Great, thanks for the release!

About the status checker:

Status checker is a way to run robot and analyze the outputs to check that certain tests have failed or that certain messages have been written to the log. It is a very specialized tool for very special situations. As far as I know it is only used in selenium library and selenium 2 library. Robot's own acceptance tests do the same thing, but haven't been moved to use status checker yet.

I would suggest that we choose between two ways to handle this:
a) get rid of status checker
b) just require that developers have updated to latest major robot version

Tests that verify that a key word fails could be changed to use "Run keyword and expect error" if they dont do that already. Log message checks could just simply be removed. I'm not too sure that it is that important to check to specific log messages in acceptance tests? (Remember that we still can do checks that check the error messages with "Run keyword and expect error") Like this we could just get rid of the status checker and make the acceptance tests as a completely normal robot test run, using what ever robot version the developer has.

In general we at robot's core team have usually built our tools so that we support as many version numbers of robot as possible for users, but for developers we assume that the latest major robot version is in use. Usually this is not an unreasonable requirement. Very few people actually need to do library development and then requiring that they update their robot has not been an issue usually. Only major objection I can think of, is to check that library itself works on robot 2.6, but as it only depends on robot's logging api, I'm dont see a big risk of it breaking between robot releases.

What do you think? Do you Ed have a specific reason you need to do your development using robot 2.6.3?

If we decide to get rid of status checker, I could take a look at changing the tests... We could also get rid of status checker in Selenium Library, but as it is deprecated already, I dont think it worthwhile.

rtomac:
Hi Jussi,

Thanks for information on statuschecker, that's helpful.

Ed wasn't suggesting that he wanted to do his development on robot 2.6, just that we as library maintainers want to be able to test our library on both robot 2.6 and robot 2.7 (since we support users running both).

I think our medium-term goal should be to remove dependency on statuschecker altogether, as you suggest. If you feel up for tackling that, that would be awesome.

In the short-term, I think what we have right now is fine (only running statuschecker if robot 2.7). At least we're still able to run the tests against robot 2.6 as a sanity check, even if it doesn't do the status checking step.

jussimalinen:
Right, sorry I misread. In theory there shouldn't be issues between robot versions as this library only depends on logging api. But as they say: "In theory there is no difference between practice and theory..." I suppose it makes sense to be able to run the tests on different robot versions to be sure.

I agree about the goals. My immediate goal is to make sure I can do my next robot training using this library, and now that installing with pip is possible, I'm probably good for now. I'll get back once I check how things work on different operating systems and browsers...

I will most likely have time to take a look at getting rid of the status checker later this summer. In long term it would be nice to get rid this unnecessary dependency.

PR #63, opened 2012-06-09
Add iframe support with acceptance test cases

kontulai:
Fixes at least "page should contain" and "select frame" to be used with iframes as well. uses @jouk0's original fix. just added acceptance tests on top of the fix.

### DIFF
--- a/src/Selenium2Library/keywords/_browsermanagement.py
+++ b/src/Selenium2Library/keywords/_browsermanagement.py
@@ -173,7 +173,10 @@ class _BrowserManagementKeywords(KeywordGroup):
         details about locating elements.
         """
         self._info("Selecting frame '%s'." % locator)
-        element = self._element_find(locator, True, True, tag='frame')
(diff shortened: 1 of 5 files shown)


kontulai:
That fixed also the "frame should contain" keyword. There is acceptance
test in this commit for that
https://github.com/kontulai/robotframework-selenium2library/commit/0232146dbff20de330349d48fc171406db36a2d8

On Wed, Jun 13, 2012 at 1:50 AM, adwu73 <
reply@reply.github.com

> wrote:
> 
> by the way, there are similiar issue related to iframe within the
> frame_contain code in element.py as well, can you also fix this?
> 
> ---
> 
> 
> https://github.com/rtomac/robotframework-selenium2library/pull/63#issuecomment-6286425

## 

Ilmari Kontulainen

emanlove:
This issue was fixed in rtomac/robotframework-selenium2library#71.

PR #64, opened 2012-06-13
Can you check this change on webdrivermonkeypatches.py whether you can merge the fix into the main branch?

adwu73:


adwu73:
this change can fix issue 58

atts = self.execute_script("return [ window.id, window.name, document.title, document.location ];")
to
atts = self.execute_script("return [ window.id, window.name, document.title, document.url ];")

adwu73:
i have submitted a new pull request

PR #81, opened 2012-07-30
fix for issue 10 - Get Window Titles break in IE

adwu73:
The cause appears to be triggered by document.location parameter in the execute_script call on line 24.
atts = self.execute_script("return [ window.id, window.name, document.title, document.location ];")

IE driver returns a web element object but Firefox driver returns a dictionary:

IE driver returns
[None, '', u'Page Title', ]

Firefox driver returns
[None, '', u'Page Title', {u'search': '', u'assign': u'function assign() {\n [native code]\n}', u'hash': '', u'hostname': u'host.domain', u'replace': u'function replace() {\n [native code]\n}', u'reload': u'function reload() {\n [native code]\n}', u'host': u'host.domain', u'href': u'http://qtst.atpco.org/fmhome/login.jsp', u'pathname': u'/path/to/file.html', u'protocol': u'http:', u'port': ''}] 

Since "Select Window" keyword offers locator options for name, title & url can you change line 24 to:
atts = self.execute_script("return [ window.id, window.name, document.title, document.url ];")

This tested ok for me on IE & Chrome and trusting this reference
http://www.w3schools.com/jsref/prop_doc_url.asp it appears document.url is supported by all browser versions

Hope this helps

### DIFF
--- a/src/Selenium2Library/keywords/_browsermanagement.py
+++ b/src/Selenium2Library/keywords/_browsermanagement.py
@@ -465,7 +465,7 @@ class _BrowserManagementKeywords(KeywordGroup):
         desired_capabilities = capabilities_type
         if capabilities_string:
             for cap in capabilities_string.split(","):
-                (key, value) = cap.split(":")
+                (key, value) = cap.split(":",1)
                 desired_capabilities[key.strip()] = value.strip()
         return desired_capabilities
     
--- a/src/Selenium2Library/utils/browsercache.py
+++ b/src/Selenium2Library/utils/browsercache.py
@@ -26,8 +26,10 @@ class BrowserCache(ConnectionCache):
             self._closed.add(browser)
 
     def close_all(self):
-        for browser in self._connections:
-            if browser not in self._closed:
-                browser.quit()
-        self.empty_cache()
+        try:
+            for browser in self._connections:
+                if browser not in self._closed:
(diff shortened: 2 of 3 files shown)


emanlove:
@adwu73: Can you test this fix both before and after this patch is applied using IE and most important selenium version 2.25.0 (possible using easy_install selenium==2.25.0)? I am interested in the results, both with and without the patch with this specific version of selenium and IE. Does it differ from what you are seeing with any other version of selenium?

adwu73:
it seems that 2.25.0 is only avialable for java binding now, python binding is still on 2.24

adwu73:
it seems that 2.25.0 is only avialable for java binding now, python binding is still on 2.24

adwu73:
it seems that 2.25.0 is only avialable for java binding now, python binding is still on 2.24

adwu73:
it seems that 2.25.0 is only avialable for java binding now, python binding is still on 2.24

adwu73:
sorry for the confusion, I installed python binding 2.25, and IE Driver Server 2.25.2, the result is still the same, without the fix I got "TypeError: object of type 'WebElement' has no len()" when calling Get Windows Titles keyword, and this fix solved the issue

adwu73:
should be fixed in 1.1, will close after 1.1 released and tested

PR #83, opened 2012-08-01
replace Maximizes current browser window from JS to webdriver 

jollychang:
multiple tab browser doen't support Javascript Maximizes browser


Harri:
This would be really nice. At least for me the current Maximize Browser Window does not work, but this fixes it. Thanks.

PR #84, opened 2012-08-03
Follow PEP8 - Style Guide for Python Code

Harri:
Only coding style changes, http://www.python.org/dev/peps/pep-0008/


Harri:
Looks like my git-skills are not really good :) The second patch was not meant to be in the same pull request as the first one. Luckily the second one is really small and simple, just couple typos fixed.

PR #85, opened 2012-08-06
Follow PEP8 - Style Guide for Python Code 

Harri:
Only coding style changes, http://www.python.org/dev/peps/pep-0008/


emanlove:
@Harri We should definitely do this.  I think one barrier though to implementing is first selecting which guidelines to be implemented and which to be strictly enforced and more so educating contributors on PEP8.

emanlove:
Needs resubmitting as changes have occurred.

PR #86, opened 2012-08-08
add actual error message to rundemo.py import exception

kontulai:
If decorator (or any other dependent module) is not installed when running the rundemo.py, it give ambiguous error message. This prints the actual error message (eg. No module named decorator)


emanlove:
Cherry picked and merged in with pull request #113. Changes should be within the next release, Selenium2Library 1.1

PR #87, opened 2012-08-08
Update INSTALL.rst

kontulai:
Added better Windows install description which also contains decorator dependency


emanlove:
Cherry picked and merged in with pull request #113. Changes should be within the next release, Selenium2Library 1.1

PR #88, opened 2012-08-09
Update demo/demoapp/server.py

IsNoGood:
added new global variable DEFAULT_HOST for more easier way to bind to other address than 'localhost'.

### DIFF
--- a/demo/demoapp/server.py
+++ b/demo/demoapp/server.py
@@ -31,6 +31,7 @@ import SimpleHTTPServer
 
 
 DEFAULT_PORT = 7272
+DEFAULT_HOST = 'localhost'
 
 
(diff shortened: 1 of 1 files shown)


emanlove:
This change looks fine. I did notice when testing this change that although one could change the DEFAULT_HOST it has to be done through the code and can't be done through the command line. So for example with the test I can run

```
python demo/rundemo.py -v SERVER:'127.0.0.1:7272' demo/login_tests
```

but I can't do something like

```
python demo/rundemo.py -v DEFAULT_HOST:'127.0.0.1' demoapp start
```

which would be, as I understand, something more like what @IsNoGood is trying to accomplish with this change.

PR #92, opened 2012-08-16
raise exception in (un)select from list...

adwu73:
when nothing match

### DIFF
--- a/src/Selenium2Library/keywords/_selectelement.py
+++ b/src/Selenium2Library/keywords/_selectelement.py
@@ -186,6 +186,8 @@ class _SelectElementKeywords(KeywordGroup):
                 try: option_index = option_labels.index(item)
                 except: continue
             select_func(select, options, option_index)
+        if not option_index:
+            return False
 
     def unselect_from_list(self, locator, *items):
         """Unselects given values from select list identified by locator.
@@ -219,6 +221,8 @@ class _SelectElementKeywords(KeywordGroup):
                 try: option_index = option_labels.index(item)
                 except: continue
             self._unselect_option_from_multi_select_list(select, options, option_index)
+        if not option_index:
+            return False
 
     # Private
     
--- a/src/Selenium2Library/keywords/_selectelement.py
+++ b/src/Selenium2Library/keywords/_selectelement.py
@@ -186,7 +186,7 @@ class _SelectElementKeywords(KeywordGroup):
                 try: option_index = option_labels.index(item)
                 except: continue
             select_func(select, options, option_index)
-        if not option_index:
+        if option_index is None:
             raise ValueError("%s doesn't exist in list '%s'." % (items_str, locator))
 
     def unselect_from_list(self, locator, *items):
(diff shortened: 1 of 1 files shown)


adwu73:
the issue is that when you "select from list" with an non-existing item name, no error will be returned

emanlove:
Looks good to me. Any objections to me merging?

emanlove:
With the performance changes to "Select Element From List ..." code this patch will not merge in. I've written a quick test and this issue is still present with the generic keyword 'Select From List'. (The more specific and speedier 'Select From List By Value', '...By Label', and '...By Index' all catch this error and throw a exception).

I have a new patch which resolves this issue but I have a question on expected functionality. Since 'Select From List'...
- handles both multi-selection lists and single-selection lists and...
- if it is multi-selection and the user provides and empty list\* then it selects all and ...
- if more than one value is given for a single-selection list, the last value will be selected

should we 
1.  Throw a generic "Item not within list" exception if any item is not present (by value or label)  ...or...
2.  Throw a specific "${THIS} item does not exist within ${THAT} list" exception if any multi-selection item is not present ...or...
3.  Throw a specific "${THIS} item does not exist within ${THAT} list" exception if the last value/label on single-selection item is not present ...or...
4. if more than one value is given for a single-selection list and any of those items excluding the last item log that those items are not present ...or...
5. Throw a specific "${THESE} items does not exist within ${THAT} list" exception for any and all items not present
6. ... some combination of the above...

Anyone with suggestions as to the best behaviour should be?

\* only empty list? what if it is None? Note to self to check this

adwu73:
I would suggest to go with the first simple option, thanks!

Adam

emanlove:
@adwu73 I've added some [code](https://github.com/emanlove/robotframework-selenium2library/blob/select-nonexisting-item/src/Selenium2Library/keywords/_selectelement.py) and [a test](https://github.com/emanlove/robotframework-selenium2library/blob/select-nonexisting-item/test/acceptance/keywords/lists.txt) to give some sandbox room to try the methods.

Here is some sample code to test it my changes above

```
$ mkdir s2l-select-non-exist
$ cd s2l-select-non-exist/
~/s2l-select-non-exist $ virtualenv -p /usr/bin/python2.6 --no-site-packages clean-python26-env
~/s2l-select-non-exist $ source clean-python26-env/bin/activate
(clean-python26-env) ~/s2l-select-non-exist $ easy_install selenium decorator docutils robotframework
(clean-python26-env) ~/s2l-select-non-exist $ git clone git@github.com:emanlove/robotframework-selenium2library.git rf-s2l
(clean-python26-env) ~/s2l-select-non-exist $ cd rf-s2l/
(clean-python26-env) ~/s2l-select-non-exist $ git checkout select-nonexisting-item
(clean-python26-env) ~/s2l-select-non-exist $ python test/run_tests.py python ff --suite acceptance.keywords.lists --include OnyThisOne
```

emanlove:
I'm getting the feeling that Selenium does not throw an exception when one tries to unselect an non-existing item. Need to verify this when I find some more time.

PR #116, opened 2012-09-19
browser cache reliability fix

adwu73:
there are cases when browser.quit will throw exceptions, then browser cache will not be cleaned,and all future close_all will fail for same reason

### DIFF
--- a/src/Selenium2Library/utils/browsercache.py
+++ b/src/Selenium2Library/utils/browsercache.py
@@ -26,8 +26,10 @@ class BrowserCache(ConnectionCache):
             self._closed.add(browser)
 
     def close_all(self):
-        for browser in self._connections:
-            if browser not in self._closed:
-                browser.quit()
-        self.empty_cache()
+        try:
+            for browser in self._connections:
+                if browser not in self._closed:
+                    browser.quit()
+        finally:  
+            self.empty_cache()
         return self.current


j1z0:
@adwu73 

Thanks for the pull request. Seems like a good change to me.  Doesn't hurt to be on the safe side. :).  Currious what happens though if you call close_all and there are no browsers open  Does empty_cache handle this issue?  Also is it possible to add some unit tests?

@emanlove -- any thoughts on this?

adwu73:
@emanlove 

I am trying to locate and fix a test stability issue here. After upgrading to 2.25, some times there are orphan IEDriverServer.exe left when test fail (close all browsers in teardown seems not close it every time). So I use AutoIt process close to close IEDriverServer.exe, and I noticed that in some situations, close all will fail in all test cases, because browser.quit will fail with error 10061, and the browser cache will not be cleaned. So all the following close all browsers will fail.

I am still trying to locate what cause the issue, and how to reproduce. However, this fix do make the situation better. Now only the faillig test case and the only one after it report 10061 error, and all following case passed.

I will try to see how to build a unit test once I have a reliable way to reproduce with simple scripts.

emanlove:
@adwu73 @j1z0 

> I am still trying to locate what cause the issue, and how to reproduce.

I'm hesitant to merge in a change when the cause of the problem you describe @adwu73 isn't quite known. It doesn't have to be fully known, but since you are still looking into it, I would prefer to wait a bit on the merge.

> After upgrading to 2.25, ...

It also sounds like it might be a Selenium issue, although it could be Selenium2Library.

> close all browsers in teardown seems not close it every time

If you could share some of the teardown code we might be able to help diagnose more.

> browser.quit will fail with error 10061

What is your test network setup? Do you have any firewalls within it?

adwu73:
I will try to see whether I can reproduce the issue. The tear down code is simple, just Close All Browsers. The test is done on a PC with two VM, one is the target system, the other is a PC where selenium is running as well as Jenkins slave. Jenkins master is running outside the firewall.

emanlove:
@adwu73: I wasn't familiar with error code 10061 but when I searched I saw traces of "firewall" and "disable security software" everywhere; that is why I asked about a firewall. Just something to keep in mind when trying to debug your problem.

emanlove:
@adwu73: I should add that in some recent document changes (see #113) I added some notes on how to write short unittest tests using only Selenium to help determine whether an issue is within Selenium or within Selenium2Library/Robot Framework.

adwu73:
I know how to repoduce this now.

Say I have two simple test, during testcase 1, if I kill IEDriverServer.exe to simulate a IEDriverServer crash, the my test case1 will report two 10054 error when calling CLOSE ALL BROWSERS; while second test case will run normally, and will report 10061 when calling CLOSE ALL BROWSERS during tear down.

here is the log:

Test Case 1                                                                                      | FAIL |
URLError: <urlopen error [Errno 10054] >
Also teardown failed:

## URLError: <urlopen error [Errno 10054] >

[ WARN ] Keyword 'Capture Page Screenshot' could not be run on failure: URLError: <urlopen error [Errno 10054] >
[ WARN ] Keyword 'Capture Page Screenshot' could not be run on failure: URLError: <urlopen error [Errno 10054] >
Test Case 2                                                                                      | FAIL |
...
Also teardown failed:
URLError: <urlopen error [Errno 10061] >

my fix will change the result to this:

Test Case 1                                                                                      | FAIL |

## URLError: <urlopen error [Errno 10054] >

[ WARN ] Keyword 'Capture Page Screenshot' could not be run on failure: No browser is open
Test Case 2                                                                                      | PASS |
...

emanlove:
Can anyone else reproduce this? (I don't have Windows machine setup to try myself, sorry.)

emanlove:
Closing as unable to reproduce.

PR #130, opened 2012-10-08
Debuging and workaround for input_text()

Hi-Fi:


emanlove:
I'm denying this pull request as it seems it contains changes for both what the title ("Debuging and workaround for input_text()") states as well as some profile changes. It is best to work in a feature branch so this doesn't happen.

Also the input_text() changes might be resolved with pull request #167 although I haven't clearly research the differences between what is here and in #167.

PR #137, opened 2012-10-29
+ added support and test for new keyword: click element at coordinates

aaltat:
Ugh

Here is a request for a new keyword: Click Element At Coordinates. 

It must be noted that due selenium constraints it only works with Chrome, with Firefox keyword only clicks at the middle of the element. 

What do think, could it be accepted to the master branch. 

-Tatu


know1nos:
that would be great if that was added, +1

aaltat:
Basically my idea was same as pierreroth64. My acheilles heal was that I suck on web coding and creating standalone html page that would work on Firefox/IE/Chrome was too difficult for me. 

Also this could benefit some unit tests but I suck on that are too.

emanlove:
The acceptance test for this new keyword is failing. I see why and am working on getting it fixed.

PR #148, opened 2012-11-26
add jquery active wait for finding elemnet

jollychang:
for every element find


j1z0:
I do like the idea of adding jQuey support but I have a concern with the way it is implemented in this pull request. 
The issue I have has to do with performance. I someone is not using jQuery this incurs and extra round trip for every single element find. This make a big difference in overall performance when using a remote client. 

I would rather see an option added to the library init such as UseJquery it should default to false abd only call the jQuery bits if its value is true.

jollychang:
@j1z0 ok，I will try to add as a option

jollychang:
update  as options from import Selenium2Library

peritus:
I always implement this in "Robot Framework userland":

```
*** Keywords ***

Wait For Ajax
    Wait for Condition   return window.jQuery.active == 0;
```

And I think it's better kept that way.

omahlama:
I agree with @peritus, there are cases where you want to do things without waiting for ajax, so the choice should be left to the end user. Maybe documenting this as a typical use scenario would be useful instead.

peritus:
@omahlama Yes, documentation would be great, also could include examples for all popular Ajax wrapping frameworks!

emanlove:
Consensus seems to be to use a Robot Framework keyword instead of suggested pull request. Added issue #175.

PR #154, opened 2012-12-10
fixes #65 

### LINKED ISSUE #65: finding elements sometimes cause error TypeError: object of type 'NoneType' has no len() or KeyError: 0
Sometimes the underlying call to Selenium methods such as find_elements_by_xpath return a dict with value=None.  This None value is propagated all the way up to _element._element_find, where len() is used to see how many elements were found.  This will cause Wait Until Page Contains Element to fail immediately, since you can't use len on None.
Unfortunately I don't have a simple test/page to reproduce, but here's a log from executing Selenium2Library.Wait Until Page Contains Element id=delivery_date_select, 1m

08:31:43.590    DEBUG   POST http://127.0.0.1:50341/session/a61ed99c-c9ed-4e49-9045-8b3c7724eeac/elements {"using": "id", "sessionId": "a61ed99c-c9ed-4e49-9045-8b3c7724eeac", "value": "delivery_date_select"}
08:31:44.160    DEBUG   POST http://127.0.0.1:50341/session/a61ed99c-c9ed-4e49-9045-8b3c7724eeac/elements {"using": "id", "sessionId": "a61ed99c-c9ed-4e49-9045-8b3c7724eeac", "value": "delivery_date_select"}
08:31:44.573    FAIL    TypeError: object of type 'NoneType' has no len()
08:31:44.574    DEBUG   Traceback (most recent call last):
  File "<string>", line 2, in wait_until_page_contains_element
  File "C:\apps\Python27\lib\site-packages\Selenium2Library\keywords\keywordgroup.py", line 12, in _run_on_failure_decorator
    return method(_args, *_kwargs)
  File "C:\apps\Python27\lib\site-packages\Selenium2Library\keywords_waiting.py", line 63, in wait_until_page_contains_element
    self._wait_until(timeout, error, self._is_element_present, locator)
  File "C:\apps\Python27\lib\site-packages\Selenium2Library\keywords_waiting.py", line 71, in _wait_until
    while not function(*args):
  File "C:\apps\Python27\lib\site-packages\Selenium2Library\keywords_element.py", line 593, in _is_element_present
    return (self._element_find(locator, True, False, tag=tag) != None)
  File "C:\apps\Python27\lib\site-packages\Selenium2Library\keywords_element.py", line 515, in _element_find
    if len(elements) == 0: return None

ombre42:
ElementFinder.find should return [] when WebDriver.find\* returns None
Sorry I took so long to submit this.
Thank you for your hard work.

### DIFF
--- a/src/Selenium2Library/locators/elementfinder.py
+++ b/src/Selenium2Library/locators/elementfinder.py
@@ -29,8 +29,8 @@ class ElementFinder(object):
     # Strategy routines, private
 
     def _find_by_identifier(self, browser, criteria, tag, constraints):
-        elements = browser.find_elements_by_id(criteria)
-        elements.extend(browser.find_elements_by_name(criteria))
+        elements = browser.find_elements_by_id(criteria) or []
+        elements.extend(browser.find_elements_by_name(criteria) or [])
         return self._filter_elements(elements, tag, constraints)
 
     def _find_by_id(self, browser, criteria, tag, constraints):
@@ -92,7 +92,7 @@ class ElementFinder(object):
             ' and '.join(xpath_constraints) + ' and ' if len(xpath_constraints) > 0 else '',
             ' or '.join(xpath_searchers))
 
-        return browser.find_elements_by_xpath(xpath)
+        return browser.find_elements_by_xpath(xpath) or []
 
     # Private
 
@@ -138,6 +138,7 @@ class ElementFinder(object):
         return True
 
     def _filter_elements(self, elements, tag, constraints):
+        if elements is None: elements = []
         if tag is None: return elements
         return filter(
             lambda element: self._element_matches(element, tag, constraints),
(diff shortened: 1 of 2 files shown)


ombre42:
After monkey patching this code in, I have had a lot less failures while waiting for elements to be present. It lead to uncovering yet another bug in Selenium where webdriver's finds return a dict because IE driver is returning an object instead of a list in the JSON. In Selenium2Library this results in 
KeyError: 0 for finds without tag constraints or
AttributeError: 'unicode' object has no attribute 'tag_name' if there are constraints

I have entered a [defect against selenium](http://code.google.com/p/selenium/issues/detail?id=4999) on returning None and an object, as both violate the JsonWireProtocol for WebDriver.

ombre42:
I have just put a method to process the result from find_by... methods that handles when the result is None and when not isinstance(result, list) in my monkey patching of S2L at work. Do you want me to abort this pull request and submit a new one that handles find returning both None and dict(not list)?

emanlove:
@ombre42 Go ahead and submit the the new one.  I'm spending some time going through issues for the 1.2 release now and plan on getting a fair amount done in the next few days.

ombre42:
combined fix re-submitted as pull request 172

PR #157, opened 2012-12-12
fix bad browser name raising AttributeError

ombre42:
fix for bad browser name raising AttributeError: 'Selenium2Library'
object has no attribute 'fireox' instead of ValueError: fireox is not a
supported browser.
PEP8 near code change and some housekeeping
assertRaisesRegexp not available in py 2.6 :(


ombre42:
See Issue #169. The AttributeError can be confusing to new users.

Issue #77, opened 2012-07-16
Please add support for jQuery in Selenium 2


hemakannan:
Is this feasible? We use jQuery a lot. Add Location Strategy was very useful to us in the front. But can the jQuery support be brought back to Selenium2?

Thank You
Hema

j1z0:
The removal of jquery or actually just sizzle. Is a design decision by the webdriver team. As such it's upstream from S2L. I'm not against shoe horning it back in from s2l but it will be just that,  a shoe horn. Because this really needs to be implemented in the webdriver (server). 

If we do it from s2l I think the main issue you would have is the implicit wait functionality would probably not work at all.  And we would either require the user to write the JavaScript (not good) or agree on what we want to expose ie just sizzle selectors? What about jquery filters etc...

If the above is acceptable then I would be happy to implement it. (should
Be fun :)

Of course if anybody else has other ideas feel free chime in. :)

hemakannan:
Yes Jeremy, I agree that this is something that webdriver should have a support for. Add Location Strategy was actually contributed by my colleague Radek in Selenium1Lib which is like super duper helpful as we use jQuery a lot in our app and these tests. 
I just opened a ticket with webdriver folks and see if there is any possiblity of support JQuery in the webdriver as I know there is definitely groups of people who could benefit from this. 
I don't know if this request will be considered but anyways knowing you can help bringing it back to Sel2Lib is helpful.
Also like you mentioned , if anyone has better suggestions on getting jQuery into Sel2Lib, please chime in..

Thank You so much Jeremy for getting back.

Hema

hemakannan:
Hi Jeremy, so here is the answer to supporting jQuery in webdriver: https://groups.google.com/forum/?fromgroups#!topic/selenium-users/RK0KqX_QPT4

Its not happening.

hemakannan:
Jeremy, any help on this please?

Hema

j1z0:
ok well I can look into developing something.  It might take a couple of weeks as I'm pretty slammed with projects right now.  I can however give you some quick and dirty code sample that may help you get started / code something up in the mean time.

Here is a very simplistic execute jQuery style function.

```
    def _execute_jquery(self, script, *args):
        '''basically just executes a script, and if it fails, loads jquery and
        tries the script again'''
        from selenium.common.exceptions import WebDriverException
        browser = self._current_browser()

        logger.info ("execute jquery with script %s" % script)
        try:
[...]

PR #161, opened 2013-01-10
Add support for jQuery selectors (uses window's jQuery)

tenwit:
This allows "jquery" and "sizzle" locator prefixes, which delegate to the window's jQuery.  Probably not suitable for merging as-is; I think it would be better if jQuery was provided by selenium2library.

But it is easy and probably\* works, so I thought I'd share.  Maybe it will inspire someone to enhance it.
- I have the change working on a work computer, but due to in-house restrictions, I had to type it in by hand on another computer in order to push to github.  The other computer doesn't have robot, so I haven't actually tested the re-typed-in code.... sorry!

### DIFF
--- a/src/Selenium2Library/__init__.py
+++ b/src/Selenium2Library/__init__.py
@@ -56,6 +56,8 @@ class Selenium2Library(
     | dom        | Click Element `|` dom=document.images[56] | Matches with arbitrary DOM express            |
     | link       | Click Element `|` link=My Link          | Matches anchor elements by their link text      |
     | css        | Click Element `|` css=div.my_class      | Matches by CSS selector                         |
+    | jquery     | Click Element `|` jquery=div.my_class   | Matches by jQuery/sizzle selector                         |
+    | sizzle     | Click Element `|` sizzle=div.my_class   | Matches by jQuery/sizzle selector                         |
     | tag        | Click Element `|` tag=div               | Matches by HTML tag name                        |
 
     Table related keywords, such as `Table Should Contain`, work differently.
--- a/src/Selenium2Library/locators/elementfinder.py
+++ b/src/Selenium2Library/locators/elementfinder.py
@@ -11,6 +11,8 @@ class ElementFinder(object):
             'dom': self._find_by_dom,
             'link': self._find_by_link_text,
             'css': self._find_by_css_selector,
+            'jquery': self._find_by_sizzle_selector,
+            'sizzle': self._find_by_sizzle_selector,
             'tag': self._find_by_tag_name,
             None: self._find_by_default
         }
@@ -55,6 +57,12 @@ class ElementFinder(object):
         if not isinstance(result, list):
             result = [result]
         return self._filter_elements(result, tag, constraints)
+
+    def _find_by_sizzle_selector(self, browser, criteria, tag, constraints):
+        js = "return jQuery('%s').get();" % criteria.replace("'", "\\'")
+        return self._filter_elements(
+            browser.execute_script(js),
+            tag, constraints)
 
     def _find_by_link_text(self, browser, criteria, tag, constraints):
         return self._filter_elements(
--- a/src/Selenium2Library/locators/tableelementfinder.py
+++ b/src/Selenium2Library/locators/tableelementfinder.py
@@ -68,6 +68,8 @@ class TableElementFinder(object):
     def _parse_table_locator(self, table_locator, location_method):
         if table_locator.startswith('xpath='):
             table_locator_type = 'xpath'
+        elif table_locator.startswith('jquery=') or table_locator.startswith('sizzle='):
+            table_locator_type = 'sizzle'
         else:
             if not table_locator.startswith('css='):
                 table_locator = "css=table#%s" % table_locator
--- a/src/Selenium2Library/locators/tableelementfinder.py
+++ b/src/Selenium2Library/locators/tableelementfinder.py
@@ -17,6 +17,20 @@ class TableElementFinder(object):
             ('css', 'row'): [' tr:nth-child(%s)'],
             ('css', 'col'): [' tr td:nth-child(%s)', ' tr th:nth-child(%s)'],
 
+            ('jquery', 'default'): [''],
+            ('jquery', 'content'): [''],
+            ('jquery', 'header'): [' th'],
+            ('jquery', 'footer'): [' tfoot td'],
+            ('jquery', 'row'): [' tr:nth-child(%s)'],
+            ('jquery', 'col'): [' tr td:nth-child(%s)', ' tr th:nth-child(%s)'],
+
+            ('sizzle', 'default'): [''],
+            ('sizzle', 'content'): [''],
+            ('sizzle', 'header'): [' th'],
+            ('sizzle', 'footer'): [' tfoot td'],
+            ('sizzle', 'row'): [' tr:nth-child(%s)'],
+            ('sizzle', 'col'): [' tr td:nth-child(%s)', ' tr th:nth-child(%s)'],
+
             ('xpath', 'default'): [''],
             ('xpath', 'content'): ['//*'],
             ('xpath', 'header'): ['//th'],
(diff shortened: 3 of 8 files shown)


hemakannan:
Hi, Guys, does this mean the latest build of selenium Library will have this? I submitted an issue a while back to have jQuery incorporated back in to selenium2Library of Robot Framework as all our tests in RC use jQuery very heavily. Please advice.

Thank You
Hema

tenwit:
If you build it from my fork, it will work (unless I checked in a typo.. see my comment above!).  But I doubt Ryan will merge this; IMO it's not the correct solution, it's just a place to start.

emanlove:
I've made some time starting yesterday to review issues for the next release.  I will look this over. - Ed

hemakannan:
Nice !

hemakannan:
Ed, any updates on this?

emanlove:
@j1z0 have you gotten a chance to review this implementation for jQuery support?  I know you were hesitant with adding in jQuery. (For those who haven't seen the historical discussion about jQuery please review Issue #77.)

hemakannan:
Hey Guys
It would be very helpful if jquery support is implemented in RF
selenium2lib just like how it is in RF selenium RC.

emanlove:
@hemakannan Again this is an upstream choice by the Selenium WebDriver team to remove support for jQuery/sizzle. As this is the robotframework-selenium2library we are tying together existing functionality within WebDriver making it accessible through Robot Framework.  So if it doesn't exist within WebDriver I'm not sure we should try to shove it in.

@tenwit Do you have any tests for this?

hemakannan:
Hi Ed
Yes we have many tests that use jquery. Selenium never had jquery support i
think. My coworker created this keyword called add location strategy and
packaged it. I was wondering if it would be feasible for RF sel2 to have
that .

emanlove:
@hemakannan Let's see his code as I am wandering if there is a solution that works like @peritus's [solution for ajax wait](https://github.com/rtomac/robotframework-selenium2library/pull/148#issuecomment-12425477).

The solution in this pull request looks clean, in that is fits into the Sel2Library locator strategy but I guessing that it will fail in that browser.execute_script(js) won't return the same thing as browser.find_elements_by_XXXXXXXXX. That's why I want to see a test for this solution.

tenwit:
I don't, but I'm sure I can cobble a few together in no time; it's just a selector prefix.  But not until Tuesday, most likely :(

hemakannan:
Hi Ed, tenwit :) 
Found some links to this implementation in selenium1Library RF. Thanks so much guys.
http://www.radekw.com/blog/2009/11/03/jquery-selectors-in-selenium/
http://robotframework-seleniumlibrary.googlecode.com/hg/doc/SeleniumLibrary.html?r=2.8#Add Location Strategy
http://code.google.com/p/robotframework-seleniumlibrary/issues/detail?id=59

emanlove:
@hemakannan Your missing the point in that all the old SeleniumLibrary did was use the jQuery functionality provided by Selenium RC. And actually this was only implemented by a jQuery plug-in which is why I think you had to use the add_location_strategy in the first place.  The SeleniumLibrary wasn't doing anything special but simply acting as a bridge between Robot Framework and the functions provided by Selenium RC.  Selenium WebDriver dropped support for jQuery/sizzle thus we don't have anything to bridge to or to connect into directly with Selenium2Library.

What I suspect @tenwit will find in testing this change is that the execute javascript will return success but not return a Selenium WebElement which is what selenium needs to perform any function on like click element.  It will be trying to "click successful javascript return value" which is wrong.

hemakannan:
I see. Gotcha.

emanlove:
@hemakannan I wish it was as simple as doing something like SeleniumLibrary but tying in the locator with the actions is really problematic with Selenium dropping support for jQuery/sizzle.

hemakannan:
:( well I hoped to see if it was as easy as adding like selenium 1 lib. All
of our tests have something like "jquery = " which we could easily start
using css=
but it would be with writing new tests.

I just hoped that we could just literally migrate all of our tests like
that we :-)

But looks like its not possible.
Also Ed while we were discussing on this thread i realized that something
like :contains("....") and wild card character would not be possible
either. Correct me if i am wrong because that's another thing we use in our
tests example jquery = *:contains("...")

We created a user keyword like this "wait until page contains elements with
text" that in turn uses wait until page contains element and the above
jquery example.

emanlove:
@hemakannan xpath has a "similar" 'contains' functionality.  I use the quotes around "similar" because I am not sure if jQuery and xpath use the exact same syntax but it is something to look into.

tenwit:
@emanlove I haven't written any tests but I am using this patch while migrating our tests to Selenium2Library.  It's working fine, as far as it goes.  It only works because we use jQuery on all our pages, of course, but that's good enough for now.

j1z0:
Ok I had a look at the code and here is my .02.

I've written something similar and I'm about 99% sure that the execute_script will indeed return the same thing as find_element.  So I think we are OK on that point.

And I like the fact that this is using a separate location strategy so it doesn't interfere will anybody who doesn't want to use it.

Now as tenwit correctly points out in his initial comment you have to have jQuery bundled with your application for this to work.  So it doesn't achieve the same functionality that was available in Selenium 1.

Thus since I think that is really the goal I would probably add a check if jQuery exists and if it doesn't exist load it.  That way this functionality will work on pages / applications that do not have jQuery installed.  The downside of that approach is it requires a second round trip from each and every call, which can really slow things down if your a using a remote connection.  Thus a different way would be to have a seperate load_jquery keyword that could be called one before any calls using the jQuery location strategy were made... of course there is an issue if the application uses something like prototype where you need to load jQuery with no conflicts so it doesn't overwrite existing frameworks.  The way we got around this in the past was with some code like this:

```
    def _load_jquery(self):
        import os
        logger.info("load jquery")

        #just the location where we were storing a local copy of jQuery
        #see comments below this may be better to load directly from the jQuery github page.
        ROOT = os.path.dirname(__file__)
        LIB_DIR = os.path.join(ROOT, '..', 'lib')
        jquery_file = os.path.join(LIB_DIR, "jquery-latest.min.js")
        jquery_lib = open(jquery_file, 'r').read()

        #NOTE: I'm loading a special version of jquery that uses $j
        #so basically we are changing jQuery from using $ to using $j
        #browser.execute_script("var $j = jQuery.noConflict();")

        browser = self._current_browser()
        browser.execute_script(jquery_lib)
```

this is less than ideal, because it relies on a modified version of jQuery that exists on users system.  We could I guess package jQuery with S2L, but it may be better, just to ignore (or at least document) the potential collision issues (read, don't use with prototype.js) and then we could just pull jQuery directly from github and thus always get the latest version.

The other thing to document / keep in mind here is that execute_script does not honor the implicit_wait functionality built into Selenium 2.  So if you really want to do it write you should build in the implict wait functionality into the js script that you call... i.e. put some sort of loop / retry in the jquery.

The ultimate decision I'll leave to however implements it, cause I don't really have time to do it at the moment.

hemakannan:
Yes tenwit and Jeremy
Our applications understands jquery very well that's why we have all tests
with query selectors. Can we try to get what you have so far working and
try it at our end please.

j1z0:
@hemakannan -- I would suspect that tenwit's fork will work just fine for you.  Have you tried your tests against it?

hemakannan:
Hey Jeremy. No
How should i use it ?

j1z0:
after you have downloaded / installed the fork, you just use a locator that starts with jquery= or sizzle= and that should trigger the query support.

Cheers,
Jeremy

hemakannan:
Ok will keep you guys posted.

zealic:
Current version(1.1.0) not support css pseudo class "contains", +1 for jQuery locator.

hemakannan:
Sorry guys. Been busy. I will let you guys know soon.

oforinash:
Hi Everybody my name is Gabriel and a colleague of Hema, I downloaded and installed the fork for Jquery and i'm pleased to inform you that it worked like a charm.  I will like to thank you all for your support and the implementation for this added code.  Our transition to webdriver will be made much easier with this addition.  Hopefully this fork will be added to an official release soon.. Thank you.

hemakannan:
Hey Guys, can this be pushed to the regular/main branch so it can be included by default in the builds we get ?

emanlove:
I think the only blocker on this was a lack of internal tests.

As for a keyword to check to see if jQuery is available this functionality request can be made in a new issue.

tenwit:
Sorry guys, completely forgot about this fork!  And I've left my dev computer at work this weekend :(  I'll write some tests next weekend if noone gets around to it first.  I'll just find all the tests for the CSS locator and duplicate them for this one, is that about right?

tenwit:
Right, some basic tests added.  They use jQuery's own CDN to fetch jQuery, so the tests will fail if there is no internet access.  I'll look into making the tests non-critical tomorrow, assuming I get time.

Suggestions on what extra tests to add will be gratefully received,

Issue #167, opened 2013-02-07
Provide support for verifying text entered into a textarea element


stevejefferies:
Currently S2L allows the verification of text fields through "Textfield Should Contain" and "Textfield Value Should Be" but this doesn't work for textarea elements since these keywords resolve to a tag of "input" in _get_tag_and_contraints. I tried "Element Text Should Be" and "Element Should Contain" keywords but these don't evaluate the text value currently set since these call _get_text rather that _get_value so there doesn't appear to be a way of evaluating the text you have input into a textarea field.

I have currently implemented an additional keyword locally along the lines of:

```
def textarea_should_contain(self, locator, expected, message=''):
    actual = self._get_value(locator, 'text area')
    try:
        if not expected in actual:
            if not message:
                message = "Text area '%s' should have contained text '%s' "\
                          "but it contained '%s'" % (locator, expected, actual)
                raise AssertionError(message)
    except TypeError:
        raise TypeError("Could not find element: {0}".format(locator))
    self._info("Text area '%s' contains text '%s'." % (locator, expected))
```

and modified _get_tag_and_constraints to allow a tag of 'text area' which returns a tag of 'textarea' which works for my needs. (also done for exact text matches in a separate keyword).
What are your thoughts on this implementation? Is it better to have a separate set of keywords for textarea elements? Or would it be better to allow support for textarea elements within the "Textfield Should Contain" keyword (and exact match keyword)?

Happy to submit a pull request for this modification if you think it's correct.

j1z0:
I think keep it separate an your implementation looks great. I think you should submit the pull request and we will merge it. 

Cheers,
Jeremy

emanlove:
Fixed in #168.

PR #168, opened 2013-02-08
Support for verifying text entered into textarea elements

stevejefferies:
Addresses the issue described in #167.


emanlove:
@stevejefferies if you add some tests I might bump this up to an earlier release. Thanks!

PR #172, opened 2013-02-27
Fixes Issue 65 and similar problem

ombre42:
this addresses the issue mentioned in #65 as well as other various errors that occur when find methods return a dict.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+language: python
+python:
+  - "2.6"
+  - "2.7"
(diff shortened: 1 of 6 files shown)


ombre42:
I don't get why the other commits show up. I merged the upstream changes before creating this fork.

emanlove:
@ombre42  I've noticed that happens to me every once in a while.  Usually those commits are ignored when merged as there are already in the branch.

Issue #181, opened 2013-03-12
click image cant use locator src-need fix


qitaos:
@rtomac @emanlove   when i user click image src=xxx.jpg, there is a error message "ValueError: Element locator with prefix 'src' is not supported".
But the document is "Clicks an image found by `locator`.

```
    Key attributes for images are `id`, `src` and `alt`. See
    `introduction` for details about locating elements."
```

It means i can use src for locators. 
Well, I checked the elementfinder.py in Selenium2Library.locators, and I add one line "'src': self._find_by_default," in __init__, then it works.
Can you fix this bug   in next version? Tks

qitaos:
well, `src` and `alt` both have the same problem in CLICK IMAGE.

ombre42:
Thank you for reporting this.
I think the problem here is that you are misunderstanding the documentation.
The key attributes refer to how the locator is used if you **do not** use a locator prefix.
When you use a locator without a prefix like this:
Click Image    foo
Selenium2Library searches using this XPath expression:
//img[(@id='foo' or @name='foo' or @src='foo' or @alt='foo' or @src='<location>foo')]
In other words, without a prefix, the **entire** locator is matched against the values of the key attributes.
This is explained in the library documentation but it could certainly be clearer.
So you should use Click Image xxx.jpg instead of Click Image src=xxx.jpg

If you want to match only alt or src, you will have to use a different prefix/strategy and not rely on the default strategy.
Do you have any suggestions on how to make the documentation clearer?

aaltat:
There has not been activity for this issue for long time and we are planning to close the old issues if they are not anymore valid ones. If you feel that this is still valid issue, could you send an update. If update is not received in a month, we will assume that this is not anymore an issues and we will close this issue.

PR #182, opened 2013-03-12
fix click image locator by 'src' and 'alt'

qitaos:
@rtomac @emanlove @j1z0 
I found 2 way to fix this issue #181.
1. add src and alt in self._strategies
2. add if before self._strategies.get(prefix), if prefix is src or alt,
then make it to None.
I choose the 1.


emanlove:
Reviewing the discussion on #181 this appears to be more of a documentation error then an issue with locator operation.  I am closing this pull request without merging.

PR #187, opened 2013-03-28
Non-breaking spaces should be treated as usual spaces

ahitrin:
Hello!

We're trying to move our tests from SeleniumLibrary to Selenium2Library. But we've found that our tests are failing on checks upon elements that contain non-breakable spaces (which works fine with SeleniumLibrary).

Here I provide the simplest test case that demonstrates that behavior. I have no fix yet, but I'll try to provide it ASAP. Maybe you can suggest the way I should follow to fix this (it would be great!).

### DIFF
--- a/test/acceptance/keywords/content_assertions.txt
+++ b/test/acceptance/keywords/content_assertions.txt
@@ -218,3 +218,7 @@ Xpath Should Match X Times
     Xpath Should Match X Times  //input[@type="text"]  ${1}
     Run Keyword And Expect Error  Xpath //input[@type="text"] should have matched 2 times but matched 1 times  Xpath Should Match X Times  //input[@type="text"]  2
 
+Non-breaking spaces should be treated as ususal spaces
+    [Setup]  Go To Page "non_ascii.html"
+    Page Should Contain  Non-breaking space
+
--- a/test/resources/html/non_ascii.html
+++ b/test/resources/html/non_ascii.html
@@ -26,5 +26,6 @@
 <li>здравствуйте
 <li>こんにちは
 </ul>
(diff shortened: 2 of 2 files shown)


ombre42:
An initial investigation shows that changing the implementation of _is_text_present to:

```
return text in self._get_text("tag=html")
```

makes the failing test pass in FF & IE without breaking other tests.
Converting the entire page to text and sending it over the wire is expensive (especially for those using Sauce Labs On Demand, etc.).
However, this is [what the watir-webdriver camp does](http://wiki.openqa.org/display/WTR/Cheat+Sheet), e.g. 

```
browser.text.include?("expected text"). 
```

where the [implementation](https://github.com/watir/watir-webdriver/blob/master/lib/watir-webdriver/browser.rb) of text is

```
    def text
      @driver.find_element(:tag_name, "body").text
    end
```

We should definitely be careful with this since this is a popular keyword.

ahitrin:
Thank you! I'll try it soon

ahitrin:
> An initial investigation shows that changing the implementation of _is_text_present to:

```
return text in self._get_text("tag=html")
```

We've tried this and found that such workaround has two serious flaws. First, it works slower than original implementation. Second, it freezes on our test suite after ~30 minutes of work. You're right, a brutal hack in such popular word cannot leave no trace.

How do you think, can we implement another (not so hacky) workaround that allows us treat nbsp's as usual whitespaces? If it's not possible for now, I'll try to find another way to find these elements in our tests (and this pull request should be closed in that way, as I think).

ahitrin:
I could not implement fix, better to close this

PR #195, opened 2013-04-19
Added 'Click Element On Coordinates' keyword

pierreroth64:
The "old" Selenium library had a coordinates optional argument for the Click Element keyword. My proposal is to add a new keyword which takes x and y offsets.

### DIFF
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -260,6 +260,18 @@ class _ElementKeywords(KeywordGroup):
         self._info("Clicking element '%s'." % locator)
         self._element_find(locator, True, True).click()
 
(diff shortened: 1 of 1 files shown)


pierreroth64:
I will try to add an acceptance test soon.

ombre42:
Pending pull request #137 for 'Click Element At Coordinates' is the same as this except the offset is from the center of the element instead of the top-left corner. Perhaps these two use cases could be combined into a single keyword that takes an extra argument relating to what to use as the base for the offset?

pierreroth64:
You're right. I'm sorry, i didn't checked the existing pull requests. I'll try to propose an extra keyword to be able to click from the center or from the top left. I close this pull request for now.

PR #212, opened 2013-06-12
Fixed compatibility with RobotFramework v2.8

F1ashhimself:
Removed unnecessary assignment of self.current_index for compatibility with RobotFramework v2.8, in new robot version current_index is a property without setter.
You don't need to take care about None in current_index anymore if this will be true "self.current is self._no_current" it will return None by default.

Also fixed keyword.async_javascripts test for 100% passing rate.

### DIFF
--- a/src/Selenium2Library/utils/browsercache.py
+++ b/src/Selenium2Library/utils/browsercache.py
@@ -22,6 +22,10 @@ class BrowserCache(ConnectionCache):
             browser = self.current
             browser.quit()
             self.current = self._no_current
+            cls_attr = getattr(type(self), 'current_index', None)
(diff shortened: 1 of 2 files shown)


pekkaklarck:
Sorry for breaking backwards-compatibility with the change to ConnectionCache.current_index. I'll make it an assignable property to fix the issue on Robot side in 2.8.1:
https://code.google.com/p/robotframework/issues/detail?id=1474

Changing this in Se2Lib is probably a good idea too. It seems that Selenium2Library doesn't actually use current_index for anything, and ConnectionCache never reads it either, so the simplest fix is removing `self.current_index = None` altogether.

emanlove:
Fixed in #251 (without unnecessary changes async_javascript tests.)

PR #237, opened 2013-10-07
Fixed Mouse Up keyword attempting to click and hold one more time before release

mayask:
I've almost fixed all failing tests, please take a look at the details below:
1. Fixed Mouse Up keyword attempting to `click_and_hold` before `release` therefore emitting `mousedown` event;
2. Fixed `Unselect Non-Existing Item From List` test post-processing `LOG` command failing with wrong keyword index;
3. Refactored `Mouse Down On Link` test to be clearer;
4. Updated `Click Element At Coordinates` test with more "expected" coordinates.

Also I would like you to take a look at `Click Element At Coordinates` test one more time. It seems to me that selenium doesn't perform `move_to_element.move_by_offset.click` action chain well. I've submitted the issue on [google code](http://code.google.com/p/selenium/issues/detail?id=6370). Actually it's the last issue remained, and it would be really cool to get it fixed soon.

### DIFF
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -274,7 +274,7 @@ class _ElementKeywords(KeywordGroup):
 
     def click_element_at_coordinates(self, locator, xoffset, yoffset):
         """Click element identified by `locator` at x/y coordinates of the element.
-        Cursor is moved and the center of the element and x/y coordinates are 
+        Cursor is moved and the center of the element and x/y coordinates are
         calculted from that point.
 
         Key attributes for arbitrary elements are `id` and `name`. See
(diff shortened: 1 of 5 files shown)


emanlove:
@myaskevich: Thanks a lot for fixing these issues!

I wanted to say though that your changes to Click Element At Coordinates were in error but I see you're correct. There is an error in the implementation of that keyword and that was my error. In dealing with a Selenium/Firefox [issue](http://code.google.com/p/selenium/issues/detail?id=6141) I was trying different methods of clicking the element. For some (very poor) reason I left it with the implementation of 

```
... move_to_element(element).move_by_offset(xoffset, yoffset) ...
```

instead of

```
... move_to_element_with_offset(element, xoffset, yoffset) ...
```

thinking these two action chains were equivalent. They are not. The former moves to the center of the element (move_to_element) then **offsets from the center**. The latter **offsets from the top left corner** which was the intended action (See [[1](http://code.google.com/p/selenium/source/browse/py/selenium/webdriver/common/action_chains.py#180)]).

Thanks @myaskevich for bringing my error to light and for fixing these issues. I will go ahead an merge these changes.

PR #249, opened 2013-12-09
Implemented "Shift Click Element" keyword

SoCalLongboard:
I've implemented a new keyword "Shift Click Element" which allows user to click an element while holding down the Shift Key.  In Firefox, this combination maps to "Open in New Window".  (In-house, we use this routine to initiate the downloads of files which are produced by the reporting features of our application.)  The implementation could easily be re-used for other key/click combinations (Alt-Click, Ctrl-Click, etc.)  We've been using this routine in-house for a few months now.

If you'd prefer, I could re-implement this as Key Click Element (a more generic implementation) and re-implement the Shift Click Element keyword as a user of the generic keyword (passing Keys.SHIFT as a parameter to the generic keyword implementation).

### DIFF
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -272,20 +272,29 @@ class _ElementKeywords(KeywordGroup):
         self._info("Clicking element '%s'." % locator)
         self._element_find(locator, True, True).click()
 
-    def shift_click_element(self, locator):
-        """Shift-Click element identified by `locator`. Simulates holding down
-        the Shift key while clicking an element.
+    def key_click_element(self, keyname, locator):
+        """Click element identified by `locator` while holding down a key.
+        The 'keyname' value should be expressed as a Keys constant as specified
+        in the selenium.webdriver.common.keys module.
 
         Key attributes for arbitrary elements are `id` and `name`. See
         `introduction` for details about locating elements.
+        
+        Example:
+        | Key Click Element | CONTROL | id=downloadLink |
         """
-        self._info("Shift-Clicking element '%s'." % locator)
+        self._info("Clicking element '%s' while holding down the '%s' key." % (locator, keyname))
+        
+        # get the key code which corresponds to the passed key name
+        # invalid values will fail here (but meaningfully)
+        key = eval("Keys." + keyname)
+        
         element = self._element_find(locator, True, True)
-        # press the Shift key
-        ActionChains(self._current_browser()).key_down(Keys.SHIFT)
+        # press the key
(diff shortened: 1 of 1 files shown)


emanlove:
@SoCalLongboard: I think a more generic solution would be best. Could you resubmit a patch for a generic key press? Thanks!

SoCalLongboard:
Hi Ed -

I'd be happy to -- I'll get sometime later today.
- Mike

SoCalLongboard:
Hi Ed - 

I've replaced the 'Shift Click Element' method with a more generic 'Key Click Element' method which allows the passing of the key names from selenium.webdriver.common.keys.

will-moore:
I would really like this feature to be merged if possible. I just tried using the code in my own library by subclassing _ElementKeywords but this failed because I then had duplicates of all the keywords in that class. So I ended up patching the source code itself, but this is not a feasible solution.

However, after doing this, I didn't see any difference in click behaviour using CONTROL or SHIFT modifiers, so maybe I'm doing something else wrong.

src/Selenium2Library/keywords/_element.py
```diff
@@ -272,6 +272,30 @@ def click_element(self, locator):
         self._info("Clicking element '%s'." % locator)
         self._element_find(locator, True, True).click()
 
+    def key_click_element(self, keyname, locator):
```
emanlove, line 275: [Thinking out loud before researching as reminder to self] Should the order of the arguments be self, locator, keyname to be consistant with the other keywords?

Also I am thinking that we should add an optional argument for specifying which button to click.

SoCalLongboard, line 275: I absolutely agree that the argument sequencing should be consistent with other methods.

emanlove, line 275: Just to let you know I'm currently working on a test page for this keyword. Should upload an intial attempt in the next few hours...

emanlove:
I've created a test page for showing both click events and key press events. Its a work in progress but I wanted to start with something which tests this change especially with @will-moore's comment about no difference between CONTROL and SHIFT modifiers. The page can be found on [my key-click branch of Selenium2Library](https://github.com/emanlove/robotframework-selenium2library/blob/key-click/test/resources/html/javascript/keypress_click.html).

just-be-dev:
I think this is too specific of an implementation. 

We should either support action chains or create keywords for things like `Key Down` and `Key Up` events.

fredweb75:
It is a lot like the Press Keys new feature discussed in https://github.com/robotframework/Selenium2Library/issues/498. So yes, it should be implemented for all special Key combo.

Issue #254, opened 2013-12-28
Better project introduction to PyPI


pekkaklarck:
The current introduction just says:

```
"""Selenium2Library is a web testing library for Robot Framework that leverages the Selenium 2 (WebDriver) libraries."""
```

It would be great to have some more information, including installation instructions there. A very easy and DRY solution is reading the current README.rst content to long_description in setup.py. We recently did that with StatusChecker project and it works great. The same information is visible in [BitBucket project pages](https://bitbucket.org/robotframework/statuschecker/src) (GitHub obviously shows READMEs same way) and on [PyPI pages](https://pypi.python.org/pypi/robotstatuschecker), and the [implementation](https://bitbucket.org/robotframework/statuschecker/src/master/setup.py) was pretty trivial.

aaltat:
I think this is now done in: https://github.com/robotframework/SeleniumLibrary/commit/a10a421bf8012d6d3927e66571c89ecd5d2571c8

PR #260, opened 2013-12-29
Setup enhancements, including reading long_description from README.rst (#254)

pekkaklarck:
This pull requests consists of two commits that enhance setup.py:
- f110449 implements reading long_description value from README.rst (#254).
- 9c3b565 contains general clean-up but doesn't change functionality.

Feel free to cherry-pick only the former commit if you don't agree changes in the latter or are afraid they break something.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -10,10 +10,8 @@ from setuptools import setup
 
 execfile(join(dirname(__file__), 'src', 'Selenium2Library', 'version.py'))
 
-DESCRIPTION = """
-Selenium2Library is a web testing library for Robot Framework
-that leverages the Selenium 2 (WebDriver) libraries.
-"""[1:-1]
+with open(join(dirname(__file__), 'README.rst')) as f:
+    DESCRIPTION = f.read()
 
 setup(name         = 'robotframework-selenium2library',
       version      = VERSION,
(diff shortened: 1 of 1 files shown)


pekkaklarck:
Travis build is failing but it seems to have been doing so already earlier.

setup.py
```diff
@@ -1,46 +1,41 @@
 #!/usr/bin/env python
 
-import sys
-from os.path import join, dirname
+from os.path import abspath, dirname, join
 
-sys.path.append(join(dirname(__file__), 'src'))
```
emanlove, line 6: Removal of this line breaks the install process as...

setup.py
```diff
@@ -1,46 +1,41 @@
...
 from ez_setup import use_setuptools
```
emanlove, line 5: ... this line requires ez_setup which we include under the src directory. We should actually remove the dependency on ez_setup.py and remove it from source.

aaltat:
@pekkaklarck is this PR still valid one, or should we close it?

pekkaklarck:
Enhancements themselves are valid but but due to conflicts it's easier to just re-do them than trying to fix this PR.

PR #263, opened 2014-01-09
Register browser even if get url fails.

emanlove:
Test and patch for fixing issue rtomac/robotframework-selenium2library#70.


ombre42:
Looks good to me

PR #266, opened 2014-01-15
Update _managementbrowser.py

githubccruz:
I Try to get the instance from Robot to use on Python, for use the current instance of the Robot Browser that created, I had to modify the source code by changing the function "def get_source (self):
         "" "Returns The entire html source of the current page or frame." ""
         self._current_browser return (). get_page_source () "
  I return to the current instance (return self._current_browser ()). The idea came to use BuiltIn that returns the instance of a specific library such as "browser = BuiltIn () get_library_instance. (" Selenium2Library ")
browser.get_source browser = () "=> modified with the function of selenium webdriver using. And so to use the power provided by the library" selenium ".

My suggestion is that a method that returns the current "webdriver" Robot is using that to make use of it from Python is added.

get_current_browser def (self):
        "" "Return current webdriver object" ""
self._current_browser return ()

En caso de que no me supe explicar bien, o existe una forma de manejar esta situación por favor me dejan saber.

gracias.

I created the next function to my propuse, i hope that helps
    def get_current_browser(self):
        """Returns the current webdriver object"""
        return self._current_browser()

### DIFF
--- a/src/Selenium2Library/keywords/_browsermanagement.py
+++ b/src/Selenium2Library/keywords/_browsermanagement.py
@@ -304,6 +304,10 @@ class _BrowserManagementKeywords(KeywordGroup):
     def get_source(self):
         """Returns the entire html source of the current page or frame."""
         return self._current_browser().get_page_source()
+                 
+    def get_current_browser(self):
+        """Returns the current webdriver object"""
+        return self._current_browser()
 
     def get_title(self):
         """Returns title of current page."""


emanlove:
I don't see the purpose of this change. What would one do with the instance of the webdriver within Robot Framework? I don't see this as being useful nor necessary. If anyone disagrees, please speak up and reopen this pull request.

Issue #270, opened 2014-01-27
Window keywords fail if window.id is an unexpected type


pr0chz:
Example:
`Select Window     name=view_more_data`

Result:
`TypeError: object of type 'bool' has no len()`

This also applies to other functions regarding working with windows. OS Windows 7, versions used:

```
robotframework-selenium2library (1.5.0)
selenium (2.39.0)
```

Tested on Firefox and IE, both broken.

Problems seems to reside in function get_current_window_info in webdrivermonkeypatches.py. If you erase `len(att)` in:

```
        atts = [ att if att is not None and len(att) else 'undefined'
            for att in atts ]
```

it starts to work ok. Not sure if it does not brake anything else though.

HelioGuilherme66:
Hi,

I have tested in Linux CentOS 6.5 (x64bit), Firefox, and no problems here.
Attached test and logs.

Hélio

ombre42:
I suspect the problem might be with the window ID. It is not part of any specification AFAIK. Probably used because popular frameworks use window.id.
Could you please run something like this to see which attribute is a bool?

```
    @{attribs}=    Execute Javascript    return new Array(window.id, window.name, document.title, document.url);
    :FOR    ${attrib}    IN    @{attribs}
    \    ${type}=    Convert To String    ${attrib.__class__}
    \    Log    ${type}: ${attrib}
```

Also are you able to reproduce this on different websites?

pr0chz:
I guess your suspicion is right:

```
INFO    Executing JavaScript: return new Array(window.id, window.name, document.title, document.url);
INFO    @{attribs} = [ False | view_more_data | title | url | None ]
```

I have not tried to reproduce problem elsewhere and I guess it'll work as the cause seems to be boolean instead of window.id. Window is created from js like this:

```
window.open('view_more_data.jsp?id_call=1234','view_more_data','location=0, menubar=0, directories=0, toolbar=0,status=0, width=600, height=410, scrollbars=yes, resizable=yes')
```

HelioGuilherme66:
In my test the window name is set onload.

## See below my tests:

testing.html

<HTML>
<HEAD>
<TITLE>Main Page</TITLE>

<script language="JavaScript" type="text/javascript">
function makeFrame() {
   ifrm = document.createElement("IFRAME");
   ifrm.setAttribute("src", "file:./view_more_data.html");
   ifrm.setAttribute("name","view_more_data" );
   ifrm.style.width = 640+"px";
[...]

PR #272, opened 2014-02-04
Possible workaround for #270

ahitrin:
Hello!

I try to move from SeleniumLibrary to Selenium2Library in tests at my work. The biggest problem I've faced on this way is an impossibility to select some dialogs created by our web system. When a dialog opens, words `Get Window Titles`, `Get Window Identifiers` and `Get Window Names` causes an internal `TypeError`. This bug is 100% reproducible on some dialogs when running tests with Firefox 24.2.0 (I did not test on other browsers).

I've investigated that this bug appears when `WebDriver` returns `WebElement` as window id. It looks like this (just a piece of debug log, when I've added logging into `get_current_window_info`):

> ...
> WARN    [None, u'', u'YC4EmApxD9_Opms.Very Temp_\u041e\u0431\u0437\u0432\u043e\u043d', None]
> WARN    [None, u'', u'\u0424\u043e\u0440\u043c\u0430_ix5EOTKF', None]
> WARN    [<selenium.webdriver.remote.webelement.WebElement object at 0x2a90390>, u'{"pos":{"x1":427,"y1":295,"w":0,"x2":426,"h":0,"y2":294},"win":{"width":429,"height":377}}', u'\u0414\u043e\u0431\u0430\u0432\u043b\u0435\u043d\u0438\u0435 \u0430\u0442\u0440\u0438\u0431\u0443\u0442\u0430 \u0438\u0437 \u043a\u043b\u0438\u0435\u043d\u0442\u0430', None]

The last record appears right before the fail.

So, I've added some code into this function in order to handle `WebElement` correctly. Now none of our dialogs produces this error, and all library unit tests still pass.

### DIFF
--- a/src/Selenium2Library/webdrivermonkeypatches.py
+++ b/src/Selenium2Library/webdrivermonkeypatches.py
@@ -1,6 +1,7 @@
 import time
 from robot import utils
 from selenium.webdriver.remote.webdriver import WebDriver as RemoteWebDriver
+from selenium.webdriver.remote.webelement import WebElement
 from locators import WindowManager
 
 class WebDriverMonkeyPatches:
@@ -21,9 +22,15 @@ class WebDriverMonkeyPatches:
         return self.current_window_handle
 
     def get_current_window_info(self):
-        atts = self.execute_script("return [ window.id, window.name, document.title, document.url ];")
-        atts = [ att if att is not None and len(att) else 'undefined'
-            for att in atts ]
+        raw_atts = self.execute_script("return [ window.id, window.name, document.title, document.url ];")
+        atts = list()
+        for att in raw_atts:
+            if isinstance(att, WebElement):
+                atts.append(att.id)
+            elif att is not None and len(att):
+                atts.append(att)
+            else:
+                atts.append('undefined')
         return (self.current_window_handle, atts[0], atts[1], atts[2], atts[3])
 
     def get_page_source(self):


pr0chz:
Seems to me like a duplicate of #270

ahitrin:
Well, I have to agree. I looked for similar issues before submitting the new one, but didn't recognize #270.

Thank you for mentioning!

ahitrin:
@ombre42 @emanlove could you please take a look on this patch? Is it appropriate for you? Do I need to add some tests on this? Do I need to add something else? Is there anything confusing or discouraging in my code? Please give me an answer :)

ombre42:
I decided to pass through whatever window.id was except if that if it is None, then return 'undefined'.
Checking if it was a WebElement and returning its id if it was did make sense to me, so I did not do that. Could you explain why that is appropriate?
Let me know if you think what I implemented in #336 is not a good solution.
I never understood where this code came from and can only guess it was from the old RC days and never applied to WebDriver or even SeleniumBackedWebDriver.

FYI your code would have still failed if someone assigned a numeric data type to window.id

ahitrin:
Well, this change was made a long time ago, so now I don't even remember why did I use `att.id` for `WebElement`s. Most possible, I did it was because such solution _just worked_ in my test suite that I used to probe Selenium2Library. I trust you when you say that my change could cause new issues because I've never analyzed it deep enough and don't know about possible corner cases.

If you feel/know that fix in #270 is better than this one, please **feel free to close** my pull request. I will try to spend some time on the next week running **your** fixes on my test suite. I'll let you know whether I find something interesting here.

ombre42:
Thanks for responding. If you can figure out a good reason for extracting the WebElement's id, I would be happy to do that. I am just not comfortable putting code in I do not understand the purpose of.
Your code would not have caused new issues, but did not address all possible types that could come over the wire. Thanks again for your help.

window.id is only used by Get Window Identifiers. You can't select a window using them. It has been this way since SeleniumLibrary 2.2.1.

ahitrin:
Sorry for delay. As I see, you've already fixed issue #270, so this pull request should be rejected. Thank you for reviewing!

Issue #246, opened 2013-10-29
Add keywords for Additional Waits 


EricZeff:
Having keywords listed below would make syncing with the application easier.

It seems when issue 174 https://github.com/rtomac/robotframework-selenium2library/issues/174 was closed and the proper issue was address but there should have been a new issue opened for the additional waits (see list below)
- [ ] Element Should Be Clickable
- [ ] Element Should Not Be Clickable
- [ ] Element Should Be Selected
- [ ] Element Should Not Be Selected
- [x] Element Should Not Contain #373 
- [ ] Element Text Should Not Be
- [ ] Wait Until Element Contains
- [ ] Wait Until Element Is Clickable
- [ ] Wait Until Element Is Not Clickable
- [ ] Wait Until Element Is Selected
- [ ] Wait Until Element Is Not Selected
- [x] Wait Until Element Is Visible
- [x] Wait Until Element Is Not Visible
- [ ] Wait Until Element Does not Contain
- [ ] Wait Until Page Does Not Contain
- [x] Wait Until Page Does Not Contain Element #375
- [ ] Wait Until Title Is
- [ ] Wait Until Title Is Not
- [ ] Wait Until Title Contains
- [ ] Wait Until Title Not Contains

just-be-dev:
`Not Contains` is improper English. I recommend `Does Not Contain` as an alternative.

aaltat:
There has not been activity for this issue for few long time and we are planning to close the old issues if they are not anymore valid ones. If you feel that this is still valid issue, could you send an update. If update is not received in a month, we will assume that this is not anymore an issues and we will close this issue.

CaptainRatax:
Hi, sorry to bother, but the keyword "Wait Until Title Is" or " Wait Until Title Is Not" would be really helpful in some cases.

PR #285, opened 2014-03-31
Added Wait Until Not Visible keyword

deiga:


ombre42:
@deiga Thanks for the contribution. I have two concerns, neither of which you need to address.
1) The Java port has this keyword already, but with a slightly different name (Wait Until Element Is Not Visible). I would suggest using the same unless there is a reason to deviate.
2) I would recommend using the phrase "not visible" in the error messages and the docstring instead of "hidden" to be consistent with the keyword name.

deiga:
@ombre42 Okay, I addressed your concerns, merge please :)

ombre42:
@diega sorry I haven't had time to take care of this. I will try to do this soon.

ombre42:
@emanlove , @j1z0 
As we add more and more waiting keywords (see #246 ), the see also part of the docstring is becoming cumbersome.
Should we keep going with it? I don't have strong feelings on this, but if we want to keep cross-references, I would like to do this programatically. I already have working (but esoteric) code:

```
def update_docstrings(_locals):
    see_also = "        See also %(kws)s and BuiltIn keyword `Wait Until Keyword Succeeds`."
    wait_kws = dict([(x, y) for x, y in _locals.iteritems() if x.startswith("wait_")])
    for name, meth in wait_kws.iteritems():
        other_kws = ["`%s`" % x.replace("_", " ").title() for x in wait_kws if name != x]
        this_see_also = see_also % {"kws": ", ".join(sorted(other_kws))}
        meth.__doc__ += "\n" + "\n        ".join(textwrap.TextWrapper().wrap(this_see_also))
```

emanlove:
@ombre42: We might be able to group those keywords or summarize them in the documentation. One thing I wanted from the RF doc tools was to be able to group the documentation by files so instead of alphabetized oreder across the library it would be sectioned by files/directories.

ombre42:
@emanlove I converted the introduction titles names to section title names. This allows us to link to a specific section in the introduction. When I submit a new pr for this, I will add a section to the introduction called _Synchronization_.

pekkaklarck:
@emanlove: The plan is to add tagging support for keywords in RF 2.9:
https://github.com/robotframework/robotframework/issues/925

Part of that enhancement is somehow listing keywords based on tags in Libdoc outputs.

Another related Libdoc enhancement is adding some kind of search functionality based on kw name, doc, and obviously also tags. Search could actually be so easy to implement that it could perhaps be implemented, without tags, already in RF 2.8.6.

PR #286, opened 2014-04-02
Update Selenium2Library.html

anuyens:
Add expiry in optional part in documentation of add Cookie.


emanlove:
Looking into the diff I see your change

```
Adds a cookie to your current session. "name" and "value" are required, "path", "domain", "secure", and "expiry" are optional.
```

But I am noting that the expiry variable is not currently used within the 'Add cookie' keyword. Reviewing keyword history and the expiry property of the underlying add_cookie webdriver method...

just-be-dev:
Closing in favor of #454.

PR #288, opened 2014-04-14
updated '_waiting.py' file

anilreddy:
added wait until element is not visible keyword will add acceptance test soon
added a highlight element keyword and test for it

image


anilreddy:
added a highlight element keyword and test

just-be-dev:
Before we can do anything with a PR it needs to have working tests and be passing the Travis build. 

Secondly, changes should be broken up into logical changes. These two keywords seem unrelated and it doesn't seem like they should be lumped into the same PR.

anilreddy:
Hi @zephraph First of all build failure is not due to these changes and for the highlight element keyword I have added a test case. As recently 'Wait Until Element Not Visible' has been added. Just merge the 'Highlight Element' keyword. Please let me know if I need to do anything.

just-be-dev:
I think the failure is just because your PR is out of date. (For more information how you can update your fork, see [this](https://help.github.com/articles/syncing-a-fork/) article). 

As far as your highlight element keyword goes, I'm not sure that it adds enough substantial value to the library to be added. The reason I say that is that it's somewhat of a niche functionality that might not be useful in all cases. What about when there's a yellow background? 

With that being said, I do want to help facilitate this functionality. It seems to me that to get this functionality we need two things: to be able to call Execute Javascript with arguments and to be able to get a WebElement just using a locator. Execute Javascript with args has been discussed in #323 but we haven't come up with a solution yet. I'll try to revive that conversation. As for getting webelements, that's coming soon! See #356. 

As a temporary solution I would simple suggest using Execute Javascript and locating the element manually with javascript.

just-be-dev:
Hey @anilreddy, just as a follow up, I am trying to make progress on the things I discussed yesterday. 

I created a PR (#387) to address using arguments with `Execute Javascript`. If that PR and #356 gets merged in you'll be able to implement the same functionality in your PR like this....

``` robotframework
*** Keywords ***
Highlight Element
    [Arguments]    ${locator}
    ${element}=    Get WebElement    ${locator}
    Execute Javascript   ${element}
    ...   element = arguments[0];
    ...   original_style = element.getAttribute('style');
    ...   element.setAttribute('style', original_style + "; background: yellow;");
    ...   setTimeout(function(){
    ...        element.setAttribute('style', original_style);
    ...   }, 300);
```

PR #296, opened 2014-06-03
Update browsercache.py

jennifer0703:
It will have the problem “[ WARN ] Keyword ‘Capture Page Screenshot’ could not be run on failure: No browser is open” using browser.quit().

### DIFF
--- a/src/Selenium2Library/utils/browsercache.py
+++ b/src/Selenium2Library/utils/browsercache.py
@@ -20,7 +20,7 @@ class BrowserCache(ConnectionCache):
     def close(self):
         if self.current:
             browser = self.current
-            browser.quit()
+            browser.close()
             self.current = self._no_current
             self._closed.add(browser)
 


emanlove:
@jennifer0703: I understand what you are trying to do here which is capture the screenshot before you shutdown the webdriver. But this needs to be done using the keywords and calling them in a fashion that allows for this not by refusing to close the webdriver process.  This change is a major problem as this is the point, and only point, when we close the webdriver process. If we don't do this you will have multiple executables left after the test and in the case of IEDriverServer this is a really bad idea. As such I am closing this pull request. I will open a issue ticket for documenting how to better handle test teardown so one can capture screenshots for closing the browser and shutting down the webdriver process.

ombre42:
@jennifer0703 Are you using Close Window as part of a teardown? That will throw the No browser open error if the current browser was closed (possibly by closing the last open window). Use Close Browser or Close All Browsers for teardown.
quit() is what should be called as Ed mentioned, as it should cleanup where close(), even on the last window, does not  try to cleanup the spawned server process.

PR #297, opened 2014-06-06
Added new keyword 'Clear Element Text'

emanlove:
Added 'Clear Text' keyword for clearing the text value of a text entry element.

### DIFF
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -248,7 +248,7 @@ class _ElementKeywords(KeywordGroup):
         """
         return self._get_text(locator)
 
-    def clear_text(self, locator):
+    def clear_element_text(self, locator):
         """Clears the text value of text entry element identified by `locator`.
 
         See `introduction` for details about locating elements.
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -254,8 +254,7 @@ class _ElementKeywords(KeywordGroup):
         See `introduction` for details about locating elements.
         """
         element = self._element_find(locator, True, True)
-        if element is not None:
-            element.clear()
+        element.clear()
 
     def get_vertical_position(self, locator):
         """Returns vertical position of element identified by `locator`.
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -248,6 +248,15 @@ class _ElementKeywords(KeywordGroup):
         """
         return self._get_text(locator)
 
+    def clear_text(self, locator):
+        """Clears the text value of text entry element identified by `locator`.
+
+        See `introduction` for details about locating elements.
+        """
+        element = self._element_find(locator, True, True)
+        if element is not None:
+            element.clear()
+
     def get_vertical_position(self, locator):
         """Returns vertical position of element identified by `locator`.
 
--- a/test/acceptance/keywords/content_assertions.txt
+++ b/test/acceptance/keywords/content_assertions.txt
@@ -175,8 +175,6 @@ TextField Should Contain
     Input Text  name  my name
     TextField Should contain  name  my name
     Run Keyword And Expect Error  Text field 'name' should have contained text 'non-existing' but it contained 'my name'  TextField Should contain  name  non-existing
-    Clear Text  name
-    TextField Should contain  name  ${EMPTY}
 
 TextField Value Should Be
     [Documentation]  LOG 2:4 Content of text field 'name' is ''.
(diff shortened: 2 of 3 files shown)


test/acceptance/keywords/content_assertions.txt
```diff
@@ -175,6 +175,8 @@ TextField Should Contain
     Input Text  name  my name
     TextField Should contain  name  my name
     Run Keyword And Expect Error  Text field 'name' should have contained text 'non-existing' but it contained 'my name'  TextField Should contain  name  non-existing
+    Clear Text  name
+    TextField Should contain  name  ${EMPTY}
```
ombre42: This would not fail regardless of the text content of the element

test/acceptance/keywords/content_assertions.txt
```diff
@@ -191,6 +195,8 @@ TextArea Should Contain
     Run Keyword And Expect Error
     ...  Text field 'comment' should have contained text 'Hello World!' but it contained 'This is a comment.'
     ...  TextArea Should Contain  comment  Hello World!
+    Clear Text  comment
+    TextArea Should Contain  comment  ${EMPTY}
```
ombre42: This would not fail regardless of the text content of the element

ombre42:
This is certainly more readable than:
`Input Text    locator    ${EMPTY}`

Even if I/we do not like the verbiage in a keyword already defined by the Java port, I think we should stay consistent with the port when possible. The Java port has this implemented as 
`Clear Element Text`
What do you think?

emanlove:
It should probably be the same as the Java port. And I will have to fix those tests. Thanks for that catch.

emanlove:
I'm going back and wandering if we can keep it as 'Clear Text' instead of the matching 'Clear Element Text'. Also wandering if it is possible for the java port to have an alternative name (something similar to a python decorator that might give a keyword [method] an alternative name). Thoughts @ombre42, @pekkaklarck @j1z0  ?!?

ombre42:
Could you share why you do not like 'Clear Element Text'? Do you not like the keyword name being chosen for you because another project implemented it first? I could relate to that.
I just want to hear reasons for differing, because you probably have good reasons. The motivation behind trying to be the same was to ease the transition of a test script _or_ author from Python to Java or vice versa. Using plugins or and IDE like RIDE does make the transition easier (highlighting unknown keywords, auto-completion).

I am thinking that if is not worth being consistent with the Java port on this keyword then it is not for others that have yet to be implemented in Python. `Wait Until Title Not Contains` is pretty awkward as far as grammar is concerned.

pekkaklarck:
There are clearly two important and contradicting sides to the larger problem noted by @ombre42:
- Good keyword names and internal consistency is very important. For some reason I also like `Clear Text` a little bit more than `Clear Element Text` and totally agree that `Wait Until Title Not Contains` is awkward.
- Compatibility with the Java port is also extremely important.

Below is how I think would be best to resolve this situation:
- Investigate which new keywords the Java port has and which of them we consider suboptimally named.
  - With keywords that aren't that bad (e.g. `Clear Element Text` could be in this category) it's probably just easiest to use that name.
  - With keywords that are too awkward, it's probably better to just name them better. We could also try to persuade the Java port devs to rename these keywords in future releases (with deprecation).
- Try to enhance communication with these two libraries so that devs on each side are informed about new keywords and changes to existing. Not sure what would be the best way or is this feasible in general. It definitely would be beneficial, though.
- Write down keyword naming conventions for Selenium2Library. This is something where I would like to help. Might even be better to write these guidelines to Robot Framework's wiki and then just refer to them from Se2Lib developer docs.

emanlove:
@ombre42 I have no good reason other than what you both have mentioned; that the name 'Clear Text' is more consistant and "sounds" better then 'Clear Element Text'. I do agree with the strong need to match up with the Java port and will change this to match. @pekkaklarck, I think you have stated all the important points succinctly and are great guildelines. For the reasons you've stated I was hoping to get @MarkusBernhardt onboard with the advisory council I mentioned.

emanlove:
@ombre42: I've reverted this to 'Clear Element Text' and updated test cases. I also removed the invalid checks (good catch) although I am still looking at these trying to figure out why it wasn't working the way I expected it too.

src/Selenium2Library/keywords/_element.py
```diff
@@ -248,6 +248,15 @@ def get_text(self, locator):
         """
         return self._get_text(locator)
 
+    def clear_element_text(self, locator):
+        """Clears the text value of text entry element identified by `locator`.
+
+        See `introduction` for details about locating elements.
+        """
+        element = self._element_find(locator, True, True)
+        if element is not None:
```
ombre42: checking for element being None does not make sense here. You are passing required=True to _element_find.  _element_find will raise an exception and not return None if there are no elements found.

emanlove: Very true. Have removed the check.

PR #298, opened 2014-06-06
Added new keyword 'Locator Should Match X Times'

emanlove:
Added 'Locator Should Match X Times' keyword similar and more generic version of 'Xpath Should Match X Times' keyword and test.

### DIFF
--- a/test/acceptance/keywords/elements.txt
+++ b/test/acceptance/keywords/elements.txt
@@ -25,10 +25,6 @@ Get Matching XPath Count
     ${count}=  Get Matching XPath Count  //div[@id="first_div"]/a
     Should Be Equal  ${count}  2
 
-Locator Should Match X Times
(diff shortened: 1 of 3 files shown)


test/acceptance/keywords/elements.txt
```diff
@@ -25,6 +25,10 @@ Get Matching XPath Count
     ${count}=  Get Matching XPath Count  //div[@id="first_div"]/a
     Should Be Equal  ${count}  2
 
+Locator Should Match X Times
```
ombre42: Shouldn't this be in content_assertions.txt where _Xpath Should Match X Times_ coverage is?

emanlove: @ombre42: I've moved this test over to content_assertions.txt. If you get a chance could you reevalute and possibly pull in this change? Thanks.  - Ed

Issue #281, opened 2014-03-11
'Select Window’ with url strategy fails to locate window


yamateh:
```
def get_current_window_info(self):
    atts = self.execute_script("return [ window.id, window.name, document.title, document.URL ];")
```

  it is 'document.URL',not 'document.url'

ombre42:
Agreed: reference [DOM level 2](http://www.w3.org/TR/DOM-Level-2-HTML/html.html#ID-46183437)
@yamateh what browser/version/OS is this causing issues with for you?

yamateh:
browser:chrome 32
os:win7

yamateh:
interface HTMLDocument : Document {
           attribute DOMString       title;
  readonly attribute DOMString       referrer;
  readonly attribute DOMString       domain;
  readonly attribute DOMString       URL;

laulaz:
I also have this problem on Firefox : 
`document.url` is `undefined` and appears in `get_current_window_info()` values.
So in my robot test, when trying to `Select Window`, I get `AssertionError: ValueError: Unable to locate window with URL` even if the URL is correct.

This commit is the cause of the change : 0bf17bc8bcec8d6f17704e74d5a7341a43dc8d7a. It refers to http://www.w3schools.com/jsref/prop_doc_url.asp, which also talks about `URL` and not `url`.

PR #300, opened 2014-06-30
Fix url retrieval

laulaz:
See http://www.w3schools.com/jsref/prop_doc_url.asp
This refs #281


ombre42:
Re-submitted in #334 and merged. Sorry this took so long.
Thank you for the contribution!

PR #301, opened 2014-07-07
Javascript execution with element located by webdriver as argument

mircoc:
I added a method to execute Javascript with some arguments. The first is identified by webdriver locator and the other are optional and passed as-is.

### DIFF
--- a/src/Selenium2Library/keywords/_javascript.py
+++ b/src/Selenium2Library/keywords/_javascript.py
@@ -91,6 +91,37 @@ class _JavaScriptKeywords(KeywordGroup):
         self._info("Executing JavaScript:\n%s" % js)
         return self._current_browser().execute_script(js)
 
+    def execute_javascript_on_element_with_args(self, code, locator, *args):
+        """Executes the given JavaScript code with first argument as
+        an element specified by `locator` and more optional arguments.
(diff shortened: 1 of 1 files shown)


ombre42:
Thank you very much for the contribution.
This feels like too specific of a use case. If someone wants to do the same thing with _Execute Async Javascript_ or with _Wait For Condition_, we would need 2 more new keywords in addition to this. I think we are going to have Get Webelement(s) keywords soon (#315). If we also add the capability to pass arguments to _existing_ keywords (#323), then  you will be able to do:

```
${elem}=    Get Webelement    xpath=//really//complicated
${args}=    Create List    ${elem}    200
Execute Javascript    arguments[0].scrollTop = arguments[1];    args=${args}
```

This is not nearly as compact as your solution, but it is more generic. It would be one less line if it were not for the fact _Execute Javascript_ has an existing vararg so we cannot add *args :(
What you submitted is good stuff. I am just looking at the long term. Too many keywords can be a bad thing.
Thanks again.

ombre42:
Still not sure how we are going to handle this. See #323 for discussion. Whenever we implement or modify a JS keyword for this, the async part needs to be handled as well.

PR #302, opened 2014-07-14
Update _browsermanagement.py

swordfly:
Enhance Selenium2Library for IE protected model issue. Now robot framework keyword will support desired_capabilities for None remote mode for ie and chrome browsers

### DIFF
--- a/src/Selenium2Library/keywords/_browsermanagement.py
+++ b/src/Selenium2Library/keywords/_browsermanagement.py
@@ -547,7 +547,23 @@ class _BrowserManagementKeywords(KeywordGroup):
         '''most of the make browser functions just call this function which creates the 
         appropriate web-driver'''
         if not remote_url: 
-            browser = webdriver_type()
+#            from selenium.webdriver.common.desired_capabilities import DesiredCapabilities            
+#            caps = DesiredCapabilities.INTERNETEXPLORER
+#            caps['ignoreProtectedModeSettings'] = True
+#            driver = webdriver.Ie(capabilities=caps)
+#            _generic_make_browser(webdriver.Ie, webdriver.DesiredCapabilities.INTERNETEXPLORER, remote, desired_capabilities)
+              
+            ###enhance to enable DesiredCapabilities for None remote mode###
+            desired_capabilities_object = desired_cap_type.copy()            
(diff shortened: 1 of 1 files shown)


src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -547,7 +547,23 @@ def _generic_make_browser(self, webdriver_type , desired_cap_type, remote_url, d
...
+#            _generic_make_browser(webdriver.Ie, webdriver.DesiredCapabilities.INTERNETEXPLORER, remote, desired_capabilities)
+              
+            ###enhance to enable DesiredCapabilities for None remote mode###
+            desired_capabilities_object = desired_cap_type.copy()            
+            if type(desired_caps) in (str, unicode):
+                desired_caps = self._parse_capabilities_string(desired_caps)
+            desired_capabilities_object.update(desired_caps or {})  
+            self._debug(desired_capabilities_object)
+            if "selenium.webdriver.ie.webdriver.WebDriver" in str(webdriver_type):                     
+                browser = webdriver_type(capabilities=desired_capabilities_object)
```
ombre42, line 563: Firefox also uses an init arg of `capabilities`, not `desired_capabilities`. The issue you are trying to fix should be fixed. People should not have to use the awkward `Create Webdriver_`.

src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -547,7 +547,23 @@ def _generic_make_browser(self, webdriver_type , desired_cap_type, remote_url, d
         '''most of the make browser functions just call this function which creates the 
         appropriate web-driver'''
         if not remote_url: 
-            browser = webdriver_type()
+#            from selenium.webdriver.common.desired_capabilities import DesiredCapabilities            
+#            caps = DesiredCapabilities.INTERNETEXPLORER
+#            caps['ignoreProtectedModeSettings'] = True
```
ombre42, line 552: If your goal was to handle "IE protected model issue", I d on't think this works as is, because your code is commented out.

ombre42:
Thanks you for contributing. This PR helps us know that the related open issues are affecting users, but we cannot merge the code as is. Thank you again.

Issue #268, opened 2014-01-23
Make the default file name for 'Capture Page Screenshot' more unique


sdarshanam:
capture_page_screenshot keyword saves screenshot by suffixing index.

Can this be changed to use the current date time instead of index to make it more unique.

```
    if not filename:
        self._screenshot_index += 1
        filename = 'selenium-screenshot-%d.png' % self._screenshot_index
```

ombre42:
Capture Page Screenshot takes a filename argument so you can have any naming scheme you want. To control the filename when there is a failure you can do something similar to the following:

```
Register Keyword To Run On Failure    Save Selenium Screenshot

*** Keywords ***
Save Selenium Screenshot
    ${screenshot_index}=    Get Variable Value    ${screenshot_index}    ${0}
    Set Global Variable    ${screenshot_index}    ${screenshot_index.__add__(1)}
    ${time}=    Evaluate    str(time.time())    time
    Capture Page Screenshot    selenium-screenshot-${time}-${screenshot_index}.png
```

The reasoning behind still using an index is that time.time is not guaranteed to be more granular than a second (on Windows it is). Two screenshots within a second is possible and we do not want one to overwrite the other.

If you are requesting that the default filename be more unique, than please say so.

sdarshanam:
Thanks Kevin. I will try this out. Nevertheless, can the default file name made more unique.

ombre42:
@sdarshanam I think this would be a good idea. I believe a lot of people are re-inventing the wheel out there.
Besides datetime, another option to consider would be using a UUID, like Selenium does for a session ID.
The only concern would be breaking existing tests due to the path being too long.

sdarshanam:
Do we really need 'selenium-screenshot-' as part of the file name? we already know its a png (screenshot) and its from selenium, moreover its a default file name, so why not just have UUID as file name?

mulmschneider:
I've created a pull request for more unique screenshot names, but a quick workaround is to just redefine the 'Run on Failure' keyword:

```
*** Settings ***
Library           Selenium2Library
Library           uuid

*** Keywords ***
Init
    Register Keyword To Run On Failure  Capture Page Screenshot With Unique Name

Capture Page Screenshot With Unique Name
    ${uuid} =  uuid4
    Capture Page Screenshot  selenium-${uuid}.png
```

aaltat:
The current code in master, which is not yet released allows more configuration on the file name. Therefore closing the issue. 

[...]

PR #309, opened 2014-08-04
Make screenshot names more unique by default

mulmschneider:
As discussed in #268

### DIFF
--- a/src/Selenium2Library/keywords/_screenshot.py
+++ b/src/Selenium2Library/keywords/_screenshot.py
@@ -1,6 +1,7 @@
 import os
 import robot
 from keywordgroup import KeywordGroup
+import uuid
 
 class _ScreenshotKeywords(KeywordGroup):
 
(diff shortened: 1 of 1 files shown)


ombre42:
I like the idea but have two concerns.
- I think the index code should be removed if using UUIDs. I do not see what value it provides anymore.
- I would like to shorten the name up a bit to reduce likelihood of hitting maximum path length - I have had issues when using Jenkins and I want to avoid breaking people's code as much as possible. How about `filename = 'selenium-%s.png' % uuid.uuid4().hex`.

So for example
`selenium-bae4c6d5794d4bb29c0f39547a380504.png`
instead of
`selenium-screenshot-4a39cca9-b4f5-43a7-83b2-cc308b50b7b9.png`

emanlove:
Beyond @ombre42's concern of breaking others code due to filename length, my concern is how this change could break others code in general. I've seen several requests for this change and I am in favor of it. But [Eliza Loeng's note back in December](https://groups.google.com/d/msg/robotframework-users/jw-YJX8W35A/D5FA-WifnMEJ) has really struck accord with me about trying not to break people's code with changes to Selenium2Library (although it is about browser versions working with the current selenium/Selenium2Library release). This why I mentioned my desire to have an advisor group of users so we could ask "Hey, if we make this change how much will it affect you?" and get feedback from the larger Selenium2Library community.

Just some food for thought...

emanlove:
Closing as no feedback on suggested changes was given. If someone wants reopen this pull request, please do so.

PR #310, opened 2014-08-04
Add new keyword get_log

mulmschneider:
Add new keyword providing access to the selenium logs

### DIFF
--- a/src/Selenium2Library/keywords/_browsermanagement.py
+++ b/src/Selenium2Library/keywords/_browsermanagement.py
@@ -311,6 +311,14 @@ class _BrowserManagementKeywords(KeywordGroup):
         """Returns the entire html source of the current page or frame."""
         return self._current_browser().get_page_source()
 
+    def get_log(self, log_type):
+        """Get the log for a given selenium log type
+
+        The `log_type` argument defines which logs to get. Possible values are:
+        `browser`, `driver`, `client` or `server`
+        """
+        return self._current_browser().get_log(log_type)
+
     def get_title(self):
         """Returns title of current page."""
         return self._current_browser().get_title()


emanlove:
@ombre42 , @j1z0, @MarkusBernhardt: Any thoughts on this one ?

pekkaklarck:
Looks potentially useful for me. The doc should perhaps at least briefly explain log formats, though, possibly with a link to Selenium docs. I'd also name this keyword to more explicit `Get Selenium Log` or  something similar.

MarkusBernhardt:
Sounds like a good idea, but what means "access" exactly?

ombre42:
I think it is a good idea, but there are several issues here.

This functionality is still a work in progress from what I gather, which I think _must_ be mentioned. This would help set expectations. In Java, [WebDriver.Options.logs()](http://selenium.googlecode.com/git/docs/api/java/org/openqa/selenium/WebDriver.Options.html#logs%28%29) is annotated `@Beta` and the [wiki page](https://code.google.com/p/selenium/wiki/Logging) says "still under work". The method used here is not available until 2.35.0, which is newer than setup.py requires. I think this should be noted in the docstring or the minimum version should be increased.

What log types are supported depends on the driver being used. I think a keyword exposing [log_types](http://selenium.googlecode.com/git/docs/api/py/webdriver_remote/selenium.webdriver.remote.webdriver.html#selenium.webdriver.remote.webdriver.WebDriver.log_types) should be implemented at the same time as `Get Log`. If I call `get_log` with an invalid log type, it returns an empty list and does not raise an exception. I think the assumption is you should only send a `log_type` that the `log_types` attribute returns.
Perhaps we should automatically check `log_types` and fail if the `log_type` requested is not supported.

The return type needs to be documented. It is a list. In my quick test, it was a list of dict, ~~but the wire protocol only specifies `{Array.<object>}`, so it may not always be a list of dict~~.

I agree with Pekka - the name should be a little less generic.

On my machine at the moment, Firefox returns only `browser` and `driver` for `log_types`. The IEDriverServer I am using does not even implement /session/:sessionId/log or  /session/:sessionId/log/types.

There should be at least one acceptance test.

ombre42:
After taking a second look, there is a guarantee of the returned structure, which should be documented. It will always be a list of dict with these keys:
level, message, and timestamp.
More here: https://code.google.com/p/selenium/wiki/JsonWireProtocol#Log_Entry_JSON_Object

pekkaklarck:
Thanks @ombre42 for detailed analysis! Some comments related to it:
- I don't think it makes sense to raise minimum required Selenium version because of this alone. Mentioning the minimum version in keyword doc ought to be enough. Obviously there could be other reasons for rising the version requirement, though.
- I don't know detailed use cases for this functionality, but I would assume most users would prefer getting logs as a string. Should there perhaps be an argument to select the return type?
- Many users would probably want to log the info to Robot's log for manual analysis. Should the keyword do it automatically, with a configured log level, or should there be separate `Log Selenium Log` keyword?

ombre42:
- I agree we should not increase the minimum version
- I do not know the valid use cases either. Even if we only support string or as-is, it would probably be a good idea if we use a pseudo-enum argument in case we want to add XML later, etc.
- How about adding a loglevel argument and log the log at that level, e.g. `get_selenium_log(log_type, return_type=STRING, loglevel='NONE')`?

mulmschneider:
Regarding use case: I'm using this to log as much information as possible on failures, since I have a few unreliable tests that only fail once in a while (without any discernible reason). 
This is the code that I'm using: 

```
....
Register Keyword To Run On Failure  Log lots of selenium data
....

Log lots of selenium data
    ${uuid} =  uuid4
    Capture Page Screenshot  results/selenium-${uuid}.png
    ${browser_log} =  Get Log  browser
    Log  ${browser_log}
    Log Source  TRACE
```

just-be-dev:
@mulmschneider, thanks for the PR! Sorry this issue has been idle for so long. Would you still like to see this in the library? If so, please write an acceptance test for this keyword and rebase to master.

aaltat:
@mulmschneider, Would you still like to see this in the library? If so, please write an acceptance test for this keyword and rebase to master.

SergiuTudos (commented):
I agree we need to merge this one.

aaltat:
Because of inactivity, closing the PR.

PR #317, opened 2014-08-22
Fixed build, added support for negative indexes in Robot keywords

eweitz:
This pull request fixes the rtomac:master build and adds support for negative indexes in Robot keywords.
# Fixed Se2Lib build

@emanlove, this fixes the build failure currently noted at https://github.com/rtomac/robotframework-selenium2library.

Passing Travis CI build:
- https://travis-ci.org/rtomac/robotframework-selenium2library/jobs/33288022

The 14 critical test failures reported in the current rtomac:master build (https://travis-ci.org/rtomac/robotframework-selenium2library/jobs/23423255) are caused by unexpected content in test/results/output.xml.  An extra line was added in the log output of a frequently-used command -- likely due to an update in the upstream Selenium 2 package.  

The failing Se2Lib tests simply needed to account for the fact that the expected log file content is now a few lines ahead of where it was previously.

I also applied a fix for the 'Choose File' bug by @HelioGuilherme66 (from 532295cbce).
# Support for negative indexes in Robot keywords

Robot keywords like `Table Row Should Contain` support looking up elements with a positional, 1-based index, e.g.

```
Table Row Should Contain   ${table_locator}    1    ${expected_first_row}
```

verifies that the first row (row 1) in the table with the ID defined in 'table_locator' has the expected content defined in 'expected_first_row'.

Unfortunately, `Table Row Should Contain` does not currently support easily looking up content from the end of a table. For example, it would be very helpful to be able to write something like

```
Table Row Should Contain   ${table_locator}    -1    ${expected_last_row}
```

to verify the content of the last row (row -1) in a table.  Row -2 would be the second-to-last row, and so on -- just like negative indexes in Python lists.

This pull request adds support for negative indexes to the following Robot keywords:
- `Table Row Should Contain`
- `Table Column Should Contain`
- `Table Cell Should Contain`

### DIFF
--- a/src/Selenium2Library/keywords/_tableelement.py
+++ b/src/Selenium2Library/keywords/_tableelement.py
@@ -16,23 +16,28 @@ class _TableElementKeywords(KeywordGroup):
         """Returns the content from a table cell.
 
         Row and column number start from 1. Header and footer rows are
-        included in the count. This means that also cell content from
-        header or footer rows can be obtained with this keyword. To
-        understand how tables are identified, please take a look at
-        the `introduction`.
+        included in the count. A negative row or column number can be used
+        to get rows counting from the end (end: -1). Cell content from header 
+        or footer rows can be obtained with this keyword. To understand how 
+        tables are identified, please take a look at the `introduction`.
         """
         row = int(row)
-        row_index = row - 1
+        row_index = row
+        if row > 0: row_index = row - 1
         column = int(column)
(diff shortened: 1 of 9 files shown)


ombre42:
Thenk you so much Eric.
Green?

eweitz:
@ombre42, glad to help.  Is there anything preventing this from being merged into rtomac:master?

Issue #312, opened 2014-08-07
Rename acceptance tests to use `.robot` extension instead of `.txt`


pekkaklarck:
This makes it easier to distinguish between test case files and possible normal text files. In practice that means that many RF editor plugins will recognize them automatically, and also GitHub will syntax highlight them.

This change was recently done for Robot Framework itself (robotframework/robotframework#1766). That required more work because we also wanted to convert remaining HTML files to plain text format and clean-up test data in general. For an example file that is now syntax highlighted see https://github.com/robotframework/robotframework/blob/master/atest/testdata/standard_libraries/xml/add_and_remove_elements.robot

HelioGuilherme66:
I have renamed tests in this commit: https://github.com/HelioGuilherme66/robotframework-selenium2library/commit/c264f5fa5af1bec21d7a66f3f4999acd499f3d52 (this commit also changed travis.yml to pass tests).

You could use only the shell script in the end of the commit.

just-be-dev:
I'm very much in favor of this. I'll try to target this after the 1.7 release.

PR #326, opened 2014-09-01
various small fixes (headers, caps, tabs, typo...)

laurentbristiel:


ombre42:
I'm a little confused about this pull request. Robot Framework is executing our tests as intended, so I do not see how this fixes anything. You can use a single or multiple asterisks, "setting" or "settings", and for data separation, two spaces or four works just as well. This project does not have guidelines for test file syntax as far as I know. What value does this add to the project? I do not want to discourage you from contributing to this project, though. You contribute a lot to the community.

laurentbristiel:
Agreed. Does not add value to the project. I was just reading the tests and made those small changes to be more homogeneous with what I see elsewhere in Robot Tests. If this adds more noise than signal to the project, I am fine with your rejecting the pull request.

pekkaklarck:
Personally I value consistency pretty highly but agree that a precondition for that would be having guidelines. I hope I have time to write such guidelines to Robot Framework itself this autumn then other projects could refer to those guidelines if they wanted to.

A concrete reason why to use consistent headers is that some Robot plugins (at lest the PyCharm plugin I use) don't recognize all possible variations. That can be argued to be a bug in such plugins, though.

ombre42:
When we implement #312 we could define standards and re-format as we rename.
Mainly my concern is there is a lot of impacting issues that were reported a long time ago and not yet fixed. Until the TODO list gets smaller, I do not want to worry about format issues.
We do not even have a coding format standard :)

pekkaklarck:
@ombre42 Totally agreed, formatting changes to tests are good to be done if/when taking `*.robot` extension to use.

PR #329, opened 2014-09-24
add locator prefix 'partial link' for find_elements_by_partial_link_text

lina1:
add a method '_find_element_by_partial_text' in elementfinder.py and a test function for it.


ombre42:
@lina1 thanks for contributing
You do not need to do any more, but here are some TODOs with this PR:
- Update the introduction
- Make the prefix _partiallink_ work as well. This is the first locator prefix to have more than one word. _parse_locator does a strip and a lower case, but the lookup would still fail if someone didn't have the space in the middle. This is just my opinion

PR #332, opened 2014-09-26
Allow spaces in locator prefix

ombre42:
@emanlove Do you see any problem with doing this?
The motivation was to avoid failing for a simple mistake like not having a space between partial and link
partial link=text
partiallink=text

### DIFF
--- a/src/Selenium2Library/locators/elementfinder.py
+++ b/src/Selenium2Library/locators/elementfinder.py
@@ -4,14 +4,14 @@ from robot.api import logger
 class ElementFinder(object):
 
     def __init__(self):
-        self._strategies = {
+        self._strategies = {  # keys must be lower-case w/ no spaces
             'identifier': self._find_by_identifier,
             'id': self._find_by_id,
             'name': self._find_by_name,
             'xpath': self._find_by_xpath,
             'dom': self._find_by_dom,
             'link': self._find_by_link_text,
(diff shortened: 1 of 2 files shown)


ombre42:
I don't like how this was implemented.
I want to just use a case and space-insensitive dict.
@pekkaklarck is there any reason to not use robot.utils.NormalizedDict in other projects like this?

pekkaklarck:
@ombre42 `robot.utils.NormalizedDict` ought to be safe to use. It's used so widely in Robot code base that I doubt it will ever be removed. In general the comment in [robot.utils API docs](https://robot-framework.readthedocs.org/en/latest/autodoc/robot.utils.html#module-robot.utils) is pretty accurate:

>  Utilities are mainly for internal usage, but external libraries and tools may find some of them useful. Utilities are generally stable, but absolute backwards compatibility between major versions is not guaranteed.

There generally shouldn't be backwards incompatible changes between minor releases, and with major releases they ought to be limited and typically noticed during pre-release period.

ombre42:
Thank you so much! Thank you for the link and sorry I didn't do my homework first.

pekkaklarck:
@ombre42 absolutely no problem! Because API docs only say that changes are _avoided_, it would have anyway been a good idea to ask there changes to do something to `NormalizedDict`.

PR #335, opened 2014-10-01
add default prefix and allow prefixes with spaces

### LINKED ISSUE #193: Cannot use default strategy when a '=' is in the locator
The following code will fail because ElementFinder assumes '=' separates a prefix from the value and interprets 'detail?name' as a locator prefix:

```
Open Browser    https://code.google.com/p/robotframework/downloads/list?can=1&q=
Page Should Contain Element    xpath=//a[@href='detail?name=robotframework-2.7.7.win32.exe&can=1&q=']
Click Link    detail?name=robotframework-2.7.7.win32.exe&can=1&q=
```

actual error:
ValueError: Element locator with prefix 'detail?name' is not supported

This forces the user to use a specific strategy. If a prefix 'default' were added, then the user could do this as a workaround:

```
Click Link    default=detail?name=robotframework-2.7.7.win32.exe&can=1&q=
```

Other options include:
- provide to provide a way to escape '='s.
- use the default strategy when a locator prefix is not recognized rather than raising an error

ombre42:
This is a change I would like peer-reviewed. If the default strategy changes are rejected, I can submit a separate request to just allow prefixes with spaces.
Allows you to use the default strategy (match key attributes) when a '=' is in the locator.
Locator `page?a=b` will still cause error, but can be changed to `default=page?a=b`.
Locator prefixes will now be space-insensitive so the following will work when the previously did not:
- `partiallink=text`
- `x path = //div`

Fixes #193

### DIFF
--- a/test/unit/locators/test_elementfinder.py
+++ b/test/unit/locators/test_elementfinder.py
@@ -12,6 +12,10 @@ class ElementFinderTests(unittest.TestCase):
             self.assertRaises(ValueError, finder.find, browser, "something=test1")
         except ValueError as e:
             self.assertEqual(e.message, "Element locator with prefix 'something' is not supported")
+        try:
+            self.assertRaises(ValueError, finder.find, browser, " by ID =test1")
+        except ValueError as e:
+            self.assertEqual(e.message, "Element locator with prefix ' by ID ' is not supported")
 
     def test_find_with_null_browser(self):
         finder = ElementFinder()
@@ -35,6 +39,22 @@ class ElementFinderTests(unittest.TestCase):
         browser = mock()
         finder.find(browser, "test1")
         verify(browser).find_elements_by_xpath("//*[(@id='test1' or @name='test1')]")
+
+    def test_find_with_explicit_default_strategy(self):
+        finder = ElementFinder()
+        browser = mock()
+        finder.find(browser, "default=test1")
+        verify(browser).find_elements_by_xpath("//*[(@id='test1' or @name='test1')]")
+
+    def test_find_with_explicit_default_strategy_and_equals(self):
+        finder = ElementFinder()
+        browser = mock()
+        when(browser).get_current_url().thenReturn("http://localhost/mypage.html")
+        finder.find(browser, "default=page.do?foo=bar", tag='a')
+        verify(browser).find_elements_by_xpath(
+            "//a[(@id='page.do?foo=bar' or @name='page.do?foo=bar' or @href='page.do?foo=bar' or " +
+            "normalize-space(descendant-or-self::text())='page.do?foo=bar' or " +
+            "@href='http://localhost/page.do?foo=bar')]")
 
     def test_find_with_tag(self):
         finder = ElementFinder()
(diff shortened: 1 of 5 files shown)


ombre42:
An alternate solution I have already mentioned is to not raise an error if the prefix is not found in our list, but fail over to the default strategy using the entire value of  `locator`.

test/unit/locators/test_elementfinder.py
```diff
@@ -12,6 +12,10 @@ def test_find_with_invalid_prefix(self):
             self.assertRaises(ValueError, finder.find, browser, "something=test1")
         except ValueError as e:
             self.assertEqual(e.message, "Element locator with prefix 'something' is not supported")
+        try:
+            self.assertRaises(ValueError, finder.find, browser, " by ID =test1")
+        except ValueError as e:
+            self.assertEqual(e.message, "Element locator with prefix ' by ID ' is not supported")
```
pekkaklarck: This looks strange. Doesn't `assertRaises` catch the `ValueError` so `except ValueError` block should never be executed? That block seems to test the actual error message. If you want to accomplish that, you can either use the standard [assertRaisesRegexp](https://docs.python.org/2/library/unittest.html#unittest.TestCase.assertRaisesRegexp) (new in Python 2.7), Robot's [assert_raises_with_msg](https://robot-framework.readthedocs.org/en/latest/autodoc/robot.utils.html#robot.utils.asserts.assert_raises_with_msg), or just execute the code failing code normally and actually catch the error with `except ValueError`.

pekkaklarck:
I think adding `default=` prefix that can be used when the actual locator contains `=` is a good idea. The motivation could probably be explained a bit better in the documentation, though.

test/unit/locators/test_elementfinder.py
ombre42: Good catch! That was copypasta on my part - the unit test was broken before my changes. It looked funny to me too, but my mind was elsewhere. Decided to use the utility method due to shortcomings in 2.6 you mentioned. More readable than standard catch. Thanks!

ombre42:
@pekkaklarck Thanks for your input. I'm glad you think its a good idea - I was a bit on the fence as this is rarely needed and easily worked around.
I have implemented the changes you suggested. There are conflicting changes probably due to CHANGES.rst so let me know if you think it is ready to be merged and I will do the actually merging.

pekkaklarck:
Looks good to me.

ombre42:
@pekkaklarck Thanks again. Manual merge is a little painful. Going to try and avoid that in the future.

emanlove:
Sorry for not chiming in sooner. My thoughts go to the question of trying not to have too many similar locators. This might not be an issue but I would prefer not to see a lot of find exact match and find partial match for links, images, etc.  As I mentioned in the [post of SmartClient/SmartGWT](https://github.com/rtomac/robotframework-selenium2library/issues/333#issuecomment-57914614) locators possible having a locator plug and play code...

Really just need some time to focus on this and get my thoughts in order.

ombre42:
@emanlove you have a point. The purpose of this project is to make testing with Selenium easier. The reason I found adding partial link acceptable was because it can improve the readability of a test and is easier than the alternative.
Compare
`Click Link    partial link=#470`
to
`Click Link    xpath=//a[contains(.,'#470')]`
I see your point though, that adding this locator may not be worth muddying things up.
If you want to remove this strategy I won't object.

emanlove:
No, go ahead and leave it in there. Someone saw a need, your code made a good fix, and a problem was solved. Keep it in.

Issue #119, opened 2012-09-24
Need Mouse Move keyword


emanlove:
Seen need for mouse move keyword.

adwu73:
second this thought.

just-be-dev:
Looking at the API it looks like we would have three keywords
- `Move Mouse by Offset`
- `Move Mouse to Element`
- `Move Mouse to Element with Offset`

emanlove:
Seeing that I opened this ticket a while ago and there has not been a lot of rumbling or requests to implement I not sure of the need. Similar to the discussion raised within #344 and elsewhere about keyword overload I question whether we should create these three keywords just for the reason of creating them.

I think @aaltat's point under #344 that some of these additional keywords could actually be implemented using existing keywords but with additional optional arguments is a good one.  We just need to be careful and not create backward compatibility issues or future blockers.

just-be-dev:
The only reason I suggested these is because they directly translate from the Selenium API. It'd pretty much be as simple as `driver.move_by_offset` `driver.move_to_element` `driver.move_to_element_with_offset`. 

See [here](http://selenium-python.readthedocs.org/en/latest/api.html#selenium.webdriver.common.action_chains.ActionChains.move_by_offset).

aaltat:
Although these could be useful keywords, no-one has requested such keywords for a long time. Therefore closing this issue.

el-nuru:
I also would like to have those keywords

m3s5ts:
I need exactly what Click Element at Coordinates provides, but _without_ action.click(); i.e. Move Mouse to Element with Offset. Unfortunately at this time I'm not able to modify the source myself and will end up writing a python script, which is not a desired workaround considering the functionality is available via the Selenium API.

Not all page elements that can be interacted with have identifiers available; e.g. map objects that are rendered down into a single map layer. I can locate the object I need by obtaining coordinates via template matching, but I need the ability to place the mouse of the coordinate pair, _without_ clicking in order to display the element's menu.

aaltat:
Please open a new issue,  this is already cluttered with different kinds of feature requests.

PR #344, opened 2014-10-16
Added 2 new keywords 'Mouse Down At' and 'Mouse Up At'

gizm0bill:
Will use it for testing <canvas> interaction. Maybe someone else could use this as well

### DIFF
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -710,4 +710,22 @@ return !element.dispatchEvent(evt);
             raise AssertionError(message)
         self._info("Current page does not contain %s '%s'."
                    % (element_name, locator))
+        
+    def mouse_down_at(self, locator, coordx, coordy):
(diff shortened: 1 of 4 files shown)


emanlove:
Do others have use for such keywords? @ombre42 ? @HelioGuilherme66 ? @aaltat ? others?

These look good to me but I know we have had discussions about keyword overload. My thought is if they (or other proposed keywords) are useful and can't be easily implemented with existing keywords then they should be added.

aaltat:
I currently do not have a need for such keywords, but I can easily think plenty of use for such functionality in multiple different applications. I was just thinking, could the keywords be implemented to existing Mouse Down and Mouse Up keywords? These do basically same thing, only provide coordinate offset.

just-be-dev:
I'm going to close this PR. 

These keywords combine two separate functions: moving the mouse and triggering mouse events. I wouldn't be opposed to adding move mouse keywords (as I suggested in #119), but both moving and acting is more of a user keyword thing.

Issue #341, opened 2014-10-08
Prevent on-failure keyword from running more than once


dpsfrishberg:
I would like to contribute a change to _runonfailure.py and keywordgroup.py to prevent multiple screenshots from being taken on failure. This could happen if one keyword calls another. The way Selenium2Library is currently implemented, this will not happen, but if another library extends Selenium2Library, and it defines a keyword that calls a Selenium2Library keyword, and the inner keyword fails, there will be a duplicate screenshot, because the decorator is applied to both methods. To solve this, I have added a flag that is set when a keyword begins executing.

At any rate, I need to write a test. I am not sure how to test that only one screenshot is taken, without calling the Robot test from a Python test and then checking the directory in which the test was run for multiple screenshot files. I could do this under the unit tests directory, but this is really an acceptance test that runs an inner scenario. What would be the recommended way of doing this? I would be happy to contribute a mechanism for "scenario"-type tests like this.

ombre42:
We love contributions, especially with tests. Avoiding duplicate on failures is good. 
It sounds like your solution is only for when Selenium2Library is extended through inheritance. I'm thinking for extending the library by using a separate class, we could provide a ExtensionKeywordGroup people could inherit from that would provide on failure in the outside library and disable S2L's on failure during the execution of a keyword in the outside library. By default, the same keyword would run whether the failure occurred in the outside library or S2L.
This is just me thinking - I'm not asking you to do that.

Registering a different keyword makes it easy to test how many times the on failure mechanism was triggered. Something like this:

```
On Failure Keyword Only Called Once
    [Setup]    Open Browser    http://www.google.com/    ie
    Set Test Variable    ${ON FAIL COUNT}    ${0}
    Register Keyword To Run On Failure    On Fail
    Run Keyword And Ignore Error    Custom Selenium Keyword    id=missing
    Should Be Equal    ${ON FAIL COUNT}    ${1}    On Failure Keyword called ${ON FAIL COUNT} times.
    [Teardown]    Close Browser

*** Keywords ***
On Fail
    ${count}=    Evaluate    ${ON FAIL COUNT} + 1
    Set Test Variable    ${ON FAIL COUNT}    ${count}
```

dpsfrishberg:
[...]

PR #348, opened 2014-11-07
Run on failure flag

dpsfrishberg:
This is for #341. It solves the problem of calling one keyword (as a class method) from another keyword.


dpsfrishberg:
I'm investigating the failures now. The tests passed the last time I ran them locally; I'll try again.

PR #352, opened 2014-12-05
Update elementfinder.py to allow of use scLocator

IlfirinPL:
Change contains new strategies scLocator and new method _find_by_sc_locator, solution is using build into smart client function isc.AutoTest.getElement

### DIFF
--- a/src/Selenium2Library/locators/elementfinder.py
+++ b/src/Selenium2Library/locators/elementfinder.py
@@ -18,6 +18,7 @@ class ElementFinder(object):
             'jquery': self._find_by_sizzle_selector,
             'sizzle': self._find_by_sizzle_selector,
             'tag': self._find_by_tag_name,
+            'scLocator': self._find_by_sc_locator,
             'default': self._find_by_default
         }
         self._strategies = NormalizedDict(initial=strategies, caseless=True, spaceless=True)
@@ -89,6 +90,10 @@ class ElementFinder(object):
         return self._filter_elements(
             browser.find_elements_by_tag_name(criteria),
             tag, constraints)
+            
+    def _find_by_sc_locator(self, browser, criteria, tag, constraints):
+        js = "return isc.AutoTest.getElement('%s')" % criteria.replace("'", "\\'")
+        return self._filter_elements([browser.execute_script(js)], tag, constraints)
 
     def _find_by_default(self, browser, criteria, tag, constraints):
         if criteria.startswith('//'):


apmanol:
:+1:

src/Selenium2Library/locators/elementfinder.py
```diff
@@ -89,6 +90,10 @@ def _find_by_tag_name(self, browser, criteria, tag, constraints):
         return self._filter_elements(
             browser.find_elements_by_tag_name(criteria),
             tag, constraints)
+            
+    def _find_by_sc_locator(self, browser, criteria, tag, constraints):
+        js = "return isc.AutoTest.getElement('%s')" % criteria.replace("'", "\\'")
```
emanlove, line 95: A couple of questions. One, you use a double backslash "\" when in all your references I see a double forward slash "//". For example, from [here](http://www.smartclient.com/smartgwtee-latest/javadoc/com/smartgwt/client/docs/UsingSelenium.html) I see

```
//ListGrid[ID='filterGrid']
 //DynamicForm[ID="autoTestForm"]/item[name=textField||title=textField||value=test||index=0||Class=TextItem]/element 
```

and from [codecentric](https://blog.codecentric.de/en/2010/12/testing-smartgwt-applications-with-selenium-and-robot-framework/)

```
//DynamicForm[ID="doSearchForm"]/item[name=mainRegisterBox]/element
//Button[ID="searchPerson"]/
//ListGrid[ID="hitList"].0.0
```

Second, you also are replacing within the given locator what seems to be a leading single quote. That makes me assume that you are expecting the user to write a locator like these

```
scLocator='ListGrid[ID='filterGrid']'
scLocator='//ListGrid[ID='filterGrid']'
```

Again I am not familiar with SmartClient and SmartGWT but the general usage for locators within Selenium2Library has been without the quotes. For example

```
Good
----
id=my element id
id=this-element

Not-used
--------
id='quoted-string'
id='no used locator'
```

I know myself I forget this and added quotes ever once in a while. If there is a possibility of having a double space in the locator I can see the need to have the sytax support quoted locators. But again in general that is not how they are used elsewhere.

IlfirinPL:
I am using currently provided solution and its using locators that could be found using selenium ide with extensions for SmartClient
This replacement is used to mask ' if used in locator during execution \' will execution work as  '
 example use is like below is working properly, its not required to have quotes. 
Wait Until Page Contains Element    scLocator=//SectionStack[ID="eventSectionStack"]/section[Class=SectionHeader||index=0||length=1||classIndex=0||classLength=1||roleIndex=0||roleLength=1||title=Event%20Summary||scRole=heading||name=EventSection]/

emanlove:
Ok, I read that wrong. You are un-escaping quotes, \', making them simply quotes, '. Just checking. Do you have any documentation you would like to have added so users will know how to use this locator? And possible any public facing websites that we can run sanity checks on the code base in case any changes occur near this?

IlfirinPL:
About documentation i would like to add new row in strategy table on Selenium2Library.html 
scLocator |
scLocator=//TreeGrid[ID="isc_SideNavTree_0"]/body/row[nodeTitle=Smart%20GWT%20Showcase||0]/col[fieldName=nodeTitle||0] | should match scLocator selector

Here you can find showcase of smartclient
http://www.smartclient.com/smartgwt/showcase/#featured_complete_app

but locators provided by SC are really long and its typical

emanlove:
Ok, I'll add some documentation within a separate pull request.

src/Selenium2Library/locators/elementfinder.py
ombre42, line 95: I think they are just handling single quotation marks embedded in a string using single quotes as the start and termination characters so the JS statement is syntactically OK. jQuery selector is very similar and does the same thing. The double backslash is to escape the backslash which is a special char in Python.
Leaning toothpick syndrome!
Without replace (bad):
`return isc.AutoTest.getElement('//ListGrid[ID='filterGrid']')`
With replace (good):
`return isc.AutoTest.getElement('//ListGrid[ID=\'filterGrid\']')`

PR #358, opened 2015-02-09
method input_password() docstring has been supplemented.

stasiek:
Since it's a known issue ( https://github.com/rtomac/robotframework-selenium2library/issues/202 ) it should at least be mentioned in the docstring.


emanlove:
I agree this should be in the docstring but you should clarify under which scenarios it shows up, that is when using loglevel DEBUG. Otherwise it is confusing to say "it does not log it" but than say "it is in the report file" as those statements seem contradictory.

aaltat:

aaltat:
This is closed because of #984

pekkaklarck:
More precisely commit 550ebaa.

PR #361, opened 2015-02-13
cleanup "Wait Until Element Is Not Visible".

joepurdy:
corrected error messages in new keyword "Wait Until Element Is Not
Visible" to reflect element being visible instead of not visible.

Previously these were left as a copy/paste of the errors from the "Wait Until Element Is Visible" keyword.

### DIFF
--- a/src/Selenium2Library/keywords/_waiting.py
+++ b/src/Selenium2Library/keywords/_waiting.py
@@ -103,9 +103,9 @@ class _WaitingKeywords(KeywordGroup):
             if not visible:
                 return
             elif visible is None:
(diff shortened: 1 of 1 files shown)


emanlove:
Good catch!

ombre42:
@emanlove One of the messages was indeed the opposite of what it should have been.
The message for when visible is None should still be correct because _is_visible only returns None when no elements are found. I think line 106 was OK.

emanlove:
@ombre42 remind me next time to not make commits with a head cold. :)

joepurdy:
@ombre42 and remind me not to tidy up errors early in the morning without coffee. :)
just shot off a new PR without the change to ln 106

Issue #333, opened 2014-09-30
SmartClient and SmartGWT


IlfirinPL:
SmartClient and SmartGWT its not possible to use locator like  'scLocator'
Maybe its possible to port keyword "Add Location Strategy" from SeleniumLibrary or extent this this type of strategy to common available strategy. Other solution to allow import of user-extensions.js/user-extensions-ide.js.

Below some links related to this topic

http://www.smartclient.com/smartgwtee-latest/javadoc/com/smartgwt/client/docs/UsingSelenium.html
https://github.com/MarkusBernhardt/robotframework-selenium2library-java/issues/1
https://blog.codecentric.de/en/2010/12/testing-smartgwt-applications-with-selenium-and-robot-framework/

emanlove:
Looking at the [codecentric article](https://blog.codecentric.de/en/2010/12/testing-smartgwt-applications-with-selenium-and-robot-framework/) it looks as if the locators are simply xpath locators. Is there anything unique with SmartClient and SmartGWT elements? Does one use javascript to find those elements?

ombre42:
I think they just look like XPath. See the JavaScript implementation for [Selenium 1](https://code.google.com/p/smartmobile/source/browse/trunk/mobile/WebContent/tools/selenium/user-extensions.js?r=36).

IlfirinPL:
I have example  like recorded from selenium IDE
type |scLocator=//ListGrid[ID="ChainList"]/filterEditor/editRowForm/item[name=name||title=Chain%20Name||value=test||index=0||Class=TextItem]/element | test

used in robot like
Wait Until Page Contains Element    xpath=//ListGrid[ID="ChainList"]/filterEditor/editRowForm/item[name=name||title=Chain%20Name||value=test||index=0||Class=TextItem]/element
Input Text  xpath=//ListGrid[ID="ChainList"]/filterEditor/editRowForm/item[name=name||title=Chain%20Name||value=test||index=0||Class=TextItem]/element  test

rise following error 

[...]

PR #363, opened 2015-02-18
Angularjs

emanlove:


emanlove:
@zephraph If possible I would really like to get these changes out in 1.7.  The build is failing on the angular tests as they require the angularjs framework. I don't want to add that code to the repository as it would bloat the git object database causing a large wasteful download and it is an external library will change over time.  I do need to figure a way to maybe load it on tests or link external in the test page.

I do consider these changes to be somewhat of a beta and expect changes in the future. But having the changes in the release makes it easier for those who are going to start to need these changes to install them. I also want to get feedback outside of my organization.

just-be-dev:
Do you really want to bake this into the library?

I had this discussion with @IlfirinPL about the smart client integration into S2L as well. Granted, in the end I relented and let him keep it in. 

I don't think we should tie javascript frameworks directly into this library. It would be better, in my opinion, to use custom locators and library extensions to create a new library to be used in conjunction with the S2L to achieve this functionality. That way you could have different versions for, say, angular 1.x and angular 2.x. 

Inside my organization we use ExtJS and there are large divergence between their 3.x versions and the 5.x versions (which are the versions we use the most). 

It would be more beneficial to you in the sense that you can update that library as your organization requires as opposed to having to wait for the long release cycles that we have on S2L.

Do you agree/disagree? I'd like to hear your thoughts.

emanlove:
[sigh] The honest answer is Yes and No. I think before you had the conversation with @IlfirinPL both @ombre42 and I discussed this with him (as you previously referenced #333). We also discussed methods for separating out locator methods. There is definitely a strong argument for moving speciality locators out of the Selenium2Library into an extension library. (Although for what I am seeing with Angular and what I intended to code for location strategy is it not because AngularJS 2 will be much different than AngularJS 1).

There is also reasons for keeping them in. And right now I don't have a lot of great arguments for keeping it in other than it is easier that way. Yes poor argument but it is late. With the angular one issue is needing to be tightly bound.  For example the model for my angular locator is Protractor which is the testing framework for AngularJS and is built on top of WebDriverJS.  One thing Protractor does, which I have not yet implemented in the angular locator code, is the it overrides all the webdriver calls to add a wait for angular promise. Could this all be implement in a separate library, probably Yes. But again right now I don't want to work through separating the library but overriding all the calls.

I'm not sure how this is much different than the Add Locator Strategy code we've added in. This question of locators should be discussed as it is core to everything the library does.

just-be-dev:
You should take a look at my [ExtJS library](https://github.com/Selenium2Library/robotframework-extjs). I just started it today so it's in the very, very early stages, but maybe it will give you an idea of what I intended with the custom locators. 

If I took some time in the next few days to port what you've got to an external library, would you take some time to see if that solution is viable for you? I don't mean extending S2L, but rather making a helper library that works along side it. I really think the modular add-on approach is the better way to go here.

src/Selenium2Library/locators/elementfinder.py
```diff
@@ -121,9 +124,108 @@ def _find_by_sc_locator(self, browser, criteria, tag, constraints):
...
+            return matches
+        
+        matches = rows
+        matches.extend(multiRows)
+        return matches
+
     def _find_by_default(self, browser, criteria, tag, constraints):
         if criteria.startswith('//'):
             return self._find_by_xpath(browser, criteria, tag, constraints)
+        elif criteria.startswith('{{'):
```
just-be-dev, line 227: Ah, now this is pretty neat. I should've definitely added this as a part of the custom locator thing.... I'll go make an issue for that.

zlorb:
What is the plan for completing this pull request?

divfor:
Looks like this is still a requirement for locator wrapper. can you please review my PR #543 to see if it can separate your desired support from S2L?

emanlove:
@zephraph and I separated out support for AngularJS locators into a separate library, [AngularJS Library](https://github.com/Selenium2Library/robotframework-angularjs). The reason we did this is so we can update or support the various Angular versions simultaneously; something we couldn't really do if it was part of the core Selenium2Library. Anyone interested in testing AngularJS applications should also checkout Richard Huang's complementary library, [the Extended Selenium2Library](https://github.com/rickypc/robotframework-extendedselenium2library). The Extended Selenium2Library adds support for waiting for angular and other useful functionality.

For more information see  [AngularJS Library](https://github.com/Selenium2Library/robotframework-angularjs).

Issue #364, opened 2015-02-24
Add Location Strategy Keyword


just-be-dev:
The first selenium library had the `Add Location Strategy` keyword. It essentially allowed for instances like [this](https://github.com/MarkusBernhardt/robotframework-selenium2library-java/blob/master/src/test/robotframework/testsuites/UnitTests/ExtJS.txt) which is exactly what I happen to need at the moment. 

Is there a particular reason why it isn't present? I found the [java implementation](https://github.com/MarkusBernhardt/robotframework-selenium2library-java/blob/master/src/main/java/com/github/markusbernhardt/selenium2library/locators/ElementFinder.java#L237) of this keyword.

emanlove:
We were actually having a discussion within #333 about allowing user specified locator strategies. Just curious what type of element are you trying to locate that is already satisfied by the current strategies?

just-be-dev:
I want to locate using ExtJs' component query (see the _this_ link above. What I have in mind is basically a javascript locator. Already working on a PR. 

It's pretty much exactly what's implemented in the Java version I linked to.

emanlove:
Looking at that [example](https://github.com/rtomac/robotframework-selenium2library/issues/333) what I see is this is a javascript locator. Just wandering if there is other locator type that we can implement with a very similar keyword/argument syntax that is not javascript based. The reason I ask is to see if we need to add an argument that states which method the locator code implements. Something like

```
Open Page
  # 'Add Location Strategy'  | strategy name | strategy method |  stategy implementation "code" 
  Add Location Strategy extJs  javascript return window.document.getElementById((Ext.ComponentQuery.query(arguments[0])[0]).getId());
```

Better to add it now then trying to patch in later.

just-be-dev:
Well, I suppose that's entirely possible. You could just provide it with a python function to run. It wouldn't be all doable from within RF, but I feel like this is somewhat of a special case anyway so I think that'd be acceptable. 

Edit: After thinking about it a bit I've got a pretty good idea of how we can implement general, reusable solution. I'll get back to you when I have a basic implementation done.

just-be-dev:
See #365 for progress on this issue.

PR #365, opened 2015-02-25
Add Location Strategy Keyword

just-be-dev:
This was discussed both in #333 and #364 and here's a preliminary thought that I had. 

Instead of tying the `Add Location Strategy` to javascript as I originally intended I've designed it so that a location strategy is actually implemented as a keyword. This gives a lot of flexibility in that location strategies can (if desired) be written directly in RF as a user keyword.

Keep in mind that this is just a rough draft to get started with. 

Idealistically I'd have this

``` robotframework
*** Settings ***
Library    Selenium2Library
Resource    Locators
Resource    Web Navigation Stuff

*** Keyword ***
Setup Locators
    Add Location Strategy    custom    My Custom Locator

*** Test Cases ***
My Test
    [Setup]    Setup Locators
    Open Browser To My Site
    Click Element    custom=mySelector[example=true]
```

### DIFF
--- a/src/Selenium2Library/locators/elementfinder.py
+++ b/src/Selenium2Library/locators/elementfinder.py
@@ -1,6 +1,7 @@
 from Selenium2Library import utils
 from robot.api import logger
 from robot.utils import NormalizedDict
+from robot.libraries.BuiltIn import BuiltIn
 
 
 class ElementFinder(object):
@@ -22,6 +23,7 @@ class ElementFinder(object):
             'default': self._find_by_default
         }
         self._strategies = NormalizedDict(initial=strategies, caseless=True, spaceless=True)
+        self._default_strategies = strategies.keys()
 
     def find(self, browser, locator, tag=None):
         assert browser is not None
@@ -34,6 +36,30 @@ class ElementFinder(object):
             raise ValueError("Element locator with prefix '" + prefix + "' is not supported")
         (tag, constraints) = self._get_tag_and_constraints(tag)
         return strategy(browser, criteria, tag, constraints)
+
+    def register(self, strategy, persist):
+        if strategy.name in self._strategies:
+            raise AttributeError("The custom locator '" + strategy.name +
+            "' cannot be registered. A locator of that name already exists.")
+        self._strategies[strategy.name] = strategy.find
+
+        if not persist:
+            # Unregister after current scope ends
+            suite = BuiltIn().get_variable_value('${SUITE NAME}')
+            test  = BuiltIn().get_variable_value('${TEST NAME}', '')
+            scope =  suite + '.' + test if test != '' else suite
+            utils.events.on('scope_end', scope, self.unregister, strategy.name)
+
+    def unregister(self, strategy_name):
+        if strategy_name in self._default_strategies:
+            raise AttributeError("Cannot unregister the default strategy '" + strategy_name + "'")
+        elif strategy_name not in self._strategies:
+            logger.info("Cannot unregister the non-registered strategy '" + strategy_name + "'")
+        else:
+            del self._strategies[strategy_name]
+
+    def has_strategy(self, strategy_name):
+        return strategy_name in self.strategies
 
     # Strategy routines, private
 
@@ -90,7 +116,7 @@ class ElementFinder(object):
         return self._filter_elements(
             browser.find_elements_by_tag_name(criteria),
             tag, constraints)
-            
+
     def _find_by_sc_locator(self, browser, criteria, tag, constraints):
         js = "return isc.AutoTest.getElement('%s')" % criteria.replace("'", "\\'")
         return self._filter_elements([browser.execute_script(js)], tag, constraints)
--- a/src/Selenium2Library/utils/events/__init__.py
+++ b/src/Selenium2Library/utils/events/__init__.py
@@ -1,0 +1,27 @@
+from scope_event import ScopeStart, ScopeEnd
+
+_registered_events = [ ScopeStart, ScopeEnd ]
+_events = []
+
+__all__ = [
+    "on",
+    "dispatch",
+    "register_event"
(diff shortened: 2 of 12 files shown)


just-be-dev:
@emanlove, could you take a look at this and give me your thoughts?

Thanks.

just-be-dev:
This is now feature complete and ready for review. I've added tests and documentation. 

It includes two new keywords:
- `Add Locator Strategy`
- `Remove Locator Strategy`

The locator strategy itself is a user keyword as shown in the documentation. This allows easy integration with JavaScript (which was my original goal), but also covers any other applicable situation. 

An added locator strategy will stay present throughout the duration of testing unless removed by `Remove Locator Strategy`.

just-be-dev:
`Add Locator Strategy` is a global action. When this keyword is called, the locator strategy that is specified doesn't go away unless Remove Locator Strategy is subsequently called. 

Global alterations of a testing environment by a test are generally unwise. @pekkaklarck, @emanlove what do you think of this behavior? I suppose it would be possible through library listener methods to automatically unregister the locator strategy after the scope it was registered in has subsided. 

Which would be better? The current behavior (after it's set anywhere it stays until it is removed) or the scoped behavior (where if it's set in a test it's removed after that test is complete).

pekkaklarck:
@zephraph, automatically removing locator strategies sounds like a good idea. Library listeners ought to make implementing it pretty simple. Should probably be configurable with an argument like `auto_remove=False`, though.

just-be-dev:
I've added a specific test because I was in talks with @IlfirinPL on #376. This test case will likely be removed before merge along with the changes mentioned by @pekkaklarck above. 

Don't merge this as is.

just-be-dev:
@pekkaklarck, when you get a chance could you review this?

Adding the auto unregister made it a bit more complicated.

just-be-dev:
@emanlove, could you please review this? Actually, look over everything in the 1.7 Milestone. I really want to go ahead and push a new release this weekend.

just-be-dev:
A large part of the bulk of this is the scoping as suggested by @pekkaklarck. The original implementation made it persistent by default. This implementation makes it scoped by default (which required hooking into listeners). I try to make a little event system so if we needed to do stuff like this in the future we could just add new events.

pekkaklarck:
Sorry, I was on a business trip with limited time for other work. What commits/code should be reviewed?

just-be-dev:
Just look [here](https://github.com/rtomac/robotframework-selenium2library/pull/365/files) and see if anything jumps out at you. I really would like you to look over the event system I wrote for scoping the  locator registration and tell me if it's too much.

src/Selenium2Library/locators/elementfinder.py
```diff
@@ -35,6 +37,30 @@ def find(self, browser, locator, tag=None):
         (tag, constraints) = self._get_tag_and_constraints(tag)
         return strategy(browser, criteria, tag, constraints)
 
+    def register(self, strategy, persist):
+        if(strategy.name in self._strategies):
```
pekkaklarck: Parens? Also in `unregister` below. #pep8

src/Selenium2Library/utils/events/__init__.py
```diff
@@ -0,0 +1,28 @@
...
+_events = []
+
+__all__ = [
+	"on",
+	"dispatch",
+	"register_event"
+]
+
+def on(event_name, *args, **kwargs):
+	for event in _registered_events:
```
pekkaklarck: Are tabs used here or is indentation this long?

src/Selenium2Library/utils/librarylistener.py
```diff
@@ -0,0 +1,18 @@
+import events as event
+from robot.api import logger
+
+class LibraryListener(object):
+
+	ROBOT_LISTENER_API_VERSION = 2
+
+	def _start_suite(self, name, attrs):
```
pekkaklarck: It would be OK not to have the leading underscore because this class is not used as a library.

pekkaklarck:
The event system looks pretty neat but I'm not entirely sure is it necessary for this usage. Are you planning to use it for something else too? If not, then there's a risk that YAGNI.

I also added some nitpicking comments about style. I'm not sure is there any agreed styles for this project so feel free to ignore them.

src/Selenium2Library/locators/elementfinder.py
just-be-dev: Ah, good catch. That's my Javascript habits kicking in.

src/Selenium2Library/utils/events/__init__.py
just-be-dev: Ah, crap. I had didn't have my editor set on soft-tabs. Good catch.

just-be-dev:
Well, it's quite possible I could use it again in the future, but I can't guarantee that. That's just how I thought to implement it. Also, I trimmed the scope events down a bit. It's more DRY now.

pekkaklarck:
Ok, looks good.

just-be-dev:
So, are we good to merge?

divfor:
I have  an idea on similar enhancement:
https://github.com/divfor/robotframework-selenium2library/pull/6
This enhancement is to simplify locator usage. Generally it is difficult for junior engineers and customer to understand the locators written in user keywords or test steps. It would be better if they can use Web UI texts/words prompting for target element instead of locators. The real locators could be mapped by prompting UI text in variable files created and maintained by senior engineers. If so, the test cases will be more easier to read.

just-be-dev:
I'm not really following you. The custom locators written here should be specified in a resource file or library, not included in a test. I wrote this feature with javascript frameworks like angular and ExtJs in mind, though it's not limited to those sorts of things. I would say it's more up to the locator developer to document their locator usage. By and large this should be a rarely used feature though.

Issue #367, opened 2015-02-27
Add get window position keyword


ktarasz:
There is a Selenium WebDriver command for get the window position.

Issue #366, opened 2015-02-27
Add set and get window position keyword


ktarasz:
There is a Selenium WebDriver command for setting the window position.

ktarasz:
Test for 'Get and Set Window Position' passed... (https://github.com/rtomac/robotframework-selenium2library/pull/368)  It's looks strange...

just-be-dev:
This is now merged.

PR #368, opened 2015-02-27
Added of 'Get Window Position' and 'Set Window Position' keywords

ktarasz:
Added 'Get Window Position' and 'Set Window Position' keywords matching the Selenium functionality.
 #366 
 #367

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -16,6 +16,10 @@ Release Notes
 - Added new locator strategy, scLocator, for finding SmartClient and SmartGWT elements.
   [IlfirinPL]
 
+- Added 'Get Window Position' and 'Set Window Position' keywords matching the
+  Selenium functionality.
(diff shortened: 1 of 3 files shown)


ktarasz:
The 'Double Click Element' test case failed, but 'Get and Set Window Position' passed. Can someone help me with the failed test?

just-be-dev:
I don't think this is actually a failure with your PR. The same thing happened with mine (#365).

just-be-dev:
@ktarasz I somewhat traced it down. It's a known failure though it doesn't actually seem to be a problem with this library. It seems like it's an actual bug in selenium. Regardless, I've added a PR to make it a non-critical failure so you test will pass after that gets merged. (See #371)

just-be-dev:
@ktarasz, if you'll update this I'll merge it.

PR #370, opened 2015-04-03
Replaced execfile with import

just-be-dev:
I'm trying to use this library with cx_Freeze, but the `execfile` was causing issues because all the dependencies are in a zipped directory. `execfile` doesn't work well with zipped files it seems. 

Regardless, this was somewhat unnecessary anyways.


just-be-dev:
This is a small change, so I'm going to go ahead and merge it in.

PR #371, opened 2015-04-03
Made known issues non-critical

just-be-dev:
The double click testcase is failing for Firefox but tagged as a known issue. I've added a non-critical flag to the travis script so that any test labeled as `Known Issue - Browser` will be treated as non-critical and not cause PRs and other tests to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ before_script:
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
 script:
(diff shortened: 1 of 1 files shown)


just-be-dev:
Okay, so I'm not sure exactly what's happening here. For whatever reason on python 2.6 the `Get Window Titles` test case is failing. It's strange because I didn't see this issue on [my build](https://travis-ci.org/zephraph/robotframework-selenium2library/builds/57097194). Could this be accounted by different versions of either firefox or selenium?

PR #375, opened 2015-04-13
Add 'Wait Until Page Not Contains Element' -keyword

molsky:
I hope that this is implemented in proper way.


molsky:
Well of course I managed to typo my commit message so now it is misleading `Wait Until Page Not Contains Element` (that was actually implemented in this commit) vs `Wait Until Page Not Contains`.

Edit:
Amended commit and then force pushed it again.

just-be-dev:
This should probably be `Wait Until Page Does Not Contain Element`. It's more verbose, but more proper.

src/Selenium2Library/keywords/_waiting.py
```diff
@@ -62,6 +62,27 @@ def wait_until_page_contains_element(self, locator, timeout=None, error=None):
...
+    def wait_until_page_not_contains_element(self, locator, timeout=None, error=None):
+        """Waits until element specified with `locator` disappears from current page.
+
+        Fails if `timeout` expires before the element disappears. See
+        `introduction` for more information about `timeout` and its
+        default value.
+
+        `error` can be used to override the default error message.
+
+        See also `Wait Until Page Contains`, `Wait For Condition`,
```
just-be-dev, line 74: Should this be ``Wait Until Page Contains Element``?

molsky:
I used the following ticket as a reference for missing keywords:
https://github.com/rtomac/robotframework-selenium2library/issues/246
The keyword name `Wait Until Page Not Contains Element` was taken from that list. 
I think this needs others opinion should keyword's name be changed from `Not Contains` to `Does Not Contain` format.

just-be-dev:
@emanlove, what do you think on the wording?

emanlove:
@zephraph We had a discussion about wording a while back. It centred around the idea that the [java port of the Selenium2Library](https://github.com/MarkusBernhardt/robotframework-selenium2library-java) had some keywords that weren't yet within this library. As some got added we tended toward using the same keyword name. But for some of them they we too grammatically incorrect.

Personally I think it should be ...Does Not Contain... Looking to see if I can find that discussion...

molsky:
Trying to do the changes to the keyword name within a few days. I agree with you that `Does Not Contain` makes more sense than `Not Contains`.

just-be-dev:
@molsky, Once you rebase and make those changes I'll merge this for you. If you need help rebasing let me know. Thanks for contributing!

molsky:
Changes done and I hope that I did rebasing and pushing changes in correct way.

just-be-dev:
Awesome job! Thanks for contributing. =]

Issue #372, opened 2015-04-07
Choose File does not fail on invalid file path


stevejefferiesIDBS:
Currently "Choose File" checks for the files existence (os.path.isfile) and will log an info statement if the file does not exist, but does not fail the keyword. This causes issues when the browser is interacting with a native select file dialog, such as the IEDriver does. In the case of IE the test then hangs since the native dialog cannot be closed.

just-be-dev:
@stevejefferiesIDBS I'll take a look at this for you.

just-be-dev:
@stevejefferiesIDBS Assuming there are no objections you can expect to see this in the coming 1.7.0 release.

PR #380, opened 2015-04-28
Fix for #372

just-be-dev:
The change makes the **Choose File** keyword fail if the chosen file is not found. 

It fixes an issue in some browsers where the webdriver will lock up if a non-existent file is entered.


just-be-dev:
Consequently I found an issue with the **Choose File** test case that's now fixed.

PR #391, opened 2015-05-11
Add 'Wait Until Element Text Is' -keyword

molsky:
I have found some use for this kind of keyword when verifying text content in error message elements on forms (Text is too short -> Text is too long -> Text contains characters that are now allowed etc. when doing input field checks with Javascript).

### DIFF
--- a/src/Selenium2Library/keywords/_waiting.py
+++ b/src/Selenium2Library/keywords/_waiting.py
@@ -155,6 +155,29 @@ class _WaitingKeywords(KeywordGroup):
 
         self._wait_until_no_error(timeout, check_enabled)
 
+    def wait_until_element_text_is(self, locator, text, timeout=None, error=None):
+        """Waits until `text` appears on given element.
+
+        Fails if `timeout` expires before the text appears on given element. See
+        `introduction` for more information about `timeout` and its
+        default value.
+
+        `error` can be used to override the default error message.
+
+        See also `Wait Until Page Contains`, `Wait Until Page Contains Element`, `Wait For Condition`,
+        `Wait Until Element Is Visible` and BuiltIn keyword `Wait Until
+        Keyword Succeeds`.
+        """
+        element = self._element_find(locator, True, True)
+        def check_text():
+            actual = element.text
+            if text == actual:
+                return
+            else:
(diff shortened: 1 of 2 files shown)


just-be-dev:
Isn't this what the `Wait Until Page Contains` Keyword is for?

just-be-dev:
If you wanted an element version of this keyword I could see maybe `Wait Until Element Contains` or something. That's pretty much how we do text matching though.

molsky:
`Wait Until Page Contains` is problematic in certain situations.
1. There is Javascript code directly in HTML file's source code. In the beginning of that file function X updates element's text and that contains the text that you are looking for so that check will pass always. 
2. Page have multiple elements that will have same text in certain situations and you are looking that all of them have specific text or only just one specific element will have it.

Now I agree that `Wait Until Element Contains` would be much better option than this implementation where text must be exactly same than given in argument.

just-be-dev:
I'd be okay for `Wait Until Element Contains`. You can change this PR to match that or open a new one. I'll list it on #246.

ombre42:
I often do something like:
`Wait Until Page Contains Element    id('some_id')/div[.='This exact text']`
But some people have a strong preference for CSS over XPath. This also fails the same way if the element is present, but with different text than if the element is not there at all.
Wait For Condition could also be an alternative here, but I have not used it to match text.
I think the need is there for a `Wait Until Element Contains`, but I would be hesitant to also have `Wait Until Element Does Not Contain`, `Wait Until Element Text Is`, `Wait Until Element Text Is Not`, `Wait Until Element Text Matches Regexp`, `Wait Until Element Text Does Not Match Regexp`, or `Wait Until Element Has Text And Is Also Purple` :)

just-be-dev:
Yeah, I'm inclined to agree.

molsky:
I will open new PR for this that also contains implementation for `Wait Until Element Does Not Contain`.

And yes I agree with you. It is true that if all possible new keyword request are going to be added to this library it will soon contain massive amount of different keywords and most of them will be useful only in very rare situations. And also it is difficult to handle and maintain that massive library.

just-be-dev:
All right, sounds good. I'll just go ahead and close this PR then.

PR #394, opened 2015-05-12
add fix for 'No Such Window' error

divfor:
reason:
from 6.4 secotion in http://www.w3.org/TR/webdriver/, it says:
Once the window has closed, future commands must return a status code of
no such window until a new window is selected for receiving commands.
how to fix:
1. let close_window() automaticly switches to prevoius window;
2. capture NoSuchWindowException when select_window() /
_select_matching()/get_current_window_handle()

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,11 @@ Release Notes
 
 1.7 (unreleased)
 ----------------
+- Fixed ‘NoSuchWindowException' issue. Running keyword 'Select Window' after 'Close Window'
+  will trigger this issue if locator has prefix 'name=','title=' or 'url='. Also fixed same
+  issue for keywords 'Get Window Ids', 'Get Window Titles' and 'Get Window Names'.
+  [divfor]
+  
 - Corrected error message in new keyword 'Wait Until Element Is Not
   Visible' to reflect element being visible instead of not visible.
   [joepurdy]
--- a/src/Selenium2Library/keywords/_browsermanagement.py
+++ b/src/Selenium2Library/keywords/_browsermanagement.py
@@ -212,8 +212,15 @@ class _BrowserManagementKeywords(KeywordGroup):
     # Public, window management
 
     def close_window(self):
-        """Closes currently opened pop-up window."""
+        """Closes currently opened pop-up window and switch back to previous window.
+        It assumes the last opened window is in the last index of window list.
+        However, as opened windows are list-returned from tree-order in remote end,
+        please use 'Select Window' if automatic switching does not work as expected.
+        """
         self._current_browser().close()
+        handles = self._current_browser().get_window_handles()
+        if len(handles) >= 1:
+            self._current_browser().switch_to_window(handles[-1])
 
     def get_window_identifiers(self):
         """Returns and logs id attributes of all windows known to the browser."""
--- a/src/Selenium2Library/keywords/_browsermanagement.py
+++ b/src/Selenium2Library/keywords/_browsermanagement.py
@@ -212,15 +212,8 @@ class _BrowserManagementKeywords(KeywordGroup):
     # Public, window management
 
     def close_window(self):
-        """Closes currently opened pop-up window and switch back to previous window.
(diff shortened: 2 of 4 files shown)


just-be-dev:
I also need you to write some tests and update the changelog.

divfor:
should I put the test case in /test/acceptance/open_and_close.txt?

just-be-dev:
Yeah, that sounds good to me.

divfor:
It is very useful to surface window handle to users that want to have the 'correct way' - comparing window handle sets, to switch window. If we do this way, the fix here for 'Close Window' is not required. instead, my new proposal is to update 'Select Window' and add 2 keywords:

``` robotframework
Close Window And Switch To | locator=None
```

This will close current window and then switch to window if locator is given. locator can be name, title, url, handle, ..., the same as 'Select Window' requires.

``` robotframework
${previous_window}= | Select Window | locator=None
```

This will switch to window matching locator and return previous window's handle. locator can be name, title, url, handle of target window, or a handle list to exclude in order to find new window. If locator is 'self', only return current window handle. 

``` robotframework
${window_list}= | List Windows
```

This will return a list of all opened window handles in same session not assuming its order. 

Above changes should not impact existing usage of 'Close Window' and 'Select Window', and can offer the strict 'correct way' for using handle to switch window.

example 1:
A user should save main window handle in the beginning if required:

``` robotframework
Open Browser 
${main_window}= | Select Window | self
```

A user can switch to correct window like below:

``` robotframework
${all_old_windows}= | List Windows
Click Link | xxx | # or any other action that generates a new window
${parent_window}= | Select Window | ${all_old_windows}
Close Window And Switch To | ${parent_window}
```

Additional proposal:
it is safe to think the first & top level browsing context, the root of tree-order of browsing contexts, will be at index[0] in list-order returned window handles. If a WD session has multiple top level browsing contexts, there will be multiple short lists joined together to form a single big list to return WD's local end, in this case the first top level browsing context will still be at index[0]. So, I think get_window_handles()[0] is safe to switch to 'main' window in current ff/gc/ie browsers. of course, it is not official defined. however, there are still many protocol gaps that are not defined:)

So an additional question is,  should we also offer a express way to select the last-position indexed window as new window without the need to comparing with old window set? I guess this will be correct in 99% cases.

``` robotframework
${parent_window}= | Select Window | new | #use window list instead 'new' to exclude them if disordered
```

divfor:
http://www.w3.org/TR/html5/browsers.html#top-level-browsing-context
5.1.1 Nested browsing contexts
The list of the descendant browsing contexts of a Document d is the (ordered) list returned by the following algorithm:

1. Let list be an empty list;
2. For each child browsing context of d that is nested through an element that is in the Document d, in the tree order of the elements nesting those browsing contexts, run these substeps:
  a. Append that child browsing context to the list;
  b. Append the list of the descendant browsing contexts of the active document of that child browsing context to the list;
3. Return the constructed list.

divfor:
Anyway this PR should focus on NoSuchWindowException fixing. I raised another PR #395 for improve selecting new window with handles.

I reversed the change for 'Close Window' and only deal with this exception in  _get_window_infos() and _select_matching(). 

I test the PR against my real web server and it works well. however I did not add test code to acceptance test as I can not get its test server started smoothly to pass testing -- tried activepython2.7.8 with chrome/fireforx on win8.1 and winxp...

just-be-dev:
Run `python run_tests.py python firefox` from the tests directory to run the acceptance tests. I'll need you to add some tests and update the changelog before I can merge.

divfor:
Added two test cases and tested that they reproduce the issues and verify the fix.
Added changelog.

just-be-dev:
Okay, I'm back from vacation so I have a bit of time to review this. I might take a little time though, because I want to test this out myself before merging.

just-be-dev:
Hey @divfor, this looks good to me. I've tested it, verified the failures on master, and verified the working solution. I'm going to go ahead and merge this. 

Thanks for your contribution!

PR #395, opened 2015-05-14
Enhancement for selecting new or any opened window

divfor:
Code changes:
added 'Close Window And Select' to simplify two steps of 'Close Window' + 'Select Window'
added 'List Windows' to return all current handles;
enable 'Select Window' to always return the handle of window it is selecting from, and only return handle if 'self' as locator;
enriched 'Select Window' to select any opened window by a window handle;
enriched 'Select Window' to select new window by excluding a list of earlier window handles;

Background:
It is very useful to surface window handle to users that want to have the 'correct way' - comparing window handle sets, to switch window. If we do this way, the fix here for 'Close Window' is not required. instead, my new proposal is to update 'Select Window' and add 2 keywords:

``` robotframework
Close Window And Select | locator=None
```

This will close current window and then switch to window if locator is given. locator can be name, title, url, handle, ..., the same as 'Select Window' requires.

``` robotframework
${previous_window}= | Select Window | locator=None
```

This will switch to window matching locator and return previous window's handle. locator can be name, title, url, handle of target window, or a handle list to exclude in order to find new window. If locator is 'self', only return current window handle. 

``` robotframework
${window_list}= | List Windows
```

This will return a list of all opened window handles in same session not assuming its order. 

Above changes should not impact existing usage of 'Close Window' and 'Select Window', and can offer the strict 'correct way' for using handle to switch window.

example 1:
A user should save main window handle in the beginning if required:

``` robotframework
Open Browser 
${main_window}= | Select Window | self
```

A user can switch to correct window like below:

``` robotframework
${all_old_windows}= | List Windows
Click Link | xxx | # or any other action that generates a child window
${parent_window}= | Select Window | ${all_old_windows}
Other Actions Against The Child Window
Close Window And Select | ${parent_window}
```

One more consideration:
should we also offer the additional simplified way to select the last-position indexed window as new window without the need to comparing with old window handles? The doc can prompt user to use strict way to select new window if the simplified way does not work for him/her. From my experiences the simplified way will be correct in 99% cases. 

``` robotframework
${parent_window}= | Select Window | new
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,6 +3,14 @@ Release Notes
 
 1.7 (unreleased)
 ----------------
+- Added 'Close Window And Select' to simplify 'Close Window' + 'Select Window'.
+  Added 'List Windows' to return a list of all current handles.
+  Enhanced 'Select Window' to select new window by excluding a list of window handles
+  (the strict way), or by special locator 'new' (the simplified but less strict way)
+  Enhanced 'Select Window' to accept window handle as locator, and special locator
+  'self' to return current window handle.
+  [divfor]
+
 - Corrected error message in new keyword 'Wait Until Element Is Not
   Visible' to reflect element being visible instead of not visible.
   [joepurdy]
--- a/src/Selenium2Library/keywords/_browsermanagement.py
+++ b/src/Selenium2Library/keywords/_browsermanagement.py
@@ -290,6 +290,7 @@ class _BrowserManagementKeywords(KeywordGroup):
 
     def select_window(self, locator=None):
         """Selects the window found with `locator` as the context of actions.
+        Return value is ether current window handle before select action or None.
 
         If the window is found, all subsequent commands use that window, until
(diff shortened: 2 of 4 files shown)


just-be-dev:
So I like most of this. I'm not so sure about the close and select keyword though. It's nice, but it seems redundant and we already have so many keywords. 

@emanlove would you take a look at this?

just-be-dev:
Hey @divfor, do you think the `Close and Select` keyword is really necessary? I know it shorter, but I'm worried about keyword bloat.

divfor:
it is not  a must and can be written as a user keyword. I will remove it.

just-be-dev:
Closing this in favor of #407.

divfor:
I create a new PR #407 to resolve conflicts - as I tried to rebase my master but failed, so I re-clone rtomac/master

pekkaklarck:
I agree separate `Close and Select` keyword isn't really needed. Adding an option for `Close` to do that (e.g. `select=None`) might be nice, though.

PR #396, opened 2015-05-14
Add qitao's table keywords

divfor:
Qitao:
Get Table Rows
Get Table Cols At Row
Adam:
Get Index In Table Column
Get Index In Table Row


divfor:
sorry. wrong submit. please reject it.

PR #397, opened 2015-05-17
New keyword Press Keys

HelioGuilherme66:
New keyword Press Keys (keeping Press Key). This keyword allows to send a sequence of keys with simultaneously pressing modifiers keys, like Alt, Control, Shift. The acceptance test file was renamed to zkeyboard.robot to be the last being run, because it may cause side effects in Firefox.
Tried to squash commits but did not get a clean PR.


HelioGuilherme66:
Please consider only commit https://github.com/HelioGuilherme66/robotframework-selenium2library/commit/0bbddd59f9745eeadac0e9ab1a82fac3140d2366. Thanks.

just-be-dev:
I like the basic idea of the overall change, but ultimately I can't merge a single commit as a PR. If you can edit this PR to be based off of master with only this commit on top then I can review it further. 

I'm going to close the PR until that time, but if you can get this commit on top of master we can reopen it and work from there. 

Thanks for your submission! If you have any questions or need any assistance let me know.

Issue #398, opened 2015-05-18
Set Window Position test failing on Ubuntu 15.04 / Latest FF 40.0a2


just-be-dev:
When attempting to set the Y position on a FF window that opens "snapped" to the top menu bar and full length (not full screen) the window will not be moved.

just-be-dev:
Upon further investigation I found that both `Set Window Size` and `Set Window Position` aren't given enough time to properly execute. When adding a delay, `Set Window Size` fails on  my Ubuntu setup because the window can't be sized above the maximum screen height if the window is snapped. I fixed this by simply reducing the overall resize height. 

The issue I reported above is fixed by resizing the window before trying to set the window position. 

@emanlove, @pekkaklarck, any ideas on addressing these timing issues? Should I just document that the action may not occur until some point after the keyword executes? I would add a timeout and just watch for the correct changes but `Set Window Size` lies and says it's the right size until the actual update occurs in which case it gives the correct values.

pekkaklarck:
Is it possible to poll and wait until windows is resized/moved? If not, then documenting this behavior ought to be fine.

just-be-dev:
It does seem possible for the move, but the resize immediately reports the correct size but gets updated again after the resize occurs. I don't know what's up with the inconsistent behavior, it's just something on the webdiver's side.

HelioGuilherme66:
I also got tests failing with Firefox (34.5) and Chrome (44.0.2403.155)  in Fedora 19 (but not in Windows 7).
Firefox:
`Get and Set Window Size :: Get and Set Window Size                    | FAIL |
846 != 800`
`Get and Set Window Position :: Get and Set Window Position            | FAIL |
100 != 0`

Chrome:
`Get and Set Window Position :: Get and Set Window Position            | FAIL |
100 != 102`

manycoding:
Here is another example of snapping in Fedora 22.
I noticed no matter what page height do you choose, after setting the position it can't be more than screen height - bar size - y position.
For example, if I resize with 860 and any height more than 1006, and set position to 1060 20 - the height after position is changed will be 1006 (=1080-54-20).

```
KEYWORD Selenium2Library . Set Window Size @{WINDOW_SIZE}
Documentation: Sets the `width` and `height` of the current window to the specified values.
Start / End / Elapsed:  20151118 13:59:38.947 / 20151118 13:59:38.977 / 00:00:00.030
[...]

PR #399, opened 2015-05-18
Fix window tests

just-be-dev:
Temporary workaround for #398. 

Waiting for more direct input on how to address this.


just-be-dev:
This doesn't need to be merged. I need to actually fix these keywords, not out a bandaid on them.

PR #401, opened 2015-05-19
Modified .gitignore to ignore virtualenv dir

just-be-dev:
I use 'env' as my virtualenv directory so I added it to the .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ chromedriver.log
 
 dist
 build
+env


emanlove:
I put my virtualenv one directory up so it is not within git source directories. For example

``` bash
 ~ $ mkdir s2l-new-feature
 ~ $ cd s2l-new-feature/
 ~/s2l-new-feature $ virtualenv -p /usr/bin/python2.7 --no-site-packages clean-python27-env
 ~/s2l-new-feature $ source clean-python27-env/bin/activate
 (clean-python27-env) ~/s2l-new-feature $ git clone git@github.com:rtomac/robotframework-selenium2library.git rf-s2l
```

just-be-dev:
Ah, that's a good point. 

Regardless I just found this [link](https://help.github.com/articles/ignoring-files/#create-a-global-gitignore) about having a global git ignore file. I'll use that instead.

PR #403, opened 2015-05-19
New keyword Press Keys, which allows pressing modifiers keys like CTR…

HelioGuilherme66:
…L, ALT, SHIFT.

Also corrects documentation and code in Press Key (needs \\KEYCODE two '\'), Documentation needs escaped '\' to show properly.


src/Selenium2Library/keywords/_element.py
```diff
@@ -480,13 +480,16 @@ def press_key(self, locator, key):
...
         `key` is either a single character, or a numerical ASCII code of the key
-        lead by '\\'. In test data, '\\' must be escaped, so use '\\\\'.
+        lead by '\\\\'. In test data, '\\' must be escaped, so use '\\\\'.
 
         Examples:
-        | Press Key | text_field   | q |
-        | Press Key | login_button | \\\\13 | # ASCII code for enter key |
+        | Press Key | text_field   | q                |                                               |
+        | Press Key | login_button | \\\\13           | # ASCII code for enter key                    |
+        | Press Key | nav_console  | \\\\\\\\ARROW_UP | # selenium.webdriver.common.keys ARROW_UP KEY |
```
just-be-dev, line 487: Why are there so many escapes here? Is it necessary?

HelioGuilherme66, line 487: The escapes were added to be correctly seen in HTML pages and RIDE context help.
They are double backslash for ASCII codes, and quadruple for KEY NAME, and then escaped for being shown.

just-be-dev, line 487: Ohhh, well that makes sense then. Sorry about that, ha.

src/Selenium2Library/keywords/_element.py
```diff
@@ -734,6 +791,79 @@ def _map_ascii_key_code_to_key(self, key_code):
...
+            'RIGHT': Keys.RIGHT,
+            'SEMICOLON': Keys.SEMICOLON,
+            'SEPARATOR': Keys.SEPARATOR,
+            'SHIFT': Keys.SHIFT,
+            'SPACE': Keys.SPACE,
+            'SUBTRACT': Keys.SUBTRACT,
+            'TAB': Keys.TAB,
+            'UP': Keys.UP
+        }
+        key = map.get(key_name)
```
just-be-dev: I know I mentioned this in the last PR, but you should probably replace this whole map with 

``` python
try:
  return getattr(Keys, key_name)
except:
  # emit proper error here
```

That way you don't have to define this whole redundant map. It's smaller and less maintenance as well. 

Oh, and you should probably give a descriptive error on the assert just so people know why it failed.

just-be-dev:
Sorry it's taking me so long to review this. I haven't forgotten you, I'll get to it soon.

src/Selenium2Library/keywords/_element.py
HelioGuilherme66: I used your suggestion. It works well.
I am having trouble with the commit.

just-be-dev: What kind of trouble? I'll assist as best I can.

HelioGuilherme66:
A new PR will be done.

Issue #414, opened 2015-06-01
add "Wait Until Page Does Not Contain" keyword


laurentbristiel:
I can see that "Wait Until Page Does Not Contain Element" and "Wait Until Element Is Not Visible" were recently added by @molsky and @deiga. Would it be possible that someone adds the "Wait Until Page Does Not Contain" keyword? 

(note that I realise that currently I can do a Wait until keyword succeeds  10s  1s  Page Should Not Contain)

just-be-dev:
Added in #415.

PR #415, opened 2015-06-01
Adds Wait Until Page Does Not Contain method

deiga:
Addressess #414 

My local environment didn't work, so I couldn't verify this method, but it should work :)


just-be-dev:
Hey @deiga, thanks for the PR. 

Could you add an acceptance test and update the changelog? 

Thanks!

deiga:
@zephraph Done!

just-be-dev:
Cool. Good work.

PR #417, opened 2015-06-01
Add read_alert_message and handle_alert keywords

KingWarin:
Added read_alert_message and handle_alert keywords to enhance the handling of alert-messages.
Read_alert_message simply returns the text from an alert without closing it, this enables us to check if the present alert is the alert we are looking for.
Handle_alert jumps in right there... it can be used to dismiss or confirm the alert without any other action.


just-be-dev:
There are already keywords that handle this sort of thing. Did you see [Get Alert Message](http://rtomac.github.io/robotframework-selenium2library/doc/Selenium2Library.html#Get%20Alert%20Message)? I would be okay with modifiying `Get Alert Message` to have a close option that defaults to `True`, but I don't think we should add a new keyword for this functionality. 

We could also modify [Alert Should Be Present](http://rtomac.github.io/robotframework-selenium2library/doc/Selenium2Library.html#Alert%20Should%20Be%20Present) to take an argument what action it should take (such as accept or dismiss). 

What are your thoughts @KingWarin?

KingWarin:
@zephraph 
Problem I tried to solve is that right now you can't check whether the alert message is the one you are expecting.

I.e. An alert was triggered and we want to confirm it if the alert displays "ALERT!" and want to dismiss it if the alert displays "foo bar". 
At the moment the only possibility is to retrieve the alert message by using 'Get Alert Message'-Keyword and if that message is what we are looking for we have to trigger the alert again but confirming it this time (right now one could use 'Confirm Action'-Keyword to reach this goal).

Alert Should Be Present looks like a step in the right direction, but still the alert will be dismissed or (if modified to take a accept-parameter) accepted before it is verified that the alert is the expected alert (or not an expected alert).

Feels somehow itchy to dismiss (/confirm) an alert without knowing it's message.
Hope this clarifies my thoughts.

KingWarin:
@zephraph Do you have any thoughts on this? Or do my thoughts need further clarification?

src/Selenium2Library/keywords/_javascript.py
```diff
@@ -129,19 +129,55 @@ def get_alert_message(self):
         """
         return self._close_alert()
 
+    def read_alert_message(self):
```
just-be-dev: This keyword is redundant. 

Let's modify `get_alert_message` to have a `dismiss` argument that defaults to `True` (for backwards compatibility).

src/Selenium2Library/keywords/_javascript.py
```diff
@@ -129,19 +129,55 @@ def get_alert_message(self):
...
+        """ Returns the text of current JavaScript alert without closing it.
+
+        This keyword will fail if no alert is present. Note that
+        following keywords will fail unless the alert is
+        dismissed by this keyword or another like `Get Alert Message`.
+        """
+        return self._read_alert()
+
+    def handle_alert(self, confirm=False):
```
just-be-dev: This keyword I'm fine with. We need an explicit keyword for dealing with alerts anyway. 

Though I'm not sure I really like the name. Would `Dismiss alert` be a better keyword name?

**Edit**: After some thought I'd like this to be

``` python
def dismiss_alert(self, accept=True):
```

This does what we were doing internally before. If the alert is acceptable, then accept it. If not, dismiss it. If it is not acceptable just ignore it and dismiss. If accept is `False` and it is acceptable, dismiss anyway.

test/acceptance/keywords/javascript.txt
```diff
@@ -24,6 +24,14 @@ Get Alert Message
     Should Be Equal  ${msg}  ALERT!
     Run Keyword And Expect Error  There were no alerts  Get Alert Message
 
+Read Alert Message
+    [Setup]  Go To Page "javascript/alert.html"
+    Click Link  Click Me!
+    ${msg} =  Read Alert Message
```
just-be-dev: This line would be changed to

``` robotframework
${msg} =   Get Alert Message   dismiss=False
```

test/acceptance/keywords/javascript.txt
```diff
@@ -24,6 +24,14 @@ Get Alert Message
...
+    Should Be Equal  ${msg}  ALERT!
+    Alert Should Be Present
```
just-be-dev: Instead of `Alert Should Be Present` I'd like this to be the `Handle Alert`/`Dissmiss Alert` keyword

KingWarin:
@zephraph I'm fine with your suggested modifications.
I added two more commits to meet your suggestions, but I'm not entirely sure I got all off your suggestions on dismiss_alert right.

src/Selenium2Library/keywords/_javascript.py
```diff
@@ -120,28 +120,59 @@ def execute_async_javascript(self, *code):
...
+
+        This keyword will fail if no alert is present. Note that
+        following keywords will fail unless the alert is
+        dismissed by this keyword or another like `Get Alert Message`.
+        """
+        return self._handle_alert(accept)
 
     # Private
 
     def _close_alert(self, confirm=False):
```
just-be-dev: Set the default of confirm to `True`.

src/Selenium2Library/keywords/_javascript.py
```diff
@@ -120,28 +120,59 @@ def execute_async_javascript(self, *code):
...
         try:
             alert = self._current_browser().switch_to_alert()
             text = ' '.join(alert.text.splitlines()) # collapse new lines chars
-            if not confirm: alert.dismiss()
-            else: alert.accept()
             return text
         except WebDriverException:
             raise RuntimeError('There were no alerts')
 
+    def _handle_alert(self, confirm=False):
```
just-be-dev: Here too, please. It will better reflect the behavior that we have now.

just-be-dev:
Overall I'm pretty happy with what you've got here. 

Once you update I'll be more than happy to merge this in.

CHANGES.rst
```diff
@@ -3,6 +3,9 @@ Release Notes
 
 1.7 (unreleased)
```
just-be-dev: Because you're on an older version 1.7 is the only thing listed. We're actually working on 1.7.2 right now, which is the current unreleased version. 

Once you merge with master you'll likely have some conflicts to resolve. In that process you should make sure that you move the changelog changes up to the 1.7.2 block.

KingWarin:
@zephraph I just pushed a commit with which the default values and the changes.rst will be updated.
I think this should be ready for merge now.

just-be-dev:
@KingWarin, the tests failed spectacularly. I'm re-running the tests now, but if that fails we're going to need to see what went wrong and fix that before I can merge your stuff in.

KingWarin:
@zephraph I've taken a short look into the errors on my local environment. For me it looks like the error is caused by the _find_by_xpath method in elementfinder.py (or more special by the browser.find_elements_by_xpath(criteria) call inside that function).

just-be-dev:
Can you merge in master? I think the error that was causing your tests to fail has been fixed (it wasn't your code).

KingWarin:
@zephraph Merged master, ran test local, everything looks fine. Pushed it, still looks fine.
Looks like you're right and it was fixed somewhere else - what a relief I was not messing the whole thing up.

just-be-dev:
Haha, yep. Good job, it looks great.

CHANGES.rst
```diff
@@ -3,6 +3,9 @@ Release Notes
 
 1.7.2
 ----------------
+- Added 'read_alert_message' and 'handle_alert' to enhance the handling of alert messages.
```
just-be-dev: One last thing. Let's update this changelog message to reflect what actually changed. 

So something like `added 'dismiss alert' and changed 'get alert message' to optionally leave alert open`. It doesn't have to be that exactly, but you get the idea.

just-be-dev:
The last thing I'd like you to do is update the changelog. The message on there is somewhat old, and I'd like to clean it up. 

After that I'll merge it in! Thanks for your contribution.

KingWarin:
Done! 
Contributing was a pleasure, maybe I'll drop in again if something comes to my mind again.

just-be-dev:
You seem to have a conflict. Have you tried merging in master?

KingWarin:
Merged again and commited the merge, looks good now.
Sorry for the inconvenience

just-be-dev:
Cool. I'm about to do a release so you'll see this very, very shortly.

emanlove:
Go ahead. Also remember to [add a tag to GIT](https://github.com/rtomac/robotframework-selenium2library/blob/master/BUILD.rst#tagging-a-release) on release. If you need any assistance just ask.

just-be-dev:
Sure thing.

PR #426, opened 2015-06-07
Added possibility to combine index and filename args

aaltat:
If test are run, to get a custom and unique file name each time screenshot is takes, it would require creating global variable in RF data and making sure that variable in unique for each time screenshot is taken. This is now much easier to do with the index param in the Capture Page Screenshot keyword.

This comes quite handy when you using pabot and want to make sure that screenshots do not overlap in names.


aaltat:
And did forget to say that removed the `css` from keyword documentation because it was not actually defined or used anywhere.

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -19,14 +19,22 @@ def capture_page_screenshot(self, filename=None):
...
+        but the path does not exist it will be created.
 
-        `css` can be used to modify how the screenshot is taken. By default
-        the bakground color is changed to avoid possible problems with
-        background leaking when the page layout is somehow broken.
-        """
-        path, link = self._get_screenshot_paths(filename)
+        `index` with index argument it is possible to define custom filename
+        but each time get unique screen capture. Example if pabot is used
+        to run multiple test suites in paraler and correct screenshots should
```
just-be-dev: paraler -> parallel

just-be-dev:
This should probably be the default behavior.

@pekkaklarck, @ombre42 If we take a screen shot and specify the file to save it to but that file already exists, should we append a `-1` to it by default? Or rather, is there any cause where we _wouldn't_ want that behavior to occur?

src/Selenium2Library/keywords/_screenshot.py
aaltat: Fixed the typo

aaltat:
@zephraph Perhaps if you use Wait Until Keyword Succeeds? But that is quite a long shot, in my mind.

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -49,10 +57,14 @@ def capture_page_screenshot(self, filename=None):
 
     # Private
 
-    def _get_screenshot_paths(self, filename):
+    def _get_screenshot_paths(self, filename, index=False):
```
just-be-dev: Change `index` here to `overwrite` and default it to False.

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -49,10 +57,14 @@ def capture_page_screenshot(self, filename=None):
...
         if not filename:
             self._screenshot_index += 1
             filename = 'selenium-screenshot-%d.png' % self._screenshot_index
+        elif filename and index:
```
just-be-dev: This will be `elif filename and not overwrite`

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -19,14 +19,22 @@ def capture_page_screenshot(self, filename=None):
...
-        `css` can be used to modify how the screenshot is taken. By default
-        the bakground color is changed to avoid possible problems with
-        background leaking when the page layout is somehow broken.
-        """
-        path, link = self._get_screenshot_paths(filename)
+        `index` with index argument it is possible to define custom filename
+        but each time get unique screen capture. Example if pabot is used
+        to run multiple test suites in paraller and correct screenshots should
+        be copied to log, then it is possible to use `filename` and
+        `index` argument to gether to create unique name for each screenshot.
```
just-be-dev, line 28: I'd update this documentation to say something like 

> By default, if `filename` is specified but a screenshot of that name already exists, an index will be appended to      the screenshot. For example, if 'capture.png' exists then the new screenshot will be 'capture-1.png.' This  behavior can be overridden by setting `overwrite` to True.

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -10,7 +10,7 @@ def __init__(self):
 
     # Public
 
-    def capture_page_screenshot(self, filename=None):
+    def capture_page_screenshot(self, filename=None, index=False):
```
just-be-dev: `index` -> `overwrite`, keep the default to `False`

just-be-dev:
@aaltat, I've made suggestions on how I think it should be. 

I'd like it to always append the index unless it's specified not to. Also, Instead of having the argument `index`, I'd rather it be `overwrite` (so it more clearly reflects what it does).

pekkaklarck:
I'd say that, in general, if you give an exact file and that file exist, it should be overwritten. Not certain about this particular case, though. Having an option to control the behavior is obviously the most generic solution.

just-be-dev:
I suppose the default doesn't really matter to me so long as its well documented.

pekkaklarck:
FWIW, `overwrite=False' sounds good to me as the default. I don't feel too strongly about this either way.

aaltat:
I feel also that false is better, I will try to do changes today or tomorrow.

src/Selenium2Library/keywords/_screenshot.py
aaltat, line 28: Hmm, the current code does not actually do this, but by quick thinking, doing this should not be too difficult. And at least in my mind, it would make life easier. 

So is this what we want, at least it is good for me (and makes my life even more simpler):

```
| Screen Capture | filename=${BROWSER}.png |
| Screen Capture | filename=${BROWSER}.png |
| File Should Exist  | ${OUTPUTDIR}${/}${BROWSER}.png |
| File Should Exist  | ${OUTPUTDIR}${/}${BROWSER}-1.png |
| Screen Capture |
| Screen Capture |
| File Should Exist  | ${OUTPUTDIR}${/}selenium-screenshot-2.png |
| File Should Exist  | ${OUTPUTDIR}${/}selenium-screenshot-3.png |
| Screen Capture | filename=overwrite-${BROWSER}.png | overwrite=${True} |
| Screen Capture | filename=overwrite-${BROWSER}.png | overwrite=${True} |
| File Should Exist  | ${OUTPUTDIR}${/}overwrite-${BROWSER}.png | 
| File Should Not Exist | ${OUTPUTDIR}${/}overwrite-${BROWSER}-4.png | 
```

just-be-dev, line 28: That's close to what I was thinking. The only thing about that I'm not that fond of is the fact that there's a global counter. I would much rather each file have it's own count, though I know that would be a little bit more work. 

Essentially the flow would be like this.

``` python
# This code is more psudo than an actual implementation
if not overwrite:
  files = listdir(path)
  file = filename
  while file in files:
    file = filename + '-' + index + extension
    index += 1
# write file
```

That's only slightly more complicated, but it works more constantly to how I would expect it to. 

Your test would then be like this

``` robotframework
| Screen Capture | filename=${BROWSER}.png |
| Screen Capture | filename=${BROWSER}.png |
| File Should Exist  | ${OUTPUTDIR}${/}${BROWSER}.png |
| File Should Exist  | ${OUTPUTDIR}${/}${BROWSER}-1.png |
| Screen Capture |
| Screen Capture |
| File Should Exist  | ${OUTPUTDIR}${/}selenium-screenshot-1.png |
| File Should Exist  | ${OUTPUTDIR}${/}selenium-screenshot-2.png |
| Screen Capture | filename=overwrite-${BROWSER}.png | overwrite=${True} |
| Screen Capture | filename=overwrite-${BROWSER}.png | overwrite=${True} |
| File Should Exist  | ${OUTPUTDIR}${/}overwrite-${BROWSER}.png | 
| File Should Not Exist | ${OUTPUTDIR}${/}overwrite-${BROWSER}-1.png | 
```

aaltat:
Changed code to more clearly to do what @zephraph expressed in the keyword documentation comment.

And now there is a global and common counter for all files and I also do not like it. 

But I was thinking that could be separate pull request and perhaps add more formatting to the counter?   Not like 1,2,3 but like 001, 002 and so on.

aaltat:
And now the acceptance test are, for this new feature, in one blob/test which is not good. Should I refactor that be more clear?

just-be-dev:
I'm still not 100% satisfied with this, but you're definitely on the right track. I'm going to submit a PR to your branch with what I had in mind.

aaltat:
Refactored the code to contain unique index for each file name. Also changed to acceptance test be more readable. 

Is this OK?

```
| Open Browser | www.someurl.com | browser=${BROWSER} |
| Screen Capture | overwrite=${True} | # overwrite is ignored |
| Screen Capture | overwrite=${True} | # overwrite is ignored |
| File Should Exist  | ${OUTPUTDIR}${/}selenium-screenshot-1.png |
| File Should Exist  | ${OUTPUTDIR}${/}selenium-screenshot-2.png |
```

This is how to code now works.

just-be-dev:
I'll review this after work today. 

At first glance though it looks pretty good.

aaltat:
Noticed that I did forget to update the keyword documentation. I am busy for next few days, but I will do it next week.

just-be-dev:
That's fine, the next minor release won't be until at least the end of next week.

aaltat:
The keyword documentation is now fixed, at least for my eyes. Comments?

just-be-dev:
This looks pretty good. It's varies a little from my original intent, but I think it's ample for this problem. Give me a little while to review it. If I don't find anything I'll merge it in later this afternoon. 

Thanks for all your hard work @aaltat.

aaltat:
How does it looks and should I do something because master has moved forward?

just-be-dev:
Oh yeah, this'll be interesting... I've done a bit of work on screenshots recently so there will be a bit to merge.

aaltat:
I squashed all my commits to a single commit and fixed all the conflicts with master. Now all acceptance test are passing (in windows and Firefox), but because because of some git magic, this can be found from #465. Should I close this pull request and should we continue on the #465?

just-be-dev:
Yeah, let's just close this in favor of #465

PR #428, opened 2015-06-16
screenshots are overwritten only if filename is excplicitly given

orlof:
I propose a change that Capture Page Screenshot would not overwrite existing files unless filename is explicitly given. This would make it easier to re-run failed test cases and combine the logs with rebot.


just-be-dev:
We'll try to address this in #426.

PR #430, opened 2015-06-17
Get Element width and height can be usefull

SergiuTudos:


src/Selenium2Library/keywords/_element.py
```diff
@@ -290,6 +290,33 @@ def get_horizontal_position(self, locator):
             raise AssertionError("Could not determine position for '%s'" % (locator))
         return element.location['x']
 
+    def get_element_height(self, locator):
+        """Returns hight of element identified by `locator`.
+            
+        The position is returned in pixels,as an integer.
```
just-be-dev: Is this supposed to be _height_ instead of position?

SergiuTudos: for Position there is Get Vertical Position and Get Horisontal Position.
These two return the coordinates for top-left corner.

with my requested methods users will be able to get size for an element.
This way, users will be able to to some elements style testing.

SergiuTudos: Thanks zerhpaph, I updated comments.

HelioGuilherme66:
Could you please add positive and negative tests to the acceptance tests? Are the .size['height'] and .size['width'] guaranteed to exist?

just-be-dev:
As @HelioGuilherme66 pointed out, we'll need tests to go along with these keywords before we can merge it in.

SergiuTudos:
Can you please approve it's correct test and I'll commit changes:
Get Element Width
    ${width}=  Get Element Width  link=Link
    Should be True  ${width} > ${0}
    Run Keyword And Expect Error  Could not determine position for 'non-existent'  Get Element Width  non-existent

Get Element Height
    ${height}=  Get Element Height  link=Link
    Should be True  ${height} > ${0}
    Run Keyword And Expect Error  Could not determine position for 'non-existent'  Get Element Height  non-existent

aaltat:
I am just wondering, why there is two keywords? For me it is hard to image use case where I would need just the height or width.

I am not against returning the height and width, but would it be better to return them both with single keyword?

SergiuTudos:
for me it's more readable in robot test.

aaltat:
In which way? Could you write some examples, explain better or something, to prove your point.

aaltat:
For my eyes, it is most easiest to do it like this:
`| ${width} | ${height} = | Get Element Size | locator=some_locator_here |`

Instead of doing it like this:

```
| ${locator} = | Set Variable | locator=some_locator_here |
| ${width} = | Get Element Width | ${locator} |
| ${height} = | Get Element Height | ${locator} |
```

or this:

```
| ${width} = | Get Element Width | locator=some_locator_here |
| ${height} = | Get Element Height | locator=some_locator_here |
```

Either one will copy/paste the _locator_ two times and copy/paste should be avoided in all times. Or one puts the _locator_ in a variable and then you have three keywords instead of one. Also there is small performance penalty, the DOM is searched two times, if you one uses two separate keywords. But if one uses one keyword to get the both width and height, the DOM is searched only one time.

just-be-dev:
Ohh, I like that much better. I didn't even know you could return multiple values!

Makes sense since python does it, but I just never thought to do so.

SergiuTudos:
can someone help me with conflicts?

aaltat:
GitHub has good documentation: https://help.github.com/articles/resolving-a-merge-conflict-from-the-command-line/

SergiuTudos:
I still believe it's configuration issue because of my access rights. Can one of collaborators look into this?

aaltat:
This has nothing to do with access rights. Your branch is quite old and it is quite common that there is merge conflicts when one tries to propose an old history to be pushed to master. Conflict usually means that someone else has changed the same file(s), in master, which you have also changed in your commit and git is not to be able to merge the two independent changes. In that case you must resolve the conflicts manually and push new commit for review.

But this case this is not the only reason. In 1.7.4 release we did have to [revert one of commits](https://github.com/robotframework/Selenium2Library/commit/75e70a033c5b06d703721a2ace2e23a872bf06ef), in practice this means that the commit(s) was removed from the master. But in your branch, the commit reverted in 1.7.4 release does still exist and this is causing the conflict. Also there seems to few other commits which have changed same files which you are trying to change. 

But do not worry, there are many ways to resolve the problem. Perhaps I would do this:
1) `git checkout -b new_branch`
2) `git fetch upstream` I am assuming that you did name this repository as `upstream` in your local repository but name could be anything. So check it by `git remote -v` command.
3) `git merge master` 
4) Resolve conflict, at least you have to copy/paste the tests from `elements.txt` to `elements.robot` and `git rm path/to/elements.txt`
5) `git commit`
6) And make new pull request. 

Please note that will leave your master as it is and you have fix it in others means.

SergiuTudos:
Invalidating this one. Will create request for Get Size separately

PR #433, opened 2015-06-25
Changed test extensions to .robot

### LINKED ISSUE #312: Rename acceptance tests to use `.robot` extension instead of `.txt`

just-be-dev:
Fixes #312


just-be-dev:
I typo'd the commit message. `resource.txt` was renamed to `resource.robot` so all references were corrected.

PR #443, opened 2015-07-13
Added screenshot_root_directory

just-be-dev:
This enhancement is to better support [PageObjects](https://github.com/ncbi/robotframework-pageobjects) as discussed [here](https://github.com/ncbi/robotframework-pageobjects/issues/54#issuecomment-121035194).

These changes are not necessarily final and reviews are welcome. 

This PR would essentially provides a new argument for S2L called `screenshot_root_directory` which can be provided to set the default root directory that screenshots will be stored in. Subdirectories should still be managed by the relevant keywords. 

By default the `screenshot_root_directory` argument is set to `None` and will use RF's log directory as the root directory for screenshots. If one is trying to run a screenshot keyword while RF isn't running and fails to provide `screenshot_root_directory` then the current working directory will be used instead. 
## Pre-merge checklist
- [x] Import `RobotNotRunningError`
- [x] Add a test to use `screenshot_root_directory`
- [x] Add Documentation
- [x] Update changelog

### DIFF
--- a/src/Selenium2Library/__init__.py
+++ b/src/Selenium2Library/__init__.py
@@ -162,6 +162,9 @@ class Selenium2Library(
         `Register Keyword To Run On Failure` keyword for more information about this
         functionality.
 
+        `screenshot_root_directory` specifies the default root directory that screenshots should be
+        stored in. If not provided the default directory will be where robotframework places its logfile.
+
         Examples:
         | Library `|` Selenium2Library `|` 15                                            | # Sets default timeout to 15 seconds                                       |
         | Library `|` Selenium2Library `|` 0 `|` 5                                       | # Sets default timeout to 0 seconds and default implicit_wait to 5 seconds |
--- a/src/Selenium2Library/__init__.py
+++ b/src/Selenium2Library/__init__.py
@@ -135,7 +135,13 @@ class Selenium2Library(
     ROBOT_LIBRARY_SCOPE = 'GLOBAL'
     ROBOT_LIBRARY_VERSION = VERSION
 
-    def __init__(self, timeout=5.0, implicit_wait=0.0, run_on_failure='Capture Page Screenshot'):
+    def __init__(self,
+                 timeout=5.0,
+                 implicit_wait=0.0,
+                 run_on_failure='Capture Page Screenshot',
+                 screenshot_root_directory=None
+    ):
+
         """Selenium2Library can be imported with optional arguments.
 
         `timeout` is the default timeout used to wait for all waiting actions.
@@ -165,6 +171,7 @@ class Selenium2Library(
         """
         for base in Selenium2Library.__bases__:
             base.__init__(self)
+        self.screenshot_root_directory = screenshot_root_directory
         self.set_selenium_timeout(timeout)
         self.set_selenium_implicit_wait(implicit_wait)
         self.register_keyword_to_run_on_failure(run_on_failure)
--- a/src/Selenium2Library/keywords/_logging.py
+++ b/src/Selenium2Library/keywords/_logging.py
@@ -1,6 +1,6 @@
 import os
 import sys
-from robot.libraries.BuiltIn import BuiltIn
+from robot.libraries.BuiltIn import BuiltIn, RobotNotRunningError
 from robot.api import logger
 from keywordgroup import KeywordGroup
 
--- a/src/Selenium2Library/keywords/_logging.py
+++ b/src/Selenium2Library/keywords/_logging.py
@@ -12,7 +12,17 @@ class _LoggingKeywords(KeywordGroup):
         logger.debug(message)
 
     def _get_log_dir(self):
-        variables = BuiltIn().get_variables()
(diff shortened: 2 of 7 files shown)


src/Selenium2Library/keywords/_logging.py
```diff
@@ -12,7 +12,17 @@ def _debug(self, message):
...
-        variables = BuiltIn().get_variables()
+        # Use screenshot root directory if set
+        if self.screenshot_root_directory is not None:
+            return self.screenshot_root_directory
+
+        # If robotframework isn't running use current directory
+        try:
+            variables = BuiltIn().get_variables()
+        except RobotNotRunningError:
+            return os.getcwd()
```
pekkaklarck: 1. `RobotNotRunningError` doesn't seem to be imported.
2. Why not just return `.` instead of `os.getcwd`? In most cases the end result is the same.
3. I was thinking  `screenshot_root_directory` would be returned only if Robot isn't running, but always returning it if defined may be a useful feature in general.

src/Selenium2Library/__init__.py
```diff
@@ -135,7 +135,13 @@ class Selenium2Library(
     ROBOT_LIBRARY_SCOPE = 'GLOBAL'
     ROBOT_LIBRARY_VERSION = VERSION
 
-    def __init__(self, timeout=5.0, implicit_wait=0.0, run_on_failure='Capture Page Screenshot'):
+    def __init__(self,
+                 timeout=5.0,
+                 implicit_wait=0.0,
+                 run_on_failure='Capture Page Screenshot',
+                 screenshot_root_directory=None
+    ):
```
pekkaklarck, line 143: This new argument needs to be documented.

src/Selenium2Library/keywords/_logging.py
just-be-dev: 1. Whoops, nice catch. I haven't written a test for this yet, but I certainly will before its merged in. 
2. Simply because I like to be explicit. 
3. If we're going to add it, we may as well make it available to everyone. I actually added it as an argument into the S2L constructor so it'd be a newly documented feature.

src/Selenium2Library/__init__.py
just-be-dev, line 143: Yep, it's on the todo list.

just-be-dev, line 143: @pekkaklarck should I use backticks or single quotes? S2L isn't very consistent on its documentation style.

pekkaklarck, line 143: In Robot's own libs we have recently switched to use double backticks to get "code style" formatting. This is what we also recommend in Libdoc's docs:
http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#representing-arguments

We used single backticks earlier but nowadays we prefer to reserve it for internal linking. Code style looks better (that's obviously subjective) and avoids accidentally creating a link.

Hmm, should I add a note about this to Robot's newly added contribution guidelines? robotframework/robotframework#1805

just-be-dev, line 143: Yes, please.

just-be-dev:
@pekkaklarck, do you have any advice on testing this feature? Is there a way to "unload" a library?

pekkaklarck:
Would `Reload Library` keyword, available since RF 2.9 alpha 2, work in this case? See robotframework/robotframework#293 and http://robotframework.org/robotframework/2.9rc1/libraries/BuiltIn.html#Reload%20Library for details.

pekkaklarck:
Oooops, sorry for accidentally closing this. Having close and comment buttons next to each others is super annoying.

just-be-dev:
I don't think so... I may just create a keyword to change the screenshot directory from within the test... Though that could lead to issues I suppose...

Hmmm.. I'll think on it.

pekkaklarck:
You should actually be able to import the library with different arguments. You just need to use the `WITH NAME` syntax to give custom name to the libraries you import. You'll also get a conflict with each keyword, so you need to use the `LibName.Keyword Name` syntax with each keyword or use `Set Library Search Order` to make one of the imported libraries the default library.

Having a keyword to change the screenshot directory might still be a good idea, though.

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -1,14 +1,35 @@
...
+
+from Selenium2Library import utils
 from keywordgroup import KeywordGroup
+from robot.libraries.BuiltIn import BuiltIn, RobotNotRunningError
 
 class _ScreenshotKeywords(KeywordGroup):
 
     def __init__(self):
         self._screenshot_index = 0
+        self._screenshot_path_stack = []
```
just-be-dev, line 13: It's likely that this could cause an issue. If the same scope sets two screenshot paths then only the first would be popped off when the scope exited. That could cause an issue where the wrong path is used for other scopes... 

**EDIT**: I've decided to go ahead and use what I've got. Being that this is a private part of the API I can solidify this in the next release.

PR #446, opened 2015-07-16
Removing unused import

dpsfrishberg:
I was looking at this file and noticed an unused import.


just-be-dev:
Nice catch! Thanks.

PR #452, opened 2015-07-21
Use is not instead of != to check for None

hgarus:
Otherwise things blow up with selenium > 2.26 since they override `WebElement.__eq__`

### DIFF
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -757,7 +757,7 @@ return !element.dispatchEvent(evt);
         return (parts[0], parts[2])
 
     def _is_element_present(self, locator, tag=None):
-        return (self._element_find(locator, True, False, tag=tag) != None)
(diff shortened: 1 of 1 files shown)


just-be-dev:
Thanks for this! I'm assuming this is what was causing everything to break?

hgarus:
You're welcome. 

Not sure about "everything breaks". We just had some problems with a failing Page Should Contain in the tests for collective.glossary, but I would expect everything using `_is_element_present` (which seems to be most of the Element Keywords) to be broken with selenium >= 2.26 before this change, reporting the error `AttributeError: 'NoneType' object has no attribute 'id'`.

just-be-dev:
Ha, sorry, in hindsight that was rather vague. I've noticed a lot of our tests have been failing for the last few days with errors like `AttributeError: 'NoneType' object has no attribute 'id'`. It seems yours didn't, so I'm assuming this fixed that issue.

PR #450, opened 2015-07-20
add keyword input_text_into_prompt

ghost:
add keyword input_text_into_prompt ( example from boakley )
fixes issue #292 "Add ability to type into a prompt"

PR #457, opened 2015-07-22
Prepare for 1.7.2 release

just-be-dev:


just-be-dev:
@emanlove, there's an issue with #450 that's causing the test to fail. I've corrected one issue that I found (which I missed in the review). I've got to go to lunch, but I'll take a fresh look at this when I get back. 

After this is cleared up I'm ready to release.

just-be-dev:
Hopefully this will shore up any errors that there were and we can push out the release :pray:

just-be-dev:
Looks like we're good! I'm going ahead with the release.

just-be-dev:
... or not. 

@emanlove, There are five tests failing from my machine from the `Acceptance.Multiple Browsers` testsuite. 

I'm getting 

```
WindowsError: [Error 145] The directory is not empty: 'c:\\users\\USER\\appdaa\\local\\temp\\tmpzvunnb.webdriver.xpi\\components'
```

I think this is likely because of my anti-virus though... will you run the suites and make sure everything is okay on your machine?

**EDIT**: I can't reproduce this on any other environment. I hate my computer.

Issue #467, opened 2015-07-25
Consider adding more automated test coverage to the Travis builds


just-be-dev:
To prevent issues like [this](https://github.com/rtomac/robotframework-selenium2library/commit/e43e782a5a0173a4dce2d60ce5f46e6ce3b78f78) from occurring, we really need greater test coverage. 

This includes determining what versions of RF that we will support and running the tests for each of those versions. 

We also need better browser coverage. I mentioned using SauceLabs in #383, and I still think we should investigate that option. It seems there a run limit before you hit a paywall though, so maybe investigating other options would be good too.

pekkaklarck:
Sauce is promotinc RF and S2L via blog posts and video tutorials. I'm fairly sure they'd give us free credits if asked.

just-be-dev:
Hmmm... I may have been wrong about that. 

I have an account, but it turns out it wasn't for their open source program? I feel like they've transitioned to a new payment scheme recently. Regardless, I'll keep that in #383.

Issue #383, opened 2015-05-02
Implement Testing on SauceLabs


just-be-dev:
Right now we're only testing against firefox. We need to integrate Travis with SauceLabs so that we can get a larger browser coverage.

aaltat:
This is actually done now days, closing issue

PR #247, opened 2013-11-05
Added Safari browser support.

emanlove:

PR #468, opened 2015-07-25
Added saucelabs config to travis

just-be-dev:
Related to #467, #383, and #247 

The start to adding sauce-labs to our CI process to get mutli-browser coverage.


just-be-dev:
Argh. I've made progress, but it isn't quiet there yet. 

I believe I need to pass in the browser to desired capabilities now... but I can't pass in a dictionary through the command line so I don't think I can set that through `run_tests.py`. Even though I'm using 2.9rc1, I can't use the dictionary variable type because of backwards compatibility reasons... which makes me wonder what we'll do after that's deprecated... 

Anyway, I'm off to bed. If anyone feels like poking around on this PR and giving me hints, I'd appreciate it!

markwinspear:
these can be set on the command line... -v DESIRED_CAPABILITIES =  name:Win8 + Chrome 43,platform:Windows 8.1,browserName:chrome,version:43
so in the CI server, if this is executed as a batch file, just have a number of jobs with varying desired capability strings.
This page generates required strings for the capabilities: https://docs.saucelabs.com/reference/platforms-configurator/

just-be-dev:
As a note to myself in the future: Use [yaml variable files](http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#variable-file-as-yaml)!

just-be-dev:
I'm no longer very active on this project and can't implement this feature.

Issue #469, opened 2015-07-27
Documentation error for desription of "Using webelement"


tarand:
The keyword documentation has:
Starting with version 1.7 of the Selenium2Library, one can pass an argument that contains a WebElement instead of a string locator. To get a WebElement, use the new Get WebElements keyword. For example:
${elem} =   Get WebElements     id=my_element
Click Element   ${elem}     

This can obviously not work, as Get WebElements returns a list. I think it should be
${elements} =   Get WebElements     id=my_element
Click Element   ${elements}.first()
Or similar

just-be-dev:
Ugh, yeah, you're right. We'd planned on adding a `Get WebElement` keyword that just returned a single item. That slipped through the cracks though.

I'll look at rewording the documentation and possibly introducing that keyword.

just-be-dev:
Should I add `Get WebElement`or just suggest people use `Get WebElements` combined with `@{list}[0]`?

just-be-dev:
@pekkaklarck, is there any precedence for this? Having two keywords, one that returns a list of items and the other that returns just the first.

pekkaklarck:
I think it would be convenient to have `Get WebElement` separately.

For example, XML library has both `Get Elements` and `Get Element` keywords. The former returns a list of elements and the list can be empty. The latter fails if there is not exactly one element. Would that make sense in S2L context?

tarand:
I would say, yes.

Thomas

---

Thomas Arand
Tel.  +49 (89) 5159 36328 / +49 (160) 90888023
thomas.arand@nokia.commailto:thomas.arand@nsn.com

Nokia Solutions and Networks Deutschland GmbH

just-be-dev:
I thought it was needed originally, and I still think it'd be nice to have.

just-be-dev:
Fixed in 1.7.3

PR #356, opened 2015-02-02
Allow keywords to use Web Elements in addition to Locator.

emanlove:

PR #473, opened 2015-07-28
Added keyword 'Get WebElement'

just-be-dev:
This may or may not be merged. 

@emanlove, we briefly discussed this in #356 and I just wanted to broach the issue again. It was raised in #469 that the documentation using `Get WebElements` was wrong. I've corrected it in #472, but I'm using `Get WebElement` which obviously doesn't exist. 

We could of course just correct the documentation and only keep `Get WebElements`. 

It would be this

``` robotframework
@{elements}=   Get WebElements  id=my_element
Click Element  @{elements}[0]
```

versus

``` robotframework
${element}=  Get WebElement  id=my_element
Click Element  ${element}
```

@pekkaklarck, feel free to jump in on this conversation too.

### DIFF
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -21,6 +21,13 @@ class _ElementKeywords(KeywordGroup):
 
     # Public, get element(s)
 
+    def get_webelement(self, locator):
+        """Returns the first WebElement matching the given locator.
+
+        See `introduction` for details about locating elements.
+        """
+        return self.get_webelements(locator)[0]
+
     def get_webelements(self, locator):
         """Returns list of WebElement objects matching locator.
 
@@ -599,7 +606,7 @@ return !element.dispatchEvent(evt);
         """Returns number of elements matching `xpath`
 
         One should not use the xpath= prefix for 'xpath'. XPath is assumed.
-        
+
         Correct:
         | count = | Get Matching Xpath Count | //div[@id='sales-pop']
         Incorrect:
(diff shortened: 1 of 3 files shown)


src/Selenium2Library/keywords/_element.py
```diff
@@ -21,6 +21,13 @@ def __init__(self):
 
     # Public, get element(s)
 
+    def get_webelement(self, locator):
+        """Returns the first WebElement matching the given locator.
+
+        See `introduction` for details about locating elements.
+        """
+        return self.get_webelements(locator)[0]
```
emanlove: Do we want to do this or follow along with what the other find element keywords do? Something like ...

```
        return self._element_find(locator, True, True)
```

where we set the first_only flag to True. I like this a little better if only for the reason that it follows what other find elements are doing and if this changing in the future Get WebElement will do the same thing.

test/acceptance/keywords/elements.robot
```diff
@@ -7,6 +7,15 @@ Get Elements
     @{links}=  Get WebElements  //div[@id="div_id"]/a
     Length Should Be  ${links}  11
 
+Get Web Element
+    @{links}=  Get WebElements  //div[@id="div_id"]/a
+    ${link}=  Get WebElement  //div[@id="div_id"]/a
+    LOG  @{links}[0]
+    LOG  ${link}
```
just-be-dev: I forgot and left these logs in here. 

We should take those out.

just-be-dev:
I'll update this to remove the log statements in the test and I suppose we should add to the changelog as well. 

Anything else to add @emanlove?

emanlove:
@zephraph Did you remove the logs? Looks like there are still there...

just-be-dev:
I haven't even submitted another commit yet. You can go ahead and do it if you'd like.

emanlove:
Done

just-be-dev:
Sorry, that commit didn't work right. 

I was in the middle of committing the same time you did, so I pulled but it ended up doing something wanky.

just-be-dev:
Regardless, I think this is good to go. Do you want to look over it? I'll let you merge when you feel it's ready.

PR #474, opened 2015-07-28
Add named keys to Press Key.

### LINKED ISSUE #275: Press Key does not support non-ASCII, non-typeable characters such as ARROW_RIGHT
Current implementation of Press Key does not seem to support sending an ARROW_RIGHT press to the webdriver.

Example, web app that allows navigation via up-down-left-right keys.

Registered keycode in app for right arrow is "39" but this is not the ASCII code 39 ('). Sending "Press Key <ELEMEND_ID> \39" just results in a ' being typed. Can't find any constant table to reference for code for arrow key nor does there seem to be any "native key" command or any other code set (apart from ASCII) with which to enter an arrow key.

Cannot test normal keyboard navigation of web app.

Using Selenium2Library with robotframework on Fedora Linux. Tried to get this going under Firefox & Chrome but now don't think that it's possible.

HelioGuilherme66:
Adds named keys escaped with double backslash, for example \\ARROW_LEFT.
Also normalized test file.
Fixes #275 .

### DIFF
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -483,10 +483,13 @@ return !element.dispatchEvent(evt);
         lead by '\\\\'.
 
         Examples:
-        | Press Key | text_field   | q |
-        | Press Key | login_button | \\\\13 | # ASCII code for enter key |
+        | Press Key | text_field   | q                |                                               |
+        | Press Key | login_button | \\\\13           | # ASCII code for enter key                    |
+        | Press Key | nav_console  | \\\\\\\\ARROW_UP | # selenium.webdriver.common.keys ARROW_UP KEY |
         """
-        if key.startswith('\\') and len(key) > 1:
+        if key.startswith('\\\\') and len(key) > 1:
+            key = getattr(Keys,key[2:])
+        elif key.startswith('\\') and len(key) > 1:
             key = self._map_ascii_key_code_to_key(int(key[1:]))
         #if len(key) > 1:
         #    raise ValueError("Key value '%s' is invalid.", key)
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -479,20 +479,27 @@ return !element.dispatchEvent(evt);
     def press_key(self, locator, key):
         """Simulates user pressing key on element identified by `locator`.
 
-        `key` is either a single character, or a numerical ASCII code of the key
-        lead by '\\\\'.
+        `key` is either a single character, a numerical ASCII code of the key,\
+         or a NAMED KEY as described at https://selenium.googlecode.com/git/docs/api/py/webdriver/selenium.webdriver.common.keys.html
 
         Examples:
-        | Press Key | text_field   | q                |                                               |
-        | Press Key | login_button | \\\\13           | # ASCII code for enter key                    |
-        | Press Key | nav_console  | \\\\\\\\ARROW_UP | # selenium.webdriver.common.keys ARROW_UP KEY |
+        | Press Key | text_field   | q        |                                               |
+        | Press Key | login_button | \\\\13   | # ASCII code for Enter key (DEPRECATED)       |
+        | Press Key | login_button | 10       | # ASCII code for Return key                   |
+        | Press Key | nav_console  | ARROW_UP | # selenium.webdriver.common.keys ARROW_UP KEY |
+        
+        NAMED KEY value is new in Selenium2Library 1.7.3.
         """
-        if key.startswith('\\\\') and len(key) > 1:
-            key = getattr(Keys,key[2:])
-        elif key.startswith('\\') and len(key) > 1:
-            key = self._map_ascii_key_code_to_key(int(key[1:]))
-        #if len(key) > 1:
-        #    raise ValueError("Key value '%s' is invalid.", key)
+        if len(key) > 1:
+            if key.startswith('\\'):
+                self._warn("Press Key: Escaped ASCII codes are deprecated. Use plain numeric value: '%s'" % (key[1:]))
+                key = self._map_ascii_key_code_to_key(int(key[1:]))
+            else:
+                try:
+                    key = (key.isdecimal() and self._map_ascii_key_code_to_key(int(key))) or\
+                          ((not key.isdecimal()) and self._map_named_key_code_to_special_key(key))
+                except:
+                    raise ValueError("Key value '%s' is invalid." % (key))
         element = self._element_find(locator, True, True)
         #select it
         element.send_keys(key)
@@ -750,6 +757,15 @@ return !element.dispatchEvent(evt);
         if key is None:
             key = chr(key_code)
         return key
+
+    def _map_named_key_code_to_special_key(self, key_name):
+        try:
+           return getattr(Keys, key_name)
+        except:
+           message = "Unknown key named '%s'." % (key_name)
+           self._debug(message)
+           raise ValueError(message)
+        return Keys.NULL
 
     def _parse_attribute_locator(self, attribute_locator):
         parts = attribute_locator.rpartition('@')
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -480,12 +480,14 @@ return !element.dispatchEvent(evt);
         """Simulates user pressing key on element identified by `locator`.
 
         `key` is either a single character, a numerical ASCII code of the key lead by '\\\\',
-         or a NAMED KEY as described at https://selenium.googlecode.com/git/docs/api/py/webdriver/selenium.webdriver.common.keys.html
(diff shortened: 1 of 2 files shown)


src/Selenium2Library/keywords/_element.py
```diff
@@ -483,10 +483,13 @@ def press_key(self, locator, key):
...
 
         Examples:
-        | Press Key | text_field   | q |
-        | Press Key | login_button | \\\\13 | # ASCII code for enter key |
+        | Press Key | text_field   | q                |                                               |
+        | Press Key | login_button | \\\\13           | # ASCII code for enter key                    |
+        | Press Key | nav_console  | \\\\\\\\ARROW_UP | # selenium.webdriver.common.keys ARROW_UP KEY |
         """
-        if key.startswith('\\') and len(key) > 1:
+        if key.startswith('\\\\') and len(key) > 1:
```
zephraph: Is it possible we could look for something other than four slashes?

If there's a double slash is it always numbers?

zephraph: Nevermind! I get it, heh.

src/Selenium2Library/keywords/_element.py
```diff
@@ -483,10 +483,13 @@ def press_key(self, locator, key):
...
+            key = getattr(Keys,key[2:])
+        elif key.startswith('\\') and len(key) > 1:
             key = self._map_ascii_key_code_to_key(int(key[1:]))
```
just-be-dev: What if we did this in this part instead:

``` python
if key.startswith('\\'):
  key = self._map_ascii_key_code_to_key(int(key[1:]))
elif key in Keys:
  key = Keys[key]
elif len(key) > 1:
  raise ValueError("Key value '%s' is invalid.", key)    
```

That way you could just issue `Press Key  |  text_field  |  ARROW_UP`

pekkaklarck:
The original escaping of ASCII codes is, in my opinion, somewhat ugly and double escaping to use WebDriver keys is even uglier. Backslashes don't tell much there in general and Robot requiring them to be escaped makes the syntax really awkward.

What about using prefixes like with locators? Something like `text=foo`, `ascii=10`, and `webdriver=ARROW_UP` with `text=` being the default if there is no prefix could work. We probably could come up with better prefix names but I believe you got the point. This would also make it easy to add more ways to specify keys in the future. The old escaping mechanism to use ASCII codes should still be supported but could perhaps be deprecated.

just-be-dev:
I very much like this!

src/Selenium2Library/keywords/_element.py
```diff
@@ -479,17 +479,27 @@ def simulate(self, locator, event):
...
+        """
+        if len(key) > 1:
+            if key.startswith('\\'):
+                self._warn("Press Key: Escaped ASCII codes are deprecated. Use plain numeric value: '%s'" % (key[1:]))
+                key = self._map_ascii_key_code_to_key(int(key[1:]))
+            else:
+                try:
+                    key = (key.isdecimal() and self._map_ascii_key_code_to_key(int(key))) or\
+                          ((not key.isdecimal()) and self._map_named_key_code_to_special_key(key))
+                except:
```
pekkaklarck: Plain`except` is not a good idea. It can mask bugs and prevents `SystemExit` and `KeyboardInterrupt` to stop execution. Should check what exceptions the code may raise and catch only them. Alternatively could just let those exception pass through if they already have a good error message.

src/Selenium2Library/keywords/_element.py
```diff
@@ -748,6 +758,15 @@ def _map_ascii_key_code_to_key(self, key_code):
             key = chr(key_code)
         return key
 
+    def _map_named_key_code_to_special_key(self, key_name):
+        try:
+           return getattr(Keys, key_name)
+        except:
```
pekkaklarck: Plain `except` isn't a good idea here either. Should have `except AttributeError:` instead.

src/Selenium2Library/keywords/_element.py
```diff
@@ -748,6 +758,15 @@ def _map_ascii_key_code_to_key(self, key_code):
...
+           message = "Unknown key named '%s'." % (key_name)
+           self._debug(message)
+           raise ValueError(message)
+        return Keys.NULL
```
pekkaklarck: This line can never be reached.

src/Selenium2Library/keywords/_element.py
```diff
@@ -479,17 +479,27 @@ def simulate(self, locator, event):
...
+        NAMED KEY value is new in Selenium2Library 1.7.3.
+        """
+        if len(key) > 1:
+            if key.startswith('\\'):
+                self._warn("Press Key: Escaped ASCII codes are deprecated. Use plain numeric value: '%s'" % (key[1:]))
+                key = self._map_ascii_key_code_to_key(int(key[1:]))
+            else:
+                try:
+                    key = (key.isdecimal() and self._map_ascii_key_code_to_key(int(key))) or\
+                          ((not key.isdecimal()) and self._map_named_key_code_to_special_key(key))
```
pekkaklarck: Normal `if/else` would be quite a bit more readable.

src/Selenium2Library/keywords/_element.py
```diff
@@ -479,17 +479,27 @@ def simulate(self, locator, event):
...
-        #if len(key) > 1:
-        #    raise ValueError("Key value '%s' is invalid.", key)
+        | Press Key | text_field   | q        |                                               |
+        | Press Key | login_button | \\\\13   | # ASCII code for Enter key (DEPRECATED)       |
+        | Press Key | login_button | 10       | # ASCII code for Return key                   |
+        | Press Key | nav_console  | ARROW_UP | # selenium.webdriver.common.keys ARROW_UP KEY |
+        
+        NAMED KEY value is new in Selenium2Library 1.7.3.
+        """
+        if len(key) > 1:
```
just-be-dev, line 494: The one thing we have to consider with this approach is that users will no longer be able to specify 0-9 ASCII codes. The only ones that I think would be remotely important in that range are 8 and 9 (backspace and tab, respectively). 

Is this an issue? They could simply use `BACKSPACE` or `TAB` to do the same thing, so I'm not necessarily of the opinion that it really matters.

HelioGuilherme66, line 494: I think that the sending numeric values below 10 should be only possible
with their NAMED values. But keeping the \ would be OK.

just-be-dev, line 494: While I generally agree, @pekkaklarck was right when he said having a different behavior between 9 and 10 could be confusing.

just-be-dev:
This looks good to me now. @pekkaklarck, comments?

@emanlove, see any issues?

src/Selenium2Library/keywords/_element.py
```diff
@@ -479,19 +479,23 @@ def simulate(self, locator, event):
...
-        lead by '\\\\'.
+        `key` is either a single character, a numerical ASCII code of the key lead by '\\\\',
+         or a NAMED KEY as described at https://selenium.googlecode.com/git/docs/api/py/webdriver/selenium.webdriver.common.keys.html
 
         Examples:
-        | Press Key | text_field   | q |
-        | Press Key | login_button | \\\\13 | # ASCII code for enter key |
+        | Press Key | text_field   | q        | # The letter 'q'                              |
+        | Press Key | login_button | \\\\13   | # ASCII code for Enter key                    |
+        | Press Key | nav_console  | ARROW_UP | # selenium.webdriver.common.keys ARROW_UP KEY |
```
pekkaklarck: KEY probably shouldn't be in upper case.

src/Selenium2Library/keywords/_element.py
```diff
@@ -748,6 +752,15 @@ def _map_ascii_key_code_to_key(self, key_code):
...
         return key
 
+    def _map_named_key_code_to_special_key(self, key_name):
+        try:
+           return getattr(Keys, key_name)
+        except AttributeError:
+           message = "Unknown key named '%s'." % (key_name)
+           self._debug(message)
+           raise ValueError(message)
+        return Keys.NULL
```
pekkaklarck: This line is unreachable.

test/acceptance/keywords/textfields.robot
```diff
@@ -1,40 +1,51 @@
-*Setting*
-Variables  variables.py
-Resource  ../resource.robot
-Test Setup  Go To Page "forms/prefilled_email_form.html"
-
-
-*Test Cases*
+*** Setting ***
+Test Setup        Go To Page "forms/prefilled_email_form.html"
+Force Tags        txtfields
```
just-be-dev: I don't think this `Force Tags` is necessary.

HelioGuilherme66: Helps when testing the acceptance tests :).

just-be-dev: You can also use `--suite textfields` to accomplish the same thing. 

That'll only run the tests in the textfields suite.

PR #477, opened 2015-07-31
Added Keyword 'Capture Screenshot Only On Failure'

harperm:
This keyword is very useful when looking back into the logs to see exactly why something failed. It also does not clutter up the logs with a large amount of unnecessary screenshots when you are waiting for a given keyword to succeed, it will simply capture a single screenshot if the keyword does fail.

### DIFF
--- a/src/Selenium2Library/keywords/_screenshot.py
+++ b/src/Selenium2Library/keywords/_screenshot.py
@@ -63,7 +63,7 @@ class _ScreenshotKeywords(KeywordGroup):
 
     def capture_screenshot_only_on_failure(self,keyword,*args):
         """Takes a screenshot of the current page and embeds it into the log
-        when the given ``keyword``  fails. If the given ``keyword`` can run miltiple times
+        when the given ``keyword``  fails. If the given ``keyword`` can run multiple times
         prior to timing out, the screenshot will only be captured on the very
         last failure.
 
--- a/src/Selenium2Library/keywords/_screenshot.py
+++ b/src/Selenium2Library/keywords/_screenshot.py
@@ -63,15 +63,14 @@ class _ScreenshotKeywords(KeywordGroup):
 
     def capture_screenshot_only_on_failure(self,keyword,*args):
         """Takes a screenshot of the current page and embeds it into the log
-        only when a specific keyword fails. If a keyword can run miltiple times
+        when the given ``keyword``  fails. If the given ``keyword`` can run miltiple times
         prior to timing out, the screenshot will only be captured on the very
         last failure.
 
         Example:
         |  Capture Screenshot Only On Failure  |  Wait Until Keyword Succeeds  |  5 min  1 sec  | On Page  |  ${page name}  | """
 
-        nothing = lambda *args: None
-        old_keyword = self.register_keyword_to_run_on_failure(nothing)
+        old_keyword = self.register_keyword_to_run_on_failure('nothing')
         status, value = BuiltIn().run_keyword_and_ignore_error(keyword,*args)
         self.register_keyword_to_run_on_failure(old_keyword)
         if (status=='FAIL'):
--- a/src/Selenium2Library/keywords/_screenshot.py
+++ b/src/Selenium2Library/keywords/_screenshot.py
@@ -61,6 +61,24 @@ class _ScreenshotKeywords(KeywordGroup):
         self._html('</td></tr><tr><td colspan="3"><a href="%s">'
                    '<img src="%s" width="800px"></a>' % (link, link))
 
+    def capture_screenshot_only_on_failure(self,keyword,*args):
+        """Takes a screenshot of the current page and embeds it into the log
+        only when a specific keyword fails. If a keyword can run miltiple times
+        prior to timing out, the screenshot will only be captured on the very
+        last failure.
+
+        Example:
+        |  Capture Screenshot Only On Failure  |  Wait Until Keyword Succeeds  |  5 min  1 sec  | On Page  |  ${page name}  | """
+
+        nothing = lambda *args: None
+        old_keyword = self.register_keyword_to_run_on_failure(nothing)
+        status, value = BuiltIn().run_keyword_and_ignore_error(keyword,*args)
+        self.register_keyword_to_run_on_failure(old_keyword)
+        if (status=='FAIL'):
+            self.capture_page_screenshot()
+            BuiltIn().fail(msg=value)
+        return value
+
     # Private
     def _create_directory(self, path):
         target_dir = os.path.dirname(path)


emanlove:
How is this different that the current 'Capture Page Screenshot' keyword and the handling of failures by 'Register Keyword To Run On Failure' keyword?

fkberthold:
This  is a powerful keyword for extending Robotframework's Selenium library.  While @emanlove is correct that it doesn't add anything that can't already be done. It makes it possible to extend the system much more cleanly both in the written code and in the logs.

There are two primary use cases for this keyword. When you want screenshots to be captured with a non-selenium keyword, and when you want to wait for a selenium keyword to succeed without adding a lot of noise to your logs.

### Expanding Selenium with UserKeywords

A fairly common pattern in my own code is something like this:

```
Element Text Should Be a Date  [ARGUMENTS]  ${locator}
    ${element text}=  Get Text  ${locator}
    ${status}=  Run Keyword and Return Status  Should Match Regexp  ${element text}  ^\d\d/\d\d/\d\d$
    Run Keyword If  not ${status}  Capture Screenshot
    Run Keyword If  not ${status}  Fail
```

This solution has a couple of flaws:
1. It has a pretty clearly repeated pattern, thus if I have several that are similar, I have repetitive code.
2. When this fails, in my logs, what I will see is the failure, with no screenshot.  I have to look up and find the screenshot.  If the person reading my logs doesn't know this pattern, they aren't likely to find the screenshot without effort.

The first problem can be solved by adding the following keyword:

```
Element Text Should Match Regexp  [ARGUMENTS]  ${locator}  ${regexp}
    ${element text}=  Get Text  ${locator}
    ${status}  ${error}=  Run Keyword and Ignore Error  Should Match Regexp  ${element text}  ${regexp}
    Run Keyword If  '${status}' == 'FAIL'  Capture Screenshot
    Run Keyword If  '${status}' == 'FAIL'  ${error}
```

Now you have a keyword that looks similar to a standard Selenium keyword, but it doesn't behave like one, it still doesn't look right in the logs.

Using the keyword suggested in this pull request I'm able to reduce the code to this:

```
Element Text Should Be a Date  [ARGUMENTS]  ${locator}
    ${element text}=  Get Text  ${locator}
    Capture Screenshot Only On Failure  Should Match Regexp  ${element text}  ^\d\d/\d\d/\d\d$
```

This removes most of the repetition, and will behave like other Selenium keywords in the logs.

### Expanding Selenium's 'Wait For' keywords

When I started writing test automation in Robotframework, there were two 'Wait Until' keywords: `Wait Until Page Contains` and `Wait for Condition`. Now we're up to 9 of them, that's because they're useful, especially in the age of JavaScript based web pages. Adding one for every possible use case is burdensome on the library maintainers, and clutters up the documentation. This keyword makes it easier instead to add keywords that behave more like native Selenium keywords.

If I want to, for example, wait for a table to finish populating, I would write the following:

```
Wait Until Table Is Populated  [ARGUMENTS]  ${table locator}  ${expected rows}
    Wait Until Keyword Succeeds  10 sec  0.5 sec  Locator Should Match X Times  ${table locator}//tr  ${expected rows}
```

This will work.  It will also:
1. Fill my directory with screenshots
2. Fill my logs with errors 

I'm aware that we can now tell RobotFramework at run time to ignore certain keywords for logging, but it's a hack, and requires the person running the code to be aware of the problem.  If I'm just using this to wait for a page to load, and I really don't expect it to ever fail, this is a ridiculous clutter to my logs.

You can fix the screenshot problem with:

```
Wait Until Table Is Populated  [ARGUMENTS]  ${table locator}  ${expected rows}
    ${original fail keyword}=  Register Keyword To Run On Failure  Nothing
    ${status}  ${error}=  Run Keyword And Ignore Error  Wait Until Keyword Succeeds  10 sec  0.5 sec  Locator Should Match X Times  ${table locator}//tr  ${expected rows}
    Register Keyword To Run On Failure  ${original fail keyword}
    Run Keyword If  '${status}' == 'FAIL'  Fail  ${error}
```

Again, this works, and has all of the failings of the previous example, it's repetitive, ugly, and doesn't work like a real Selenium keyword.

We could instead use this new keyword, and get all the benefits of a naitve Selenium keyword.

```
Wait Until Table Is Populated  [ARGUMENTS]  ${table locator}  ${expected rows}
    Capture Screenshot Only On Failure  Wait Until Keyword Succeeds  10 sec  0.5 sec  Locator Should Match X Times  ${table locator}//tr  ${expected rows}
```

For these reasons, I think this new keyword will make it easier for Selenium's user's to mold the tool to their needs.

pekkaklarck:
Explanation by @fkberthold is pretty convincing and I can see that this kind of functionality could be useful. I'm looking a bit bigger picture, though, and wonder could such functionality be added to Robot Framework itself. It would generally be nice to be able to _easily_ do something if a keyword fails.

On test case level something can be executed if there has been a failure by using `Run Keyword If Test Failed` in test teardown. We have also been thinking about adding a separate setting to specify an action to run on failure (robotframework/robotframework#720) but decided not to implement it because teardowns and the aforementioned keyword already make it possible. In programming terms you can think "teardown" as a finally block in a try/except/finally structure and this potential "on failure" would be the except block.

Would it possibly make sense to add separate `Run Keyword If Keyword Failed` keyword that could be used in keyword teardowns? Or should we rather enhance `Run Keyword If Test Failed` so that it could be used also elsewhere than test teardowns? If the latter would be done, it would allow the following:

``` robotframework
*** Keywords ***
Wait Until Table Is Populated 
    [Arguments]    ${table locator}    ${expected rows}
    Wait Until Keyword Succeeds    10 sec    0.5 sec    Locator Should Match X Times    ${table locator}//tr    ${expected rows}
    [Teardown]    Run Keyword If Test Failed    Capture Screenshot
```

fkberthold:
I could see @pekkaklarck 's suggestion of making a more general `[Teardown]` setting for actions like taking a screenshot or closing a file that's being worked on.  I think we would have to add a separate `Run Keyword If Keyword Failed` keyword since the test itself hasn't necessarily failed. To fully implement the functionality that I've described you would also have to add a keyword like 'Without Teardown' to suppress the Teardown of contained keywords:

```
Wait Until Table Is Populated  [ARGUMENTS]  ${table locator}  ${expected rows}
    Without Teardown  Wait Until Keyword Succeeds  10 sec  0.5 sec  Locator Should Match X Times  ${table locator}//tr  ${expected rows}
    [Teardown]  Run Keyword If Keyword Failed  Capture Screenshot
```

Without something like this, you'll still have the problem of Selenium capturing as many as 20 screenshots when only one is needed.

Unfortunately this would run into trouble if there were Teardown's that performed cleanup activities such as closing socket connections, or files that were created.  Ensuring that a keyword user was aware of all of the activities performed by a keyword's Teardown would be complicated at best.

I think having a Teardown specific to keywords would be useful in the general case. Unfortunately, for the same reason I think Teardown would be useful, cleanup of files and connections when a keyword fails, it wouldn't work well for this particular case.

pekkaklarck:
Could you clarify when a keyword could fail but a test containing it wouldn't? Are you referring to `Run Keyword And Ignore Error` and similar keywords that can catch a failure make the test pass? It ought to be possible to define `Run Keyword If Test Failed` behavior with them too.

Also, could you clarify why `Without Teardown` would be needed? Why would you want to avoid running keywords teardowns in the first place?

fkberthold:
The case of `Run Keyword And Ignore Error` is exactly the case I'm talking about.  I could see the contained keyword setting the test's PASS/FAIL flag to FAIL, then `Run Keyword And Ignore Error` setting it back to PASS, thus making it possible to just use `Run Keyword If Test Failed`.

As for the need for `Without Teardown`, suppose the Selenium keywords were fully implemented with this Teardown functionality, my case (with the keyword defined as a user keyword) might look like this:

```
Locator Should Match X Times  [ARGUMENTS]  ${locator}  ${expected locator count}
    ${count}=  Get Matching Locator Count  ${locator}
    Should Be Equal As Integers    ${count}    ${expected locator count}
    [TEARDOWN]  Run Keyword If Keyword Failed  Capture Screenshot

Wait Until Table Is Populated  [ARGUMENTS]  ${table locator}  ${expected rows}
    Without Teardown  Wait Until Keyword Succeeds  10 sec  0.5 sec  Locator Should Match X Times  ${table locator}//tr  ${expected rows}
    [Teardown]  Run Keyword If Keyword Failed  Capture Screenshot
```

I would expect `Locator Should Match X Times` to be implemented similarly in the Selenium library, though admittedly as a Python keyword.  In such a case I would want to suppress screenshot capture until `Wait Until Table Is Populated` had either passed or failed. There's no need to capture each screenshot during a "Wait Until" keyword.

pekkaklarck:
1. I still don't get what's the problem with `Run Keyword And Ignore Error` and the idea of making `Run Keyword If Test Failed` to work outside teardowns. I'd design it so that if `Run Keyword If Test Failed` is used inside `Run Keyword And Ignore Error`, it will run if there is a failure, but outside `Run Keyword And Ignore Error` it would newer run (unless there was some other failure). If there would be `Run Keyword If Keyword Failed`, it would behave exactly the same way.
2. I don't understand the need for `Without Teardown` either. If the problem is that you are using a keyword that does something you don't want to always be done inside `Wait Until Keyword Succeeds`, the logical solution would be changing the keyword so that it doesn't do it. I don't see why there should be a keyword to change the behavior. Notice also that no such keyword could influence how keywords implemented in Python behave. Such keywords have no teardowns, that concept exists only on Robot level. Python keywords can use `try/finally` to get similar effect, but we cannot outside tell them to skip `finally`.

fkberthold:
Keep in mind that I'm viewing the idea for a keyword based `Teardown` entirely through the lens of this pull request.  As I said above, I think it would be useful in the general case, but would not make the actions in this pull request any easier to implement.  So with that in mind.
1. There isn't a problem. I was thinking out loud about how you would implement it.
2. The `Without Teardown` keyword would be in reference to the problem that the pull request addresses. This is making a set of assumptions, including:
   1. You won't always be the one writing the libraries that you are using, as is the case with the Selenium2Library, so just not including the screenshot isn't always an option.
   2. Even when you are, you don't necessarily want to write 2 general purpose keywords each time. In this case `Locator Should Match X Times` and `Locator Should Match X Times Without Screenshot`. Obviously that could be dealt with through an argument if you are the one who implemented the library. In either case it will add repetitive code to your Resource/Library files.
   3. That, as with TAGS and BDD style names, it would be possible to also include RF style `Teardown` attributes in python libraries for consistency of behavior between Python Libraries and Robot Framework Resource files.

As I said above, I think `Teardown` would be useful, but I don't think it applies to the problem at hand.

pekkaklarck:
If I got the idea of the proposed keyword right, the following example would do exactly the same if `Run Keyword If Test Failed` would work in keyword teardowns. Notice that keyword teardowns themselves are supported since RF 2.7 or perhaps even 2.6.

``` robotframework
*** Keywords ***
Capture Screenshot Only On Failure
    [Arguments]    ${keyword}    @{args}
    ${original on failure} =    Register Keyword To Run On Failure    Nothing
    Run Keyword    ${keyword}    @{args}
    [Teardown]    Run Keywords
    ...    Run Keyword If Test Failed    Capture Page Screenshot    AND
    ...    Register Keyword To Run On Failure    ${original on failure}
```

Do you @fkberthold see any differences in functionality? Notice that the example would get a lot simpler if screenshots on failure were disabled altogether when importing the library.

aaltat:

aaltat:
Closing due inactivity

PR #480, opened 2015-08-07
New keyword Press Keys

HelioGuilherme66:
This keyword simulates user typing keys by Name, ASCII code or symbol, eventually pressing a modifier key like SHIFT, ALT or CONTROL.
Use with care, because you may see unexpected behavior in browser. For example the browser developer tool might pop up.


pekkaklarck:
A very brief review below. My feeling is that there's too much to fix that 2.7.3 should wait for this.
1. In my opinion `Press Keys` should accept `*keys`, not `keys`. That ought to make usage and also implementation much easier. In the current implementation something like `END` is ambiguous and could mean either single special key `END` or three normal characters `E`, `N`, `D`.
2. There are lot of seemingly unrelated changes to documentation and other keywords in the PR. My guess is that there has been some problem in merging earlier changes. I don't see how adding a single keyword should introduce changes to 40 files.
3. This PR cannot be merged. Might be related to 2.
4. The implementation looks very complicated, most likely at least partly due to 1. There's already things like bare `except:` and code that has been commented out that ought to be fixed.

HelioGuilherme66:
Some fixes done. I am happy to say that these tests all passed with Google Chrome  44.0.2403.125 m (64-bit), Mozilla Firefox ESR 38.1.0, Internet Explorer 11.0.9600.17801 (11.0.19) (32-bit driver) and PhantomJs 2.0.0.

just-be-dev:
I thought I'd left feedback earlier, but it seems it didn't submit or was deleted. 

One thing: I think you may be using `git rebase master` or `git merge --rebase master` when you're trying to update your branch. You should really just use `git merge master`. I _think_ rebasing would cause all these changes to be listed. Is that what you're using?

As @pekkaklarck said, there are way, way too many changes in this PR which aren't really related to what it's doing. Also the implementation details seem... really complicated? I was thinking this was going to be a really short, really simple feature. 

I'm about to write a more detailed response and maybe even make another PR. I really appreciate the work you're doing, and I'd just like to keep this as simple as possible.

pekkaklarck:
Would a good first step be adding support for pressing modifiers to `Press Key`? If modifiers will be supported by `Press Keys`, it would be really strange if `Press Key` wouldn't support them also. With it checking are there modifiers could be as easy as this:

``` python
def press_key(locator, key):
    if '+' in key and len(key) > 1:
        modifiers = key.split('+')
        key = modifiers.pop()
    else:
        modifiers = []
```

just-be-dev:
@pekkaklarck, I don't know that I necessarily agree with you there. 

By reading the keyword, I'm expecting exactly one key to be pressed. I know it's semantics, but these keywords should explicitly express your intent. 

I'd also like to add that we should make the separator configurable. If I want to press `Ctrl` and `+` or `Ctrl` and `-` I could call `Press Keys   separator=-   Ctrl-+` or if we use `+` as the default `Press Keys   Ctrl+-`

HelioGuilherme66:
The way it is implemented allows both cases of Ctrl- or Ctrl+. But maybe I need to test the 4 different combinations. For now, there is no need to set a default.

pekkaklarck:
I though pressing `CTRL+C` as pressing a key, but it is true that you press two keys. You do, however, press two keys also to just press `A` too. Regardless the semantics, I wouldn't see any problem with `Press Key` supporting `CTRL+C`. Having a clear keyword name is very important, but the whole keyword call being easy to understand is even more important. This looks very clear to me:

```
Press Key    CTRL+C
```

Personally I don't see much need for allowing multiple separators between modifiers. Both `CTRL+C` and `CTRL-C` are good and easy to type and I'd just pick one. Supporting both would arguable add a bit more convenience but would also make implementation a little harder, require more testing, and need more documentation with examples. Making the separator configurable is, in my opinion, a total overkill.

Notice that all these are just my opinions and I don't feel too strongly about either of these issues. Opinions from people actually writing code and merging it have much more weight.

just-be-dev:
I can see where you're coming from. From an implementation perspective it would be convenient if press key did combinations because then press keys would just loop through each arg and call press key. 

The configurable separator was for the case when you want to use the separator in a combination. It would be easier than trying to detect ++.

pekkaklarck:
Yes, `Press Key` acting as a helper for `Press Keys` would be handy. Consistency would also be important.

I'm not sure did I got your reasoning about configurable sepatator. `++` would be invalid and `CTRL++` (or `CTRL--`) would be a special case to take into account.

HelioGuilherme66:
This pull request is going to be locked in favor of #490 which only has the relevant changes.
We will most probably loose the log history, but it's better this way.

PR #482, opened 2015-08-09
New keyword Get Browser Capabilities.

HelioGuilherme66:
Returns current capabilities in a dictionary with keys accessible as attributes.

### DIFF
--- a/src/Selenium2Library/keywords/_browsermanagement.py
+++ b/src/Selenium2Library/keywords/_browsermanagement.py
@@ -4,6 +4,7 @@ from robot.errors import DataError
 from selenium import webdriver
 from Selenium2Library import webdrivermonkeypatches
 from Selenium2Library.utils import BrowserCache
+from Selenium2Library.utils import AttrDict
 from Selenium2Library.locators import WindowManager
 from keywordgroup import KeywordGroup
 from selenium.common.exceptions import NoSuchWindowException
@@ -157,7 +158,7 @@ class _BrowserManagementKeywords(KeywordGroup):
         | Create Webdriver            | PhantomJS    | service_args=${service args}              |                         |
 
         Example for Robot Framework < 2.8:
-        | # debug IE driver |                   |                  |                                | 
+        | # debug IE driver |                   |                  |                                |
         | ${kwargs}=        | Create Dictionary | log_level=DEBUG  | log_file=%{HOMEPATH}${/}ie.log |
         | Create Webdriver  | Ie                | kwargs=${kwargs} |                                |
         """
@@ -491,7 +492,6 @@ class _BrowserManagementKeywords(KeywordGroup):
             browser.implicitly_wait(self._implicit_wait_in_secs)
         return old_wait
 
-
     def set_browser_implicit_wait(self, seconds):
         """Sets current browser's implicit wait in seconds.
 
@@ -506,6 +506,35 @@ class _BrowserManagementKeywords(KeywordGroup):
         """
         implicit_wait_in_secs = robot.utils.timestr_to_secs(seconds)
         self._current_browser().implicitly_wait(implicit_wait_in_secs)
+
+    def get_browser_capabilities(self):
+        """Gets current browser capabilities from webdriver.
+
+        Returns a Dictionary with webdriver (browser) desired capabilities attributes.
+        You may also query keys providing a default value when those keys are nonexistent.
+
+        Examples of returned attributes are:
+        'rotatable','takesScreenshot', 'acceptSslCerts', 'cssSelectorsEnabled', 'javascriptEnabled',
+        'databaseEnabled', 'locationContextEnabled', 'platform', 'browserName', 'version',
+        'nativeEvents', 'applicationCacheEnabled', 'webStorageEnabled', 'handlesAlerts'
+
+        Examples:
+        | ${capabilities}= | Get Browser Capabilities                            |                           |
+        | Log              | ${capabilities.browserName} ${capabilities.version} | # Using attributes        |
+        | Log              | Capabilities = ${capabilities}                      |                           |
+        | Log              | ${capabilities['platform']}                         | # Using dictionary        |
+        | Log              | ${capabilities.get('may_not_exist', 'default')}     | # Gets 'default'          |
+        | Log              | ${capabilities['may_not_exist']}                    | # Produces KeyError       |
+        | Log              | ${capabilities.may_not_exist}                       | # Produces AttributeError |
+
+        References, see [http://selenium-python.readthedocs.org/en/latest/api.html?highlight=desired_capabilities#selenium.webdriver.remote.webdriver.WebDriver.desired_capabilities|Desired Capabilities] and [https://selenium.googlecode.com/svn/trunk/docs/api/py/webdriver/selenium.webdriver.common.desired_capabilities.html#module-selenium.webdriver.common.desired_capabilities|Common Desired Capabilities].
+
+        New in Selenium2Library 1.8.0.
+        """
+
+        capabilities = self._current_browser().capabilities
+
+        return AttrDict(capabilities)
 
     # Private
 
@@ -532,9 +561,7 @@ class _BrowserManagementKeywords(KeywordGroup):
 
         return browser
 
-
     def _make_ff(self , remote , desired_capabilites , profile_dir):
-
         if not profile_dir: profile_dir = FIREFOX_PROFILE_DIR
         profile = webdriver.FirefoxProfile(profile_dir)
         if remote:
@@ -590,7 +617,7 @@ class _BrowserManagementKeywords(KeywordGroup):
         return browser
 
     def _create_remote_web_driver(self , capabilities_type , remote_url , desired_capabilities=None , profile=None):
-        '''parses the string based desired_capabilities if neccessary and
+        '''parses the string based desired_capabilities if necessary and
         creates the associated remote web driver'''
 
         desired_capabilities_object = capabilities_type.copy()
@@ -617,3 +644,4 @@ class _BrowserManagementKeywords(KeywordGroup):
             desired_capabilities[key.strip()] = value.strip()
 
         return desired_capabilities
+
--- a/src/Selenium2Library/keywords/_browsermanagement.py
+++ b/src/Selenium2Library/keywords/_browsermanagement.py
@@ -531,9 +531,7 @@ class _BrowserManagementKeywords(KeywordGroup):
 
         New in Selenium2Library 1.8.0.
         """
-
         capabilities = self._current_browser().capabilities
-
         return AttrDict(capabilities)
 
     # Private
--- a/src/Selenium2Library/utils/attrdict.py
+++ b/src/Selenium2Library/utils/attrdict.py
@@ -1,0 +1,25 @@
+class AttrDict(dict):
+
+    def __getattr__(self, name):
+        try:
+            return self[name]
+        except KeyError:
+            raise AttributeError(name)
+
+    def __setattr__(self, name, value):
+        self[name] = value
+
+    def __delattr__(self, name):
+        try:
+            del self[name]
+        except KeyError:
+            raise AttributeError(name)
+
+    def __eq__(self, other):
+        return dict.__eq__(self, other)
+
+    def __str__(self):
+        return '{%s}' % ', '.join('%r: %r' % item for item in self.iteritems())
+
+    # Must use original dict.__repr__ to allow customising PrettyPrinter.
+    __repr__ = dict.__repr__
--- a/src/Selenium2Library/utils/attrdict.py
+++ b/src/Selenium2Library/utils/attrdict.py
@@ -14,12 +14,3 @@ class AttrDict(dict):
             del self[name]
         except KeyError:
             raise AttributeError(name)
-
-    def __eq__(self, other):
-        return dict.__eq__(self, other)
-
-    def __str__(self):
-        return '{%s}' % ', '.join('%r: %r' % item for item in self.iteritems())
-
-    # Must use original dict.__repr__ to allow customising PrettyPrinter.
-    __repr__ = dict.__repr__
--- a/test/acceptance/create_webdriver.robot
+++ b/test/acceptance/create_webdriver.robot
@@ -29,6 +29,40 @@ Create Webdriver With Bad Keyword Argument Dictionary
     Run Keyword And Expect Error    kwargs must be a dictionary.
     ...    Create Webdriver    Firefox    kwargs={'spam': 'eggs'}
 
+Get Browser Capabilities Using Dictionary
+    [Documentation]    Gets current browser capabilities from WebDriver
+    [Setup]    Set Driver Variables
+    Create Webdriver    ${DRIVER_NAME}
+    Go To    ${FRONT PAGE}    # If we don't do this then no version details
+    ${capabilities}=    Get Browser Capabilities
+    Log    Capabilities = ${capabilities}
+    Log    Browser is ${capabilities['browserName']}, version ${capabilities['version']}
+    ${browser}=    Normalize Browser Name
+    Should Match    ${capabilities['browserName']}    ${browser}
+    Run Keyword If    "${browser}" != "internet explorer"    Log    Rotatable = ${capabilities['rotatable']}
+    # Disabled next step to PASS in RF pre 2.9
+    # Run Keyword If    "${browser}" == "internet explorer"    Run Keyword And Expect Error    *    Log
+    #     ...    Rotatable = ${capabilities['rotatable']}
+    Log    Platform = ${capabilities['platform']}
+    [Teardown]    Close Browser
+
+Get Browser Capabilities Using Attributes
+    [Documentation]    Gets current browser capabilities from WebDriver
+    [Setup]    Run Keywords    Cannot Run In RF Pre 2.9    Set Driver Variables
+    Create Webdriver    ${DRIVER_NAME}
+    Go To    ${FRONT PAGE}    # If we don't do this then no version details
+    &{capabilities}=    Get Browser Capabilities
+    Log    Capabilities = &{capabilities}
+    Log    Browser is ${capabilities.browserName}, version ${capabilities.version}
+    ${browser}=    Normalize Browser Name
+    Should Match    ${capabilities.browserName}    ${browser}
+    Run Keyword If    "${browser}" != "internet explorer"    Log    Rotatable = ${capabilities.rotatable}
+    Run Keyword If    "${browser}" == "internet explorer"    Run Keyword And Expect Error    *    Log
+        ...    Rotatable = ${capabilities.rotatable}
+    Log    Platform = ${capabilities.platform}
+    Log    Wearable = ${capabilities.get('wearable', 'What were you expecting? To wear a browser?')}
+    [Teardown]    Close Browser
+
 *** Keywords ***
 Set Driver Variables
     [Documentation]    Selects proper driver
@@ -50,3 +84,16 @@ Set Driver Variables
     Run Keyword If    "${name}"=="Remote"    Set To Dictionary    ${kwargs}    command_executor
     ...    ${url as str}    desired_capabilities    ${caps}
     Set Test Variable    ${KWARGS}    ${kwargs}
+
+Cannot Run In RF Pre 2.9
+    [Documentation]    Skips the test if running in Robot Framework previous to 2.9.
+    Import Library    robot.version    WITH NAME    Version
+    ${VERSION}=    Version.Get Version
+    Run Keyword If    '${VERSION}' < '2.9.0'    Fail And Set Non-Critical
+    ...    This test does not work with Robot Framework ${VERSION}. Please use version 2.9 or newer.
+
+Normalize Browser Name
+    [Documentation]    Returns a normalized name for ${DRIVER_NAME} especially for IE
+    ${BROWSER}=    Set Variable    ${DRIVER_NAME.lower()}
+    ${ret}=    Set Variable If    "${BROWSER}" == "ie"    internet explorer    ${BROWSER}
+    [Return]    ${ret}
--- a/test/acceptance/create_webdriver.robot
+++ b/test/acceptance/create_webdriver.robot
@@ -33,16 +33,19 @@ Get Browser Capabilities Using Dictionary
     [Documentation]    Gets current browser capabilities from WebDriver
     [Setup]    Set Driver Variables
     Create Webdriver    ${DRIVER_NAME}
-    Go To    ${FRONT PAGE}    # If we don't do this then no version details
     ${capabilities}=    Get Browser Capabilities
     Log    Capabilities = ${capabilities}
-    Log    Browser is ${capabilities['browserName']}, version ${capabilities['version']}
+    Log    Common capabilities: 'takesScreenshot'=${capabilities['takesScreenshot']}, 'javascriptEnabled'=${capabilities['javascriptEnabled']},
+    Log    'cssSelectorsEnabled'=${capabilities['cssSelectorsEnabled']}, 'browserName'=${capabilities['browserName']},
+    Log    'platform'=${capabilities['platform']}, 'version'=${capabilities['version']}, 'nativeEvents'=${capabilities['nativeEvents']},
+    Log    'handlesAlerts'=${capabilities['handlesAlerts']}
     ${browser}=    Normalize Browser Name
     Should Match    ${capabilities['browserName']}    ${browser}
-    Run Keyword If    "${browser}" != "internet explorer"    Log    Rotatable = ${capabilities['rotatable']}
+    Run Keyword If    "${browser}" != "internet explorer"    Log
+    ...    Rotatable = ${capabilities['rotatable']}
     # Disabled next step to PASS in RF pre 2.9
     # Run Keyword If    "${browser}" == "internet explorer"    Run Keyword And Expect Error    *    Log
-    #     ...    Rotatable = ${capabilities['rotatable']}
+    #    ...    Rotatable = ${capabilities['rotatable']}
     Log    Platform = ${capabilities['platform']}
     [Teardown]    Close Browser
 
@@ -50,15 +53,18 @@ Get Browser Capabilities Using Attributes
     [Documentation]    Gets current browser capabilities from WebDriver
     [Setup]    Run Keywords    Cannot Run In RF Pre 2.9    Set Driver Variables
     Create Webdriver    ${DRIVER_NAME}
-    Go To    ${FRONT PAGE}    # If we don't do this then no version details
     &{capabilities}=    Get Browser Capabilities
     Log    Capabilities = &{capabilities}
-    Log    Browser is ${capabilities.browserName}, version ${capabilities.version}
+    Log    Common capabilities: 'takesScreenshot'=${capabilities.takesScreenshot}, 'javascriptEnabled'=${capabilities.javascriptEnabled},
+    Log    'cssSelectorsEnabled'=${capabilities.cssSelectorsEnabled}, 'browserName'=${capabilities.browserName},
+    Log    'platform'=${capabilities.platform}, 'version'=${capabilities.version}, 'nativeEvents'=${capabilities.nativeEvents},
+    Log    'handlesAlerts'=${capabilities.handlesAlerts}
     ${browser}=    Normalize Browser Name
     Should Match    ${capabilities.browserName}    ${browser}
-    Run Keyword If    "${browser}" != "internet explorer"    Log    Rotatable = ${capabilities.rotatable}
-    Run Keyword If    "${browser}" == "internet explorer"    Run Keyword And Expect Error    *    Log
-        ...    Rotatable = ${capabilities.rotatable}
+    Run Keyword If    "${browser}" != "internet explorer"    Log
+    ...    Rotatable = ${capabilities.rotatable}
+    Run Keyword If    "${browser}" == "internet explorer"    Run Keyword And Expect Error    *
+    ...    Log    Rotatable = ${capabilities.rotatable}
     Log    Platform = ${capabilities.platform}
     Log    Wearable = ${capabilities.get('wearable', 'What were you expecting? To wear a browser?')}
     [Teardown]    Close Browser
--- a/test/unit/utils/test_attrdict.py
+++ b/test/unit/utils/test_attrdict.py
@@ -2,42 +2,36 @@ import unittest
 
 from robot.utils.asserts import assert_equal, assert_not_equal, assert_raises
 from Selenium2Library.utils import AttrDict
-try:
-    from robot.utils import IRONPYTHON
-except ImportError:  # New in RF 2.9
-    import sys
-    IRONPYTHON = sys.platform == 'cli'
-    pass
 
 
 class TestAttrDict(unittest.TestCase):
 
     def setUp(self):
-        self.dd = AttrDict([('z', 1), (2, 'y'), ('x', 3)])
+        self.ad = AttrDict([('z', 1), (2, 'y'), ('x', 3)])
 
     def test_get(self):
-        assert_equal(self.dd[2], 'y')
-        assert_equal(self.dd.x, 3)
-        assert_raises(KeyError, self.dd.__getitem__, 'nonex')
-        assert_raises(AttributeError, self.dd.__getattr__, 'nonex')
+        assert_equal(self.ad[2], 'y')
+        assert_equal(self.ad.x, 3)
+        assert_raises(KeyError, self.ad.__getitem__, 'nonex')
+        assert_raises(AttributeError, self.ad.__getattr__, 'nonex')
 
     def test_equality_with_normal_dict(self):
-        assert_equal(self.dd, {'z': 1, 2: 'y', 'x': 3})
+        assert_equal(self.ad, {'z': 1, 2: 'y', 'x': 3})
 
     def test_set(self):
-        self.dd.x = 42
-        self.dd.new = 43
-        self.dd[2] = 44
-        self.dd['n2'] = 45
-        assert_equal(self.dd, {'z': 1, 2: 44, 'x': 42, 'new': 43, 'n2': 45})
+        self.ad.x = 42
+        self.ad.new = 43
+        self.ad[2] = 44
+        self.ad['n2'] = 45
+        assert_equal(self.ad, {'z': 1, 2: 44, 'x': 42, 'new': 43, 'n2': 45})
 
     def test_del(self):
-        del self.dd.x
-        del self.dd[2]
-        self.dd.pop('z')
-        assert_equal(self.dd, {})
-        assert_raises(KeyError, self.dd.__delitem__, 'nonex')
-        assert_raises(AttributeError, self.dd.__delattr__, 'nonex')
+        del self.ad.x
+        del self.ad[2]
+        self.ad.pop('z')
+        assert_equal(self.ad, {})
+        assert_raises(KeyError, self.ad.__delitem__, 'nonex')
+        assert_raises(AttributeError, self.ad.__delattr__, 'nonex')
 
     def test_same_str_and_repr_format_as_with_normal_dict(self):
         D = {'foo': 'bar', '"\'': '"\'', '\n': '\r', 1: 2, (): {}, True: False}
--- a/test/unit/utils/test_attrdict.py
+++ b/test/unit/utils/test_attrdict.py
@@ -1,0 +1,50 @@
+import unittest
+
+from robot.utils.asserts import assert_equal, assert_not_equal, assert_raises
+from Selenium2Library.utils import AttrDict
+try:
+    from robot.utils import IRONPYTHON
+except ImportError:  # New in RF 2.9
+    import sys
+    IRONPYTHON = sys.platform == 'cli'
+    pass
+
+
+class TestAttrDict(unittest.TestCase):
+
+    def setUp(self):
+        self.dd = AttrDict([('z', 1), (2, 'y'), ('x', 3)])
+
+    def test_get(self):
+        assert_equal(self.dd[2], 'y')
+        assert_equal(self.dd.x, 3)
+        assert_raises(KeyError, self.dd.__getitem__, 'nonex')
+        assert_raises(AttributeError, self.dd.__getattr__, 'nonex')
+
+    def test_equality_with_normal_dict(self):
+        assert_equal(self.dd, {'z': 1, 2: 'y', 'x': 3})
+
+    def test_set(self):
+        self.dd.x = 42
+        self.dd.new = 43
+        self.dd[2] = 44
+        self.dd['n2'] = 45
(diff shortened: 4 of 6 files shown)


aaltat:
Is there a reason why this is a separate keyword and why it could not be returned by the Open Browser and Create Webdriver keywords?

HelioGuilherme66:
My reasoning is:
- Do not touch the existing keywords if there is no really need of it
- Separating the creation from the informative functions is better because
  getting info is secondary
- The new keyword could be improved in the way it returns info (dictionary
  query, for example).

pekkaklarck:
What about returning a custom object with available information in attributes? It would make it easy to access info like `${info.version}` or `${info.platform}`.

aaltat:
I like @pekkaklarck idea. Some more work in code side but I think it gives better result.

If the desired capabilities can not be changed after the creation of the WebDriver, the most logical place, in my mind, is to return it after the creation of the WebDriver object. But I have not faced a situation, where I would have a need to change the WebDriver after the creation, so I am not sure is it possible.

But if we change the existing keywords to return the capabilities, it creates a backwards incompatible change (because Open Browser and Create Webdriver do already return the browser index). What do others think, @zephraph, @emanlove?

just-be-dev:
I'm fine with it being its own keyword. It makes it more explicit and doesn't break any backwards compatibility.

I also agree with @pekkaklarck that it should be an object, not a JSON string. The conversion should be simple.

HelioGuilherme66:
Trying to get object attributes instead of dictionaries.

HelioGuilherme66:
Ready to be reviewed.

pekkaklarck:
I'm sorry to say, but the code in `CapabilitiesConfig` is so strange in so many ways that I don't know where to start commenting it. Some of the weird things I noticed:
1. `__init__` accepts `*args` but only processes first of the given args.
2. None of the code below `__init__` doesn't seem to be ever used.
3. `property` is used wrong.

Could you try to find some intermediate Python book or other resource to study these topics a bit more? It would be quite time taking for me or others to explain the above issues in detail.

There's also a functional problem that nothing guarantees that the object always has all capabilities defined. That's related to problems 2. and 3. above.

HelioGuilherme66:
Please do a review again. If the code is not acceptable, please consider it as a prototype. I think it is a useful keyword.

src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -617,3 +644,39 @@ def _parse_capabilities_string(self, capabilities_string):
...
         return desired_capabilities
+
+
+class CapabilitiesConfig(object):
+
+    def __init__(self, arg):
+        for k in arg:
+            self.__setattr__(k, arg[k])
+
+"""
```
aaltat: If the code is not needed, deleted it.

test/acceptance/create_webdriver.robot
```diff
@@ -15,38 +15,54 @@ Create Webdriver Creates Functioning WebDriver
 
 Create Webdriver With Bad Driver Name
     [Documentation]    Invalid browser name
-    Run Keyword And Expect Error    'Fireox' is not a valid WebDriver name
-    ...    Create Webdriver    Fireox
+    Run Keyword And Expect Error    'Fireox' is not a valid WebDriver name    Create Webdriver
+    ...    Fireox
```
aaltat: Why this was changed?

test/acceptance/create_webdriver.robot
```diff
@@ -15,38 +15,54 @@ Create Webdriver Creates Functioning WebDriver
...
 
 Create Webdriver With Duplicate Arguments
     [Documentation]    Invalid values in arguments
     ${kwargs}=    Create Dictionary    arg=1
-    Run Keyword And Expect Error    Got multiple values for argument 'arg'.
-    ...    Create Webdriver    Firefox    kwargs=${kwargs}    arg=2
+    Run Keyword And Expect Error    Got multiple values for argument 'arg'.    Create Webdriver
```
aaltat: Why this was changed?

test/acceptance/create_webdriver.robot
```diff
@@ -15,38 +15,54 @@ Create Webdriver Creates Functioning WebDriver
...
+    ...    Firefox    kwargs=${kwargs}    arg=2
 
 Create Webdriver With Bad Keyword Argument Dictionary
     [Documentation]    Invalid arguments types
-    Run Keyword And Expect Error    kwargs must be a dictionary.
-    ...    Create Webdriver    Firefox    kwargs={'spam': 'eggs'}
+    Run Keyword And Expect Error    kwargs must be a dictionary.    Create Webdriver    Firefox
+    ...    kwargs={'spam': 'eggs'}
```
aaltat: Why this was changed?

test/acceptance/create_webdriver.robot
```diff
@@ -15,38 +15,54 @@ Create Webdriver Creates Functioning WebDriver
...
+
+Get Browser Info
```
aaltat: Could you write cleaner tests and writing more tests would be useful. At least: 
1) Clean the not required logging
2) Write tests that verifies something
3) Also it would be good idea to test when Open Browser keyword is used.
4) I think we do not use selenium grid, but at least give it some thought.

The 3&4 are pretty much extra because it is always the WebDriver driver object giving the return value.

test/acceptance/create_webdriver.robot
```diff
@@ -15,38 +15,54 @@ Create Webdriver Creates Functioning WebDriver
...
-    ${caps}=    Evaluate    sys.modules['selenium.webdriver'].DesiredCapabilities.${dc name}
-    ...    selenium.webdriver,sys
+    ${caps}=    Evaluate    sys.modules['selenium.webdriver'].DesiredCapabilities.${dc name}    selenium.webdriver,sys
     ${url as str}=    Evaluate    str('${REMOTE_URL}')    # cannot be unicode for versions >= 2.32
     ${kwargs}=    Create Dictionary
-    Run Keyword If    "${name}"=="Remote"    Set To Dictionary    ${kwargs}    command_executor
-    ...    ${url as str}    desired_capabilities    ${caps}
-    Set Test Variable    ${KWARGS}    ${kwargs}
+    Run Keyword If    "${name}"=="Remote"    Set To Dictionary    ${kwargs}
+    ...    command_executor=${url as str}    desired_capabilities=${caps}
```
aaltat: Why this was changed?

test/acceptance/create_webdriver.robot
```diff
@@ -15,38 +15,54 @@ Create Webdriver Creates Functioning WebDriver
...
+    Set Test Variable    ${KWARGS}    ${kwargs}
```
aaltat: Add line feed at the end.

test/acceptance/create_webdriver.robot
```diff
@@ -15,38 +15,54 @@ Create Webdriver Creates Functioning WebDriver
...
-    ...    chrome=CHROME    opera=OPERA    phantomjs=PHANTOMJS    htmlunit=HTMLUNIT
-    ...    htmlunitwithjs=HTMLUNITWITHJS    android=ANDROID    iphone=IPHONE
-    ...    safari=SAFARI
+    ...    internetexplorer=INTERNETEXPLORER    googlechrome=CHROME    gc=CHROME    chrome=CHROME
+    ...    opera=OPERA    phantomjs=PHANTOMJS    htmlunit=HTMLUNIT    htmlunitwithjs=HTMLUNITWITHJS
+    ...    android=ANDROID    iphone=IPHONE    safari=SAFARI
     ${dc name}=    Get From Dictionary    ${dc names}    ${BROWSER.lower().replace(' ', '')}
-    ${caps}=    Evaluate    sys.modules['selenium.webdriver'].DesiredCapabilities.${dc name}
-    ...    selenium.webdriver,sys
+    ${caps}=    Evaluate    sys.modules['selenium.webdriver'].DesiredCapabilities.${dc name}    selenium.webdriver,sys
```
aaltat: Why this was changed?

test/acceptance/create_webdriver.robot
```diff
@@ -15,38 +15,54 @@ Create Webdriver Creates Functioning WebDriver
...
     ${name}=    Evaluate    "Remote" if "${REMOTE_URL}"!="None" else ${drivers}["${BROWSER.lower().replace(' ', '')}"]
     Set Test Variable    ${DRIVER_NAME}    ${name}
     ${dc names}=    Create Dictionary    ff=FIREFOX    firefox=FIREFOX    ie=INTERNETEXPLORER
-    ...    internetexplorer=INTERNETEXPLORER    googlechrome=CHROME    gc=CHROME
-    ...    chrome=CHROME    opera=OPERA    phantomjs=PHANTOMJS    htmlunit=HTMLUNIT
-    ...    htmlunitwithjs=HTMLUNITWITHJS    android=ANDROID    iphone=IPHONE
-    ...    safari=SAFARI
+    ...    internetexplorer=INTERNETEXPLORER    googlechrome=CHROME    gc=CHROME    chrome=CHROME
+    ...    opera=OPERA    phantomjs=PHANTOMJS    htmlunit=HTMLUNIT    htmlunitwithjs=HTMLUNITWITHJS
+    ...    android=ANDROID    iphone=IPHONE    safari=SAFARI
```
aaltat: Why this was changed?

test/acceptance/create_webdriver.robot
```diff
@@ -15,38 +15,54 @@ Create Webdriver Creates Functioning WebDriver
...
+    [Teardown]    Close Browser
 
 *** Keywords ***
 Set Driver Variables
     [Documentation]    Selects proper driver
     ${drivers}=    Create Dictionary    ff=Firefox    firefox=Firefox    ie=Ie
-    ...    internetexplorer=Ie    googlechrome=Chrome    gc=Chrome
-    ...    chrome=Chrome    opera=Opera    phantomjs=PhantomJS safari=Safari
+    ...    internetexplorer=Ie    googlechrome=Chrome    gc=Chrome    chrome=Chrome    opera=Opera
+    ...    phantomjs=PhantomJS    safari=Safari
```
aaltat: Why this was changed?

src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -507,6 +507,34 @@ def set_browser_implicit_wait(self, seconds):
...
+        | Log              | "${info.browserName} ${info.version}" |             |
+        | ${capabilities}= | Get Browser Info                      | type='dict' |
+        | Log              | "Capabilities = ${capabilities}"      |             |
+        """
+
+        capabilities = self._current_browser().desired_capabilities
+
+        if type == 'object':
+            kappa = CapabilitiesConfig(capabilities)
+            return kappa
```
aaltat: You could write 531 - 532 as one liner: return CapabilitiesConfig(capabilities)

src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -507,6 +507,34 @@ def set_browser_implicit_wait(self, seconds):
...
+        else:
+            dkappa = dict(capabilities)
+            return dkappa
```
aaltat: You could write 534 - 535 as one liner: return dict(capabilities)

src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -507,6 +507,34 @@ def set_browser_implicit_wait(self, seconds):
...
+        | ${capabilities}= | Get Browser Info                      | type='dict' |
+        | Log              | "Capabilities = ${capabilities}"      |             |
+        """
+
+        capabilities = self._current_browser().desired_capabilities
+
+        if type == 'object':
+            kappa = CapabilitiesConfig(capabilities)
+            return kappa
+        else:
```
aaltat: Is there a reason why we would need to return dictionary?

src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -507,6 +507,34 @@ def set_browser_implicit_wait(self, seconds):
         implicit_wait_in_secs = robot.utils.timestr_to_secs(seconds)
         self._current_browser().implicitly_wait(implicit_wait_in_secs)
 
+    def get_browser_info(self, type='object'):
```
aaltat: Should the keyword name be changed. Info is quite vague word and does not describe what the keyword does. 

Perhaps get_browsers_desired_capabilities is better name?

src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -507,6 +507,34 @@ def set_browser_implicit_wait(self, seconds):
...
+        """Gets information from webdriver on current browser.
```
aaltat: Info is quite vague name and does not explain what the keyword does.

src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -507,6 +507,34 @@ def set_browser_implicit_wait(self, seconds):
...
+
+        By default returns an Object, otherwise if the argument `type`is different of 'object', returns
+        a Dictionary.
+
```
aaltat, line 515: Perhaps it would be good to refer the Selenium API documentation too: http://selenium-python.readthedocs.org/en/latest/api.html?highlight=desired_capabilities#selenium.webdriver.remote.webdriver.WebDriver.desired_capabilities

src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -507,6 +507,34 @@ def set_browser_implicit_wait(self, seconds):
...
+        Examples of returned attributes are:
+        'rotatable','takesScreenshot', 'acceptSslCerts', 'cssSelectorsEnabled', 'javascriptEnabled',
+        'databaseEnabled', 'locationContextEnabled', 'platform', 'browserName', 'version',
+        'nativeEvents', 'applicationCacheEnabled', 'webStorageEnabled', 'handlesAlerts'
+
+        Examples:
```
aaltat, line 521: Could the example be more concrete. Explain that when this type of browser is opened then x contains y.

src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -617,3 +644,39 @@ def _parse_capabilities_string(self, capabilities_string):
...
+
+    @capabilities.getter
+    def capabilities(self, item):
+        try:
+            result = self.__getattribute__(item)
+            return result
+        except AttributeError:
+            self._debug("Unknown capability %s for this WebDriver session." % (item))
+            return None
+"""
```
aaltat: Remember to add line feed.

src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -617,3 +644,39 @@ def _parse_capabilities_string(self, capabilities_string):
             desired_capabilities[key.strip()] = value.strip()
 
         return desired_capabilities
+
+
+class CapabilitiesConfig(object):
+
+    def __init__(self, arg):
+        for k in arg:
+            self.__setattr__(k, arg[k])
```
aaltat: I am not familiar with **setattr** and can not therefore judge is this good/bad/correct?

HelioGuilherme66:
New changes after first review.

src/Selenium2Library/keywords/_browsermanagement.py
pekkaklarck: I agree `get_browser_info` is pretty vague. If I got it right, _desired capabilities_ is a commonly understood term in the Selenium community, and this keyword could just be named `get_desired_capabilities`. I don't thing `browser` is needed there but wouldn't mind it either.

src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -507,6 +507,36 @@ def set_browser_implicit_wait(self, seconds):
...
+        Examples of returned attributes are:
+        'rotatable','takesScreenshot', 'acceptSslCerts', 'cssSelectorsEnabled', 'javascriptEnabled',
+        'databaseEnabled', 'locationContextEnabled', 'platform', 'browserName', 'version',
+        'nativeEvents', 'applicationCacheEnabled', 'webStorageEnabled', 'handlesAlerts'
+
+        Examples:
+        | ${info}=         | Get Browser Info                      |             |
+        | Log              | "${info.browserName} ${info.version}" |             |
+        | &{capabilities}= | Get Browser Info                      | type='dict' |
+        | Log              | "Capabilities = &{capabilities}"      |             |
```
pekkaklarck: Why all these quotes in the example??

src/Selenium2Library/keywords/_browsermanagement.py
pekkaklarck: There's nothing bad `__setattr__` itself but I wouldn't use it here. A cleaner implementation would be:

``` python
def __init__(self, capabilities):
    self.__dict__.update(capabilities)
```

test/acceptance/create_webdriver.robot
pekkaklarck: Cleanup to code is fine bu unrelated changes should be submitted as separate PRs. In this case the code doesn't actually get any better, on the contrary, so this change should be reverted.

This comment applies also to all other places where @aaltat has commented `Why this was changed?`.

src/Selenium2Library/keywords/_browsermanagement.py
HelioGuilherme66: Well, my point of view as a Tester I just usually want to get the Browser name and version. I thought it was more natural to get browser information. I will rename it to `get_browser_capabilities` because they are not necessarily the "desired" capabilities.

test/acceptance/create_webdriver.robot
pekkaklarck: I agree with @aaltat that the test below isn't testing much. I'm not sure how much could actually be verified, because knowing exactly what information is returned may be hard. Perhaps the test should just access all information we think should always be there.

Regardless what is done, the last lines that ignore all errors should just be removed. They just add noise.

test/acceptance/create_webdriver.robot
HelioGuilherme66: It was done in another PR, already merged, maybe the editor changed it.

src/Selenium2Library/keywords/_browsermanagement.py
HelioGuilherme66: I don't understand. This was the only solution I found to return an object. What is the alternative?

pekkaklarck:
I added some line notes, but I also have a bigger question about the usage. What is the point to allow returning either an object or a dictionary? An object with all information in easily accessible attributes is much more useful when just querying info. What use cases are there for returning a dict?

If returning a dict is a useful feature, then it would probably be a good idea to always return a dictionary but use custom dict that allows accessing keys as attributes like `${dict.key}`. RF 2.9 added [DotDict](https://github.com/robotframework/robotframework/blob/master/src/robot/utils/dotdict.py) that behaves this way. S2L could use it directly, but it would make it dependent on RF 2.9. I don't think that is generally acceptable, but perhaps only this kw could require it. Alternatively S2L could include its own version. It probably wouldn't need to be ordered and thus wouldn't need Py 2.6 compatible `OrderedDict`.

test/acceptance/create_webdriver.robot
pekkaklarck: What was done? If this change has already been merged, it shouldn't anymore show in this PR. If this test has been changed after creating this PR, then this PR should be updated.

src/Selenium2Library/keywords/_browsermanagement.py
pekkaklarck: I already proposed an alternative to use `self.__dict__.update(capabilities)`. As I already commented the PR itself, it might be better to return a custom dict instead.

src/Selenium2Library/keywords/_browsermanagement.py
pekkaklarck: Could you clarify why not using the term commonly used by Selenium is not a good idea here?

test/acceptance/create_webdriver.robot
HelioGuilherme66: There is a bug in the spacing of `phantomjs=PhantomJS safari=Safari`.

pekkaklarck: Looks like a valid fix. Should anyway be submitted separately.

src/Selenium2Library/keywords/_browsermanagement.py
HelioGuilherme66: Most of the Testers I know, only use Selenium2Library, they don't need to get into Selenium (or know about it). As I said before, for example, we don't control the browser version, and we are getting the current information or capabilities.

src/Selenium2Library/keywords/_browsermanagement.py
HelioGuilherme66: But the keyword already returns a dictionary or an object as proposed.

HelioGuilherme66:
Returning a dictionary is useful because we can easily see the keys. For example PhantomJS returns `driverVersion` which I haven't see in others.

src/Selenium2Library/keywords/_browsermanagement.py
pekkaklarck: S2L is a [leaky abstraction](http://www.joelonsoftware.com/articles/LeakyAbstractions.html) on top of Selenium and many/most people working with the former for a longer time need to understand something about the latter too. Using same terms makes understanding concepts a lot easier.

The above said, it seems Selenium also uses just term capabilities when talking about the current capabilities. In that regard `get_browser_capabilities` is a fine name. It would probably then be a good idea to also access `self._current_browser().capabilities` in the code below and not its alias `self._current_browser().desired_capabilities`.

src/Selenium2Library/keywords/_browsermanagement.py
pekkaklarck: It seems you don't understand what `self.__dict__.update(capabilities)` actually does.

Also, I don't think it is a good idea that the keyword returns either an object or a dictionary. If returning a dictionary is needed, it should _always_ return a dict that allows accessing keys as attributes. This is something I commented in the general PR comments in a bit more detail.

HelioGuilherme66:
Request for new review (hopefully the last one ;) ). Thanks Tatu and Pekka for your teaching!

src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -1,6 +1,12 @@
 import os
 import robot
 from robot.errors import DataError
+try:
+    from robot.utils import DotDict
+    NO_DOT_DICT= False
+except ImportError:  # New in RF 2.9
+    NO_DOT_DICT= True
+    pass
```
pekkaklarck: This pattern is typically better:

``` python
try:
    from robot.utils import DotDict
except ImportError:  # New inf RF 2.9
    DotDict = None

# later:
if DotDict is None:
    # ...
```

If there is a replacement for the missing import, like `dict` could be used here, this variant works even better:

``` python
try:
    from robot.utils import DotDict
except ImportError:  # New inf RF 2.9
    DotDict = dict
```

In this particular case I think S2L should rather implement its own `DotDict` variant. It doesn't need to be ordered so it can be simpler than the version in RF and should also be renamed differently to avoid confusion. I already included one possible implementation of such `AttrDict` in my earlier comment to this PR.

src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -507,6 +513,35 @@ def set_browser_implicit_wait(self, seconds):
...
+        | Log              | ${capabilities.browserName} ${capabilities.version} | # Depends on RF 2.9 |
+        | Log              | Capabilities = ${capabilities}                      |                     |
+        | Log              | ${capabilities['platform']}                         |                     |
+
+        References, see [http://selenium-python.readthedocs.org/en/latest/api.html?highlight=desired_capabilities#selenium.webdriver.remote.webdriver.WebDriver.desired_capabilities|Desired Capabilities] and [https://selenium.googlecode.com/svn/trunk/docs/api/py/webdriver/selenium.webdriver.common.desired_capabilities.html#module-selenium.webdriver.common.desired_capabilities|Common Desired Capabilities].
+
+        New in Selenium2Library 1.8.0.
+        """
+
+        capabilities = CapabilitiesConfig(self._current_browser().capabilities)
```
pekkaklarck: `CapabilitiesConfig` doesn't do anything useful anymore. It should be removed.

pekkaklarck:
One thing to consider is what to do if a user tries to access capability that doesn't exist. Currently this causes `AtttributeError` or `KeyError`, depending on how it is accessed, with not too informative message. That can happen both when there is a typo but also when trying to access a capability only certain browsers have. there  are various ways how to enhance it:
1. Just enhance the error message to explain that this capability wasn't found. Notice that the exception type should not be changed.
2. Always return something when a user access a capability that exists in some browser. This requires getting the information what capabilities various browsers have from WebDriver and thus requires a bit more work.
3. Add separate method to the returned object to get an attribute that accepts a default value: `${capabilities.get('may_not_exist', 'default')`.
4. Just explicitly document the current behavior.

After writing the above, I realized that if capabilities are returned as a (custom) dict, they already have `get()` method that works exactly like proposed in 3. Perhaps it would be enough to just document that accessing a non-existing capability causes an error and show how to use default values with `get()`? Making sure the exception message is somewhat clear (1.) would be a good idea too.

HelioGuilherme66:
This is ready to be reviewed again. I will only adapt the test, because when using IE it reports browserName as "internet explorer" (it fails because I'm comparing with ${BROWSER} == "ie").

src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -1,6 +1,11 @@
 import os
 import robot
 from robot.errors import DataError
+try:
+    from robot.utils import DotDict
+except ImportError:  # New in RF 2.9
+    DotDict = dict
```
just-be-dev: While this works, I agree with @pekkaklarck. We should just do our own implementation of DotDict, even though it'll be a bit more work. That will remove the 2.9 dependency which could be potentially confusing to users who don't read the documentation well.

src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -507,6 +512,34 @@ def set_browser_implicit_wait(self, seconds):
...
+        | Log              | ${capabilities['may_not_exist']}                    | # Produces KeyError |
+
+        References, see [http://selenium-python.readthedocs.org/en/latest/api.html?highlight=desired_capabilities#selenium.webdriver.remote.webdriver.WebDriver.desired_capabilities|Desired Capabilities] and [https://selenium.googlecode.com/svn/trunk/docs/api/py/webdriver/selenium.webdriver.common.desired_capabilities.html#module-selenium.webdriver.common.desired_capabilities|Common Desired Capabilities].
+
+        New in Selenium2Library 1.8.0.
+        """
+
+        capabilities = self._current_browser().capabilities
+
+        return DotDict(capabilities)
```
pekkaklarck: Unnecessary empty lines.

pekkaklarck:
As I already commented, and as @zephraph agreed, S2L should have its own `DotDict` implementation and not use the one from RF 2.9. It should be named differently to avoid confusion, though. This ought to be enough:

``` python
class AttrDict(dict):

    def __getattr__(self, name):
        try:
            return self[name]
        except KeyError:
            raise AttributeError(name)

    def __setattr__(self, name, value):
        self[name] = value

    def __delattr__(self, name):
        try:
            del self[name]
        except KeyError:
            raise AttributeError(name)
```

The above class should be added to S2L utils or similar place where it can be reused by other keywords if needed.

pekkaklarck:
Another thing to decide before the merger is how to handle non-existing attributes. This is something I already commented earlier.

aaltat:
Because the desired capabilities do change between browsers (version) or might change between selenium versions, I am favour to return exactly what underlying selenium returns. I am afraid that it comes a maintenance burden to take account different possibilities and fix them when world changes.

Also if one uses browser from cloud services, those services have an habit to extend the desired capabilities to their needs.

pekkaklarck:
Very good point. Using `${caps.get('capName', 'default')}` ought to be OK when capabilitis may not exist. This usage should anyway be documented with an example.

HelioGuilherme66:
Added your suggestions, and adapted to pass with RF 2.8.7. Please review, thanks.

src/Selenium2Library/utils/dotdict.py
```diff
@@ -0,0 +1,48 @@
...
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+
+try:
+    from collections import OrderedDict
+except ImportError:  # New in Python 2.7
+    from .ordereddict import OrderedDict
+
+
+class DotDict(OrderedDict):
```
just-be-dev: This doesn't really need to be an `OrderedDict` does it?

In this case I don't think the order really matters.

src/Selenium2Library/utils/ordereddict.py
```diff
@@ -0,0 +1,128 @@
...
+#     NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+#     HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+#     WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+#     FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+#     OTHER DEALINGS IN THE SOFTWARE.
+
+from UserDict import DictMixin
+
+
+class OrderedDict(dict, DictMixin):
```
just-be-dev: This _definitely_ doesn't need to be here. Is this just a reimplementation of `OrderedDict`? I suppose RF may include this for legacy support, but as far as I'm aware all the newer versions of python have this baked in.

just-be-dev:
I glanced over this quickly. I think including the `OrderedDict` library isn't necessary.

src/Selenium2Library/utils/dotdict.py
pekkaklarck: Yes, this dict doesn't need to be ordered. Also, it shouldn't be named `DotDict` because it causes confusion with Robot's `DotDict`. I already included a working `AttrDict` in my earlier comments. Was there a reason you didn't use that?

HelioGuilherme66:
Ready for review. Sorry for having changed run_unit_tests.py (is needed in Fedora 22).

just-be-dev:
Looks good to me. I didn't see any glaring deficiencies. 

@pekkaklarck; @aaltat, will one of you guys look this over one more time before we merge?

src/Selenium2Library/utils/attrdict.py
```diff
@@ -0,0 +1,25 @@
...
+            raise AttributeError(name)
+
+    def __eq__(self, other):
+        return dict.__eq__(self, other)
+
+    def __str__(self):
+        return '{%s}' % ', '.join('%r: %r' % item for item in self.iteritems())
+
+    # Must use original dict.__repr__ to allow customising PrettyPrinter.
+    __repr__ = dict.__repr__
```
pekkaklarck: There is no reason to override `__eq__`, `__str__`, or `__repr__` here.

test/unit/utils/test_attrdict.py
```diff
@@ -0,0 +1,50 @@
+import unittest
+
+from robot.utils.asserts import assert_equal, assert_not_equal, assert_raises
+from Selenium2Library.utils import AttrDict
+try:
+    from robot.utils import IRONPYTHON
+except ImportError:  # New in RF 2.9
+    import sys
+    IRONPYTHON = sys.platform == 'cli'
+    pass
```
pekkaklarck: The above `try/except` should be removed because `IRONPYTHON` doesn't seem to be used for anything. Also, there's no need to have `pass` at the end of a non-empty block.

test/unit/utils/test_attrdict.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class TestAttrDict(unittest.TestCase):
+
+    def setUp(self):
+        self.dd = AttrDict([('z', 1), (2, 'y'), ('x', 3)])
```
pekkaklarck: It's a very good idea to have tests for `AttrDict` and copying Robot's tests for `DotDict` is fine too. `dd` is a bit weird name for `AttrDict` instance, though.

test/acceptance/create_webdriver.robot
```diff
@@ -29,6 +29,40 @@ Create Webdriver With Bad Keyword Argument Dictionary
...
+    Go To    ${FRONT PAGE}    # If we don't do this then no version details
+    ${capabilities}=    Get Browser Capabilities
+    Log    Capabilities = ${capabilities}
+    Log    Browser is ${capabilities['browserName']}, version ${capabilities['version']}
+    ${browser}=    Normalize Browser Name
+    Should Match    ${capabilities['browserName']}    ${browser}
+    Run Keyword If    "${browser}" != "internet explorer"    Log    Rotatable = ${capabilities['rotatable']}
+    # Disabled next step to PASS in RF pre 2.9
+    # Run Keyword If    "${browser}" == "internet explorer"    Run Keyword And Expect Error    *    Log
+    #     ...    Rotatable = ${capabilities['rotatable']}
```
pekkaklarck: Why wouldn't this pass with RF < 2.9? Leaving commented out tests here isn't a good idea anyway.

src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -507,6 +507,35 @@ def set_browser_implicit_wait(self, seconds):
...
+        | Log              | ${capabilities.may_not_exist}                       | # Produces AttributeError |
+
+        References, see [http://selenium-python.readthedocs.org/en/latest/api.html?highlight=desired_capabilities#selenium.webdriver.remote.webdriver.WebDriver.desired_capabilities|Desired Capabilities] and [https://selenium.googlecode.com/svn/trunk/docs/api/py/webdriver/selenium.webdriver.common.desired_capabilities.html#module-selenium.webdriver.common.desired_capabilities|Common Desired Capabilities].
+
+        New in Selenium2Library 1.8.0.
+        """
+
+        capabilities = self._current_browser().capabilities
+
+        return AttrDict(capabilities)
```
pekkaklarck, line 535: Unnecessary empty lines above. This PR removes some unnecessary empty lines elsewhere, which is fine, but shouldn't add new ones.

pekkaklarck:
Left some line notes. @aaltat, do you have time to look at the acceptance tests?

aaltat:
Yes, I can. I will do it today or tomorrow

aaltat:
Because the capabilities dictionary values do change between the different browsers, perhaps it would be best to test that there is certain set of keys is available in the dictionary. Perhaps I would not even look the dictionary values, because doing  `| Run Keyword If | ${browser} == ff | Do Something` makes the test hard to read. If it's possible to find common and good enough set of keys, which exist in all major browsers, like version, perhaps it would be enough to test that those keys exist.

test/acceptance/create_webdriver.robot
HelioGuilherme66: Before RF 2.9 it fails with exception in code before reaching the error. I though it is useful info to leave there.

HelioGuilherme66:
For now the common attributes tested are: browsername, version and platform.

test/acceptance/create_webdriver.robot
pekkaklarck: How, why and where does it fail?

pekkaklarck:
I agree with @aaltat that it would be good to test common capabilities between browsers. They should also be documented.

I also agree that `Run Keyword If` makes tests hard to read. It would be cleaner to have separate tests for features only certain browsers support and tag them with them like `only-firefox` and `no-firefox`. We've been doing something similar for Robot's own acceptance tests lately.

test/acceptance/create_webdriver.robot
```diff
@@ -29,6 +29,40 @@ Create Webdriver With Bad Keyword Argument Dictionary
     Run Keyword And Expect Error    kwargs must be a dictionary.
     ...    Create Webdriver    Firefox    kwargs={'spam': 'eggs'}
 
+Get Browser Capabilities Using Dictionary
+    [Documentation]    Gets current browser capabilities from WebDriver
+    [Setup]    Set Driver Variables
+    Create Webdriver    ${DRIVER_NAME}
+    Go To    ${FRONT PAGE}    # If we don't do this then no version details
```
pekkaklarck: Could this be fixed so that `Go To` would be needed? If not, then this limitation should probably be documented.

test/acceptance/create_webdriver.robot
HelioGuilherme66: It fails with KeyError:
10:15:54.103 FAIL Resolving variable '${capabilities['rotatable']}' failed: KeyError: 'rotatable' 
rotatable_key_not_in_ie

jerry57:
Excellent, I look forward to getting this merged and released.  THANKS!

HelioGuilherme66:
Just adding a reminder that it will be interesting to test this keyword against Selenium3.

aaltat:

vairamsvsjdo:
@HelioGuilherme66   is this a good candidate as an S2L extension. I would like to see many of these functionalities are made available as extension (plugin) and made available in a faster life cycle of usage before such features can be added as mainstream. Is this available as a plugin?

HelioGuilherme66:
Due to new developments, this code must be revisited, and more research done to match w3c webdriver standards. I have plans to do that after Python 3 compatibility is mainstream. I do not see the need to have it as a plugin, because it is accepted to be part of S2L.

vairamsvsjdo:
@HelioGuilherme66  
In general browser capability detection scripts use key/value pairs with or without JSON. IMHO, pushing through as dictionary should be perfect in that case.
For KB: What w3c webdriver standards and new developments is blocking?


http://www.isqft.com/cst.web/browserinfo/
https://msdn.microsoft.com/en-us/library/system.web.httpbrowsercapabilities.aspx
https://github.com/SeleniumHQ/selenium/wiki/DesiredCapabilities

vairamsvsjdo (approved):

HelioGuilherme66:
@aaltat @pekkaklarck I would be nice if we double-check, and review this keyword to get it in the next release. For example, what would be the final keyword name? Should we do an extensive research on the different browsers capabilities?

aaltat:
It's possible to get this in the next release, but it needs some work from you.

- Rebase with the master. Or it might be easier to get new branch from master and redo the changes because the library architecture has changed a lot. Do in a way which is easier for you and keeps the history intact.
- My comment about the test is still valid: Because the capabilities dictionary values do change between the different browsers, perhaps it would be best to test that there is certain set of keys is available in the dictionary. Perhaps I would not even look the dictionary values, because doing | Run Keyword If | ${browser} == ff | Do Something makes the test hard to read. If it's possible to find common and good enough set of keys, which exist in all major browsers, like version, perhaps it would be enough to test that those keys exist.
- I don't anymore remember, why we need the `AttrDict`?

aaltat:
Now I remember why we have `AttrDict`. I think it would be best to copy/paste Robot Framework DotDict implementation to SeleniumLibrary to `utils`.

aaltat:
Is there an issue for this PR? If not could you create one.

HelioGuilherme66:
Closing in favour of #981

PR #487, opened 2015-08-14
Fix dictionary creation

just-be-dev:
RF 2.9 deprecated creating dictionaries by using separated args. This
commit aims to fix that.


just-be-dev:
@HelioGuilherme66 want to review this for me?

Issue #489, opened 2015-08-20
Press Key keyword no longer works with strings


deaglejv:
Following example results in error (actual variable names replaced):

Press Key    ${EnterSomething}    test-something

This will error with message: ValueError: Unknown key named 'test-something'

just-be-dev:
@deaglejv, the update we made to the `Press Key` keyword in 1.7.3 had some unintended consequences. For now, I recommend rolling back to 1.7.2 if it's time critical, but we're definitely working hard to get a patched release out. 

The issue isn't that the `-` doesn't work, but rather that `Press Key` no longer accepts a string of characters to be pressed. This is mostly my fault for missing that detail in the code review. 

As I said, we'll have a patch out ASAP. Sorry for any inconvenience.

just-be-dev:
There's a few solutions that I can think of off the top of my head. 
- Check to see if a string with length > 1 matches a special key. If so, use the special key. Otherwise just send the string.

``` robotframework
Press Key   text_field    q          # The letter 'q'
Press Key   nav_console   ARROW_UP   # Named ARROW_UP key
Press Key   login_button  \\\\13     # ASCII code for Enter key
Press Key   custom_input   mystring  # String to be entered
```
- Make the special keys all some sort of variable that's built into the global context. 

``` robotframework
Press Key   nav_console   ${SELENIUM_KEYS.ARROW_UP}   #Arrow up key
```
- Add a new keyword that just performs `send_keys`

``` robotframework
Send Keys   custom_input   mystring   # String to be entered
```
- Add a optional argument that can take a string

``` robotframework
Press Key   custom_input   text=mystring   # String to be entered
```

I'm up for hearing any other suggestions as well. 

/cc @pekkaklarck, @emanlove, @HelioGuilherme66, @aaltat

pekkaklarck:
I think I'd just revert the change done in earlier version. This is so severe regression that a fix is needed ASAP and I doubt we can come up with a good design that quick.

pekkaklarck:
Possible new design should be backwards compatible with 1.7.1 and earlier. Bigger changes require deprecation in 1.8 first and may not be a good idea even then.

aaltat:
I agree with @pekkaklarck. Let's revert back the changes. It gives us time to consider what we need to do. Also I fear that jumping to new code, will lead to more problems than it will solve.

If we decide to make the changes, let's review the current test and make corrective actions to prevent this from happening. 

[...]

PR #491, opened 2015-08-22
Quick fix for #489.

HelioGuilherme66:
Press Key can send strings (if they do not match named keys).
With this PR, you can send strings of texts if they do not match exactly a named key.


pekkaklarck:
I don't think this is a good idea. I would rather revert the change and design this functionality along with `Press Keys` without a hurry. Also, `return(value)` is against PEP-8 in multiple ways and you should never ever leave commented out code into a PR.

aaltat:
I agree with @pekkaklarck

HelioGuilherme66:
Closed in favor of #493.

PR #493, opened 2015-08-25
Reverts Press Key to use double escaped Named keys.

### LINKED ISSUE #489: Press Key keyword no longer works with strings

HelioGuilherme66:
Fixes #489.

### DIFF
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -27,7 +27,6 @@ class _ElementKeywords(KeywordGroup):
         See `introduction` for details about locating elements.
         """
         return self._element_find(locator, True, True)
-
     def get_webelements(self, locator):
         """Returns list of WebElement objects matching locator.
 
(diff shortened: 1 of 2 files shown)


just-be-dev:
Was the escaped name keys in there before your changes?

HelioGuilherme66:
I think is was not added before. It was originated by this issue #275.

aaltat:
If there are lot of commits, which needs to reverted and there are less commits that we want to the new release. Perhaps it is easiest to checkout the 1.7.2 tag, cherry pick the commits that we want and leave others out and make new release. 

Or if the pull request #474 is the only one that we need to revert, then lets revert only that one and make a new release.

just-be-dev:
I think we need to roll back the changes all the way to 1.7.2 as @aaltat was suggesting. What about #494?.

aaltat:
After the revert, how does the git log looks like? I am not sure how github does it, but I would aim to a method which:
1) Does the revert with less pain
2) We can be sure that revert did the thing it should have been doing. 
3) Clean commit history (optional)

I can review for #494 and try to see, does it solve the problem. I am busy with next few days, but on Friday, I should have some time to spare. But also it would be really nice if someone else would also take a look.

HelioGuilherme66:
My revert PR is a cherry-pick only for the Press Key keyword. Let me add that the acceptance test passed also in Windows with IE 11.

just-be-dev:
I didn't revert a commit so much as make a new commit with the contents of what 1.7.2 had. This is one of those issues where squashing would've been immensely helpful, but we didn't do it. Lesson learned.

just-be-dev:
I'm going to close this for now. We'll create an issue on how to approach this in 1.8.

PR #494, opened 2015-08-25
Reverted press key changes

just-be-dev:
Completely reverted press key changes to how it was in 1.7.2.

### DIFF
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -485,26 +485,16 @@ return !element.dispatchEvent(evt);
 
     def press_key(self, locator, key):
         """Simulates user pressing key on element identified by `locator`.
-
-        `key` is either a single character, a numerical ASCII code of the key lead by '\\\\',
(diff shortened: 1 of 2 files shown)


just-be-dev:
@emanlove, @pekkaklarck, @aaltat 

Could any of you take a look at this? Feel free to test it as well. 

This is essentially a direct copy of what 1.7.2 had for this keyword.

pekkaklarck:
Using a direct copy of 1.7.2 version is fine. It would be a good idea to enhance docs to explain that multiple chars work too and add a test for the functionality. That should probably be a separate issue, though.

aaltat:
For my eyes it looks good. Code is more or less identical with 1.7.2, only few comments are missing. And the comments are only commenting out old code, so we are better not to have it in.

just-be-dev:
Yeah, I removed the commented out code intentionally, other than that it's exactly like 1.7.2. I did do another PR to add tests/documentation for the string support. That's all been released now in 1.7.4.

PR #496, opened 2015-08-27
Added keyword: Press Keys

Tset-Noitamotua:
Added 'Press Keys' keyword as introduced [here](https://github.com/rtomac/robotframework-selenium2library/pull/490)

NOTE: It does not support ASCII code for keys, but it should support all 'Special Keys' (Named Keys) which are documentated in [Selenium Python Bindings](http://selenium-python.readthedocs.org/en/latest/api.html#module-selenium.webdriver.common.keys)

NOTE: It is not thoroughly tested yet!

KNOWN ISSUE:  `CONTROL + +` or `CONTROL + ++`  does not ZOOM IN in a browser as expected but `CONTROL + -` does ZOOM OUT.

### DIFF
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -499,57 +499,57 @@ return !element.dispatchEvent(evt);
         element.send_keys(key)
 
     def press_keys(self, locator, keys):
-		"""Simulates user pressing one or more KEY(s) or SPECIAL KEY(s) simultaniuosly or in sequence
-		on element identified by `locator` while HOLDING DOWN one or more MODIFICATION KEY(s).
-		
-		`KEY`:					0-9, aA-zZ, -+*/% etc.
-		`SPECIAL KEY`:			F1-F12, DEL, TAB, ARROW_UP/_DOWN/_LEFT/_RIGHT etc.
-		`MODIFICATION KEY`:		ALT, ALT_GR, CTRL, R_CTRL, SHIFT, ALT_SHIFT, ALT_GR_SHIFT etc.
-		
-		`KEY` is either a single character, (TODO: or a numerical ASCII code of the key lead by '\\\\'),
-		or a sequence of single characters.
-		`SPECIAL KEY` and `MODIFICATION KEY` are special key names defined at selenium.webdriver.common.keys.
-		
-		Examples:	 # locator	 # key sequence				# usage example
-		| Press Keys | <locator> | CONTROL + a				| # select
-		| Press Keys | <locator> | CONTROL + A				| # select
-		| Press Keys | <locator> | CONTROL + C				| # copy
-		| Press Keys | <locator> | CONTROL + V				| # paste
-		| Press Keys | <locator> | CONTROL + ABC			| # custom
-		| Press Keys | <locator> | CONTROL + SHIFT + A		| # e.g. opens Firefox's Add-ons-Manager
-		| Press Keys | <locator> | CONTROL + ALT + DELETE	| # e.g. shows lock screen on Windows
-		| Press Keys | <locator> | CONTROL + ARROW_UP	 	| # and so on
-		"""
-		
-		keys =  keys.split(' + ')
-		
-		i = 0
-		named_keys = []
-		named_key_sequence = []
-		unnamed = ''
-		for key in keys:
-			try:
-				named_key = getattr(Keys, keys[i])
-				print "%s. named key is %s." % (i + 1, key)
-				named_keys.append(keys[i])
-				i = i + 1
-			except:
-				print "The rest '%s' is unnamed." % key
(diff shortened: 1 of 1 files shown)


aaltat:
Briefly did read the code and some observations:
1) Using variable number of arguments could result cleaner code. Also it does resolve the problem with CTRL++ key combination.
2) Didn't quite understand why there are two for loops, one should be enough. And if can do without any is best, but do not know is it possible in this case.
3) Using print directly is not allowed. Use the internal logging method.
4) Following pep008 is required
5) Write tests
6) Squash to single commit
7) Fixing the conflict is required

pekkaklarck:
I agree with the comments by @aaltat and have few more:
- Documentation is broken. You should read about the syntax from User Guide and, even more importantly, generate docs to see how it looks like. I'm not sure is generating docs for S2L actually documented anywhere, though. Do you @zephraph remember? Definitely something that should be mentioned in contribution guidelines.
- Using plain `except:` is a very bad practice. I think that's mentioned in PEP-8 that @aaltat already mentioned.
- I don't see why the code is using `exec`.

The most important concern I have about this is the overall usage and design. I think it would make sense to accept `*keys`, I'd rather have use just `+` and not `+` as a separator, etc. We should agree on these topics before trying to implement this. What would be the best place to discuss that? Is there an issue about `Press Keys` we could use or should we discuss on Slack? I think comments in the issue tracker would work best.

HelioGuilherme66:
I started a new topic in the Wiki at
https://github.com/rtomac/robotframework-selenium2library/wiki

What you think?

For this keyword it does not make much sense to have multiple PRs or
separate developments.

pekkaklarck:
I think issue tracker works better for discussing about enhancements. I agree we don't need new pull requests before we have agreed on the design.

Tset-Noitamotua:
@ all 
Thanx for your time and comments.

@aaltat

> 1) Using variable number of arguments could result cleaner code...

`Press Keys | CONTROL | ALT | DELETE` vs.
`Press Keys | CONTROL + ALT + DELETE`

Personally I prefer the last. But of course this may be a matter of taste ;)

> 2) Didn't quite understand why there are two for loops, one should be enough ... 

Fixed to one for loop. I don't know how to iterate through the attribute of 'webdriver.common.keys.Keys' without a for loop.

> 3) Using print directly is not allowed. Use the internal logging method.

Fixed

> 4) Following pep008 is required

Please help me with that. I'm a noob and my programming and python skills are very basic.

> 5) Write tests

Will do! I promise ;-)

> 6) Squash to single commit

What does it mean?

> 7) Fixing the conflict is required

I'll try. But I'll need help. Have no idea why `CONTROL + -` is working and `CONTROL + +` not.

pekkaklarck:
I submitted issue #498 about `Press Keys`. We can use that to discuss the design. I'll write my proposal there shortly, now trying to get RF 2.9.1 released.

Tset-Noitamotua:
@pekkaklarck 

> Using plain except: is a very bad practice.

What should one use instead?

> I don't see why the code is using exec.

Removed `exec`

pekkaklarck:
Why plain `except:` is not a good idea is discussed, along with other good recommendations, in PEP-8: https://www.python.org/dev/peps/pep-0008/#programming-recommendations

Tset-Noitamotua:
Noob question:

How can I make the following line work without 'exec'?
`exec("element.send_keys(%s, '%s')" % (named_key_seq_as_string, unnamed))` (This is what actually works.)

If I use `element.send_keys(named_key_seq_as_string, unnamed)` then just the string values of the two variables are send to the textfield. Here for example is the result of `Press Keys | CONTROL + A`:

image

It seems to work as expected with 'eval':
`element.send_keys(eval(named_key_seq_as_string), unnamed)`
Is that a better solution than with 'exec'? And why should one not use 'exec'?

pekkaklarck:
`eval` isn't any better here than `exec` here. They both evaluate Python code in a string that you pass to them. That's not a good practice when could just run the same code directly. You are basically doing `eval('Keys.%s' % name)` when you should use `getattr(Keys, name)`.

just-be-dev:
@pekkaklarck we haven't really documented how to format the documents. Our whole contribution guide hasn't really even started. 

So much to do, so little time.

Tset-Noitamotua:
I close this one in favor of  #498 .

inarvekar:
Do I need to include any other library for using Press keys ???
because I'm getting 
No keyword with name 'Press Keys' found. Did you mean:
    Selenium2Library.Press Key

HelioGuilherme66:
@inarvekar This is a Closed issue, and it was not Merged, so there is no `Press Keys` keyword. There is a long discussion at #498. In my fork there is a (probably outdated) implementation for that keyword.

By the way, you should ask questions for help on our other communication channels, which you can see [here](http://robotframework.org/#support).

PR #499, opened 2015-08-28
Fix for alerts in Chrome Browsers

aaltat:
This commit moves all the alert keyword in to a own _AlertKeywords
class. This did make it easier to fix the bug with WebDriverWait
class.


just-be-dev:
The build failed, but it's an unrelated issue. I queued it up again. Also, :+1: for the separation of alerts into its own file.

just-be-dev:
Looks good to me. Feel free to merge pending @pekkaklarck's feedback.

PR #501, opened 2015-09-01
When screen shot is taken file is not overwritten

aaltat:
When screen shot is taken and if the file allready
exit, the file name is changed to contain -001 in
the end. Example foobar.png is changed to:
foobar-1.png

And yes, this is take 3.


aaltat:
Closed because #504 offers better solution.

PR #504, opened 2015-09-03
Using Capture Screenshot with %d in filename

aaltat:
This commit enhances the Capture Screenshot keyword to support
internal running counter also when filename is not None. The place
of the counter is indicated with %d character.

Also the absolute path of the screenshot filename is always returned
by the keyword.

Also made small drive by refactoring to follow pep8 better.

### DIFF
--- a/src/Selenium2Library/keywords/_screenshot.py
+++ b/src/Selenium2Library/keywords/_screenshot.py
@@ -8,7 +8,7 @@ from keywordgroup import KeywordGroup
 class _ScreenshotKeywords(KeywordGroup):
 
     def __init__(self):
-        self._screenshot_index = 0
+        self._screenshot_index = {}
         self._screenshot_path_stack = []
         self.screenshot_root_directory = None
 
@@ -17,49 +17,86 @@ class _ScreenshotKeywords(KeywordGroup):
     def set_screenshot_directory(self, path, persist=False):
         """Sets the root output directory for captured screenshots.
 
-        ``path`` argument specifies the absolute path where the screenshots should
-        be written to. If the specified ``path`` does not exist, it will be created.
-        Setting ``persist`` specifies that the given ``path`` should
-        be used for the rest of the test execution, otherwise the path will be restored
-        at the end of the currently executing scope.
+        ``path`` argument specifies the absolute path where the screenshots
+        should be written to. If the specified ``path`` does not exist,
+        it will be created. Setting ``persist`` specifies that the given
+        ``path`` should be used for the rest of the test execution, otherwise
+        the path will be restored at the end of the currently executing scope.
         """
         path = os.path.abspath(path)
         self._create_directory(path)
         if persist is False:
             self._screenshot_path_stack.append(self.screenshot_root_directory)
             # Restore after current scope ends
-            utils.events.on('scope_end', 'current', self._restore_screenshot_directory)
+            utils.events.on('scope_end', 'current',
+                            self._restore_screenshot_directory)
 
         self.screenshot_root_directory = path
 
-    def capture_page_screenshot(self, filename=None):
+    def capture_page_screenshot(self,
+                                filename='selenium-screenshot-{index}.png'):
         """Takes a screenshot of the current page and embeds it into the log.
 
-        `filename` argument specifies the name of the file to write the
-        screenshot into. If no `filename` is given, the screenshot is saved into file
-        `selenium-screenshot-<counter>.png` under the directory where
-        the Robot Framework log file is written into. The `filename` is
+        ``filename`` argument specifies the name of the file to write the
+        screenshot into. If no ``filename`` is given, the screenshot is saved
+        into file _selenium-screenshot-{index}.png_ under the directory where
+        the Robot Framework log file is written into. The ``filename`` is
         also considered relative to the same directory, if it is not
         given in absolute format. If an absolute or relative path is given
         but the path does not exist it will be created.
 
-        `css` can be used to modify how the screenshot is taken. By default
-        the bakground color is changed to avoid possible problems with
-        background leaking when the page layout is somehow broken.
+        Starting from Selenium2Library 1.8 if ``filename`` contains _{index}_
+        characters, it will be automatically replaced with running index.
+        The running index is unique for each different filename. The absolute
+        path of the saved screenshot is always returned and it does not depend
+        does the ``filename`` contain _{index}_. See example 1 and 2 for more
+        details.
+
+        The _{index}_ is replaced with the actual index by using Python's
+        [https://docs.python.org/2/library/stdtypes.html#str.format|
+        str.format] method, and it can be formatted using the standard
+        [https://docs.python.org/2/library/string.html#format-string-syntax|
+        format string syntax]. The example 3 shows this by setting the width and
+        the fill character.
+
+        If there is a need to write literal _{index}_ or if ``filename``
+        contains _{_ or _}_ characters, then the braces must be doubled.
+
+        Example 1:
+        | ${file1} = | Capture Page Screenshot |
+        | File Should Exist | ${OUTPUTDIR}${/}selenium-screenshot-1.png |
+        | Should Be Equal | ${file1} | ${OUTPUTDIR}${/}selenium-screenshot-1.png |
+        | ${file2} = | Capture Page Screenshot |
+        | File Should Exist | ${OUTPUTDIR}${/}selenium-screenshot-2.png |
+        | Should Be Equal | ${file2} | ${OUTPUTDIR}${/}selenium-screenshot-2.png |
+
+        Example 2:
+        | ${file1} = | Capture Page Screenshot | ${OTHER_DIR}${/}other-{index}-name.png |
+        | ${file2} = | Capture Page Screenshot | ${OTHER_DIR}${/}some-other-name-{index}.png |
+        | ${file3} = | Capture Page Screenshot | ${OTHER_DIR}${/}other-{index}-name.png |
+        | File Should Exist | ${OTHER_DIR}${/}other-1-name.png |
+        | Should Be Equal | ${file1} | ${OTHER_DIR}${/}other-1-name.png |
+        | File Should Exist | ${OTHER_DIR}${/}some-other-name-1.png |
+        | Should Be Equal | ${file2} | ${OTHER_DIR}${/}some-other-name-1.png |
+        | File Should Exist | ${OTHER_DIR}${/}other-2-name.png |
+        | Should Be Equal | ${file3} | ${OTHER_DIR}${/}other-2-name.png |
+
+        Example 3:
+        | Capture Page Screenshot | ${OTHER_DIR}${/}sc-{index:06}.png |
+        | File Should Exist | ${OTHER_DIR}${/}sc-000001.png |
         """
         path, link = self._get_screenshot_paths(filename)
         self._create_directory(path)
-
         if hasattr(self._current_browser(), 'get_screenshot_as_file'):
-          if not self._current_browser().get_screenshot_as_file(path):
-              raise RuntimeError('Failed to save screenshot ' + filename)
+            if not self._current_browser().get_screenshot_as_file(path):
+                raise RuntimeError('Failed to save screenshot ' + link)
         else:
-          if not self._current_browser().save_screenshot(path):
-            raise RuntimeError('Failed to save screenshot ' + filename)
-
+            if not self._current_browser().save_screenshot(path):
+                raise RuntimeError('Failed to save screenshot ' + link)
         # Image is shown on its own row and thus prev row is closed on purpose
         self._html('</td></tr><tr><td colspan="3"><a href="%s">'
                    '<img src="%s" width="800px"></a>' % (link, link))
+        return path
 
     # Private
     def _create_directory(self, path):
@@ -87,14 +124,17 @@ class _ScreenshotKeywords(KeywordGroup):
         self.screenshot_root_directory = self._screenshot_path_stack.pop()
 
     def _get_screenshot_paths(self, filename):
-        if not filename:
-            self._screenshot_index += 1
-            filename = 'selenium-screenshot-%d.png' % self._screenshot_index
-        else:
-            filename = filename.replace('/', os.sep)
-
-        screenshotDir = self._get_screenshot_directory()
-        logDir = self._get_log_dir()
-        path = os.path.join(screenshotDir, filename)
-        link = robot.utils.get_link_path(path, logDir)
+        filename = filename.format(
+            index=self._get_screenshot_index(filename))
+        filename = filename.replace('/', os.sep)
+        screenshotdir = self._get_screenshot_directory()
+        logdir = self._get_log_dir()
+        path = os.path.join(screenshotdir, filename)
+        link = robot.utils.get_link_path(path, logdir)
         return path, link
+
+    def _get_screenshot_index(self, filename):
+        if filename not in self._screenshot_index:
+            self._screenshot_index[filename] = 0
+        self._screenshot_index[filename] += 1
+        return self._screenshot_index[filename]
--- a/test/acceptance/keywords/screenshots.robot
+++ b/test/acceptance/keywords/screenshots.robot
@@ -7,9 +7,10 @@ Resource          ../resource.robot
 Capture page screenshot to default location
     [Documentation]    LOG 2:3 REGEXP: </td></tr><tr><td colspan="3"><a href="selenium-screenshot-\\d.png"><img src="selenium-screenshot-\\d.png" width="800px"></a>
     [Setup]    Remove Files    ${OUTPUTDIR}/selenium-screenshot-*.png
-    Capture Page Screenshot
+    ${file} =    Capture Page Screenshot
     ${count} =    Count Files In Directory    ${OUTPUTDIR}    selenium-screenshot-*.png
     Should Be Equal As Integers    ${count}    1
+    Should Be Equal    ${file}    ${OUTPUTDIR}${/}selenium-screenshot-1.png
     Click Link    Relative
     Wait Until Page Contains Element    tag=body
     Capture Page Screenshot
@@ -38,11 +39,42 @@ Capture page screenshot to custom root directory
     [Documentation]    Capture page screenshot to custom root directory
     [Setup]    Remove Directory    ${OUTPUTDIR}/custom-root    recursive
     Set Screenshot Directory    ${OUTPUTDIR}/custom-root
-    Capture Page Screenshot    custom-root-screenshot.png
+    ${file} =    Capture Page Screenshot    custom-root-screenshot.png
     File Should Exist    ${OUTPUTDIR}/custom-root/custom-root-screenshot.png
+    Should Be Equal    ${file}    ${OUTPUTDIR}${/}custom-root${/}custom-root-screenshot.png
 
 Ensure screenshot captures revert to default root directory
     [Documentation]    Ensure screenshot captures revert to default root directory
     [Setup]    Remove Files    ${OUTPUTDIR}/default-root-screenshot.png
     Capture Page Screenshot    default-root-screenshot.png
     File Should Exist    ${OUTPUTDIR}/default-root-screenshot.png
+
+Capture page screenshot with unique index
+    [Setup]    Remove Directory    ${OUTPUTDIR}${/}screenshot-and-index    recursive
+    ${file1} =    Capture Page Screenshot    ${OUTPUTDIR}${/}screenshot-and-index${/}other-{index}-name.png
+    ${file2} =    Capture Page Screenshot    ${OUTPUTDIR}${/}screenshot-and-index${/}some-other-name-{index}.png
+    ${file3} =    Capture Page Screenshot    ${OUTPUTDIR}${/}screenshot-and-index${/}other-{index}-name.png
+    File Should Exist    ${OUTPUTDIR}${/}screenshot-and-index${/}other-1-name.png
+    Should Be Equal    ${file1}    ${OUTPUTDIR}${/}screenshot-and-index${/}other-1-name.png
+    File Should Exist    ${OUTPUTDIR}${/}screenshot-and-index${/}some-other-name-1.png
+    Should Be Equal    ${file2}    ${OUTPUTDIR}${/}screenshot-and-index${/}some-other-name-1.png
+    File Should Exist    ${OUTPUTDIR}${/}screenshot-and-index${/}other-2-name.png
+    Should Be Equal    ${file3}    ${OUTPUTDIR}${/}screenshot-and-index${/}other-2-name.png
+
+Capturing a page screenshot with two indexes should not cause an error
+    ${file} =    Capture Page Screenshot    ${OUTPUTDIR}${/}screenshot-and-index${/}two-{index}-in-{index}-name.png
+    File Should Exist    ${OUTPUTDIR}${/}screenshot-and-index${/}two-1-in-1-name.png
+    Should Be Equal    ${file}    ${OUTPUTDIR}${/}screenshot-and-index${/}two-1-in-1-name.png
+
+Capture page screenshot with index formatting
+    Capture Page Screenshot    ${OUTPUTDIR}${/}screenshot-and-index${/}format-{index:06}-name.png
+    Capture Page Screenshot    ${OUTPUTDIR}${/}screenshot-and-index${/}format-{index:06}-name.png
+    File Should Exist    ${OUTPUTDIR}${/}screenshot-and-index${/}format-000001-name.png
+    File Should Exist    ${OUTPUTDIR}${/}screenshot-and-index${/}format-000002-name.png
+
+Capture page screenshot with escaped braces
+    ${file} =    Capture Page Screenshot    ${OUTPUTDIR}${/}screenshot-and-index${/}brackets-{{index}}-name.png
+    File Should Exist    ${OUTPUTDIR}${/}screenshot-and-index${/}brackets-{index}-name.png
+    Should Be Equal    ${file}    ${OUTPUTDIR}${/}screenshot-and-index${/}brackets-{index}-name.png
+    ${file} =    Capture Page Screenshot    ${OUTPUTDIR}${/}screenshot-and-index${/}brackets-{{index-name.png
+    File Should Exist    ${OUTPUTDIR}${/}screenshot-and-index${/}brackets-{index-name.png


just-be-dev:
:+1:

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -17,49 +17,77 @@ def __init__(self):
...
-        background leaking when the page layout is somehow broken.
+        New in Selenium2Library 1.8
+
+        If ``filename`` contains %d character, it will be automatically
+        replaced with running counter. The running counter is unique for each
+        different filename. The absolute path of the saved screenshot is always
+        returned.
+
+        If ``filename`` contains two or more %d characters or there is only one
+        or more % characters in filename, then a TypeError exception is raised.
```
just-be-dev: Why would an exception be raised because of there being an extra `%` in the name?

aaltat: It is because of Python older string formatting, which the code uses, see in here: https://github.com/robotframework/Selenium2Library/pull/504/files#diff-5346a546c6b37e3c016ad51e7155887dR122. Example this: `python -c "'foo % bar %d' % 'string'"` will also cause exception. There is bug in documentation, that exception is not always TypeError and it would be best to say only exception. 

The Python 2.6 onward supports `str.format()`, which could solve the problem in different way but because we must support Python 2.5, I did not use it. I could also do simple `str.replace()` in the code, but I would like to keep some formatting options open for future enhancements.

pekkaklarck:
I initially proposed `%d` as a placeholder for the index. It's an interesting idea to possible support also other formatting options, but after thinking a bit I don't think that would be very useful. Such formatting would work well if there would be other automatically generated information, but I doubt the would be. Formatting wouldn't be too useful in hypothetical cases like

```
Capture Page Screenshot    foo-{bar}-{index}.png    bar=zap
Capture Page Screenshot    foo-{bar}-{index}.png    bar=${var}
```

because you could just use

```
Capture Page Screenshot    foo-zap-{index}.png
Capture Page Screenshot    foo-${var}-{index}.png
```

Because I think other formatting isn't likely very useful, I don't think we need to take that into account. That means we could easily use something more explicit than `%d` to specify an index is needed. We could, for example, allow `<index>` or `:index:` and just use `.replace('<index>', index)` in the code.

pekkaklarck:
Also, @aaltat mentioned that `str.format` cannot be used because Python 2.5 needs to be supported. I think Selenium has dropped 2.5 support years ago so that shouldn't be an issue. Is Python 2.5 support promised somewhere?

aaltat:
It is not promised directly. But we do claim to support quite old versions of Robot Framework (2.6 is the minimum version) and older versions do support Python 2.5. How realistic this is, ido not know.

But if we announced that we only support 2.6 onwards, it would allow to use `str.format`. I still keep the formatting option open because it would be really nice to get example 00001, 00002... and not 1, 2... Or perhaps adding epoch would be nice too.

pekkaklarck:
As I already wrote, Selenium itself hasn't supported Py2.5 for years. Do you think someone would be using new S2L and downgrade Se separately to an ancient version?

Supporting formatting like leading zeros is a very good point. Supporting `{index}` and using `.format()` sounds like a good idea.

aaltat:
Ah my mistake, in [INSTALL.rst](https://github.com/robotframework/Selenium2Library/blob/master/INSTALL.rst) we say that 2.6 is the minimum version. Because of the trouble caused by `%d`, I will change the implementation so direction which was suggested by @pekkaklarck. The `str.format` feels also quite temping to use, because it opens quite nice [formatting possibilities](https://docs.python.org/2/library/string.html#format-examples). Perhaps it would be best to do something like this: `python -c "print 'file-name-{index}.png'.format(index=1)"`

aaltat:
Refactored the pull request to use `str.format`

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -17,49 +17,85 @@ def __init__(self):
...
+        into file `selenium-screenshot-{index}.png` under the directory where
+        the Robot Framework log file is written into. The ``filename`` is
         also considered relative to the same directory, if it is not
         given in absolute format. If an absolute or relative path is given
         but the path does not exist it will be created.
 
-        `css` can be used to modify how the screenshot is taken. By default
-        the bakground color is changed to avoid possible problems with
-        background leaking when the page layout is somehow broken.
+        New in Selenium2Library 1.8
```
pekkaklarck: This note should be more explicit about what's new in 1.8. I would also either move it at the end of the doc or embed it into the doc like `Starting from version 1.8, if ...`.

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -17,49 +17,85 @@ def __init__(self):
...
+
+        If ``filename`` contains {index} characters, it will be automatically
+        replaced with running index. The running index is unique for each
+        different filename. The absolute path of the saved screenshot is always
+        returned and it does not depend does the ``filename`` contain {index}.
+        See example 1 and 2 for more details.
+
+        The {index} uses Python [https://docs.python.org/2/library/string.html#string-formatting|String Formatting]
+        and it is possible to format the {index} with [https://docs.python.org/2/library/string.html#grammar-token-width|width]
+        . The example 3 uses width to format the {index} to desired length.
```
pekkaklarck: I would prefer long links on their own lines. Alternatively you can split them from `|`. Also, no space before the `.` character.

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -17,49 +17,85 @@ def __init__(self):
...
         also considered relative to the same directory, if it is not
         given in absolute format. If an absolute or relative path is given
         but the path does not exist it will be created.
 
-        `css` can be used to modify how the screenshot is taken. By default
-        the bakground color is changed to avoid possible problems with
-        background leaking when the page layout is somehow broken.
+        New in Selenium2Library 1.8
+
+        If ``filename`` contains {index} characters, it will be automatically
```
pekkaklarck: It would probably be a good idea to enclose also `{index}` in double backticks.

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -17,49 +17,85 @@ def __init__(self):
...
     def capture_page_screenshot(self, filename=None):
         """Takes a screenshot of the current page and embeds it into the log.
 
-        `filename` argument specifies the name of the file to write the
-        screenshot into. If no `filename` is given, the screenshot is saved into file
-        `selenium-screenshot-<counter>.png` under the directory where
-        the Robot Framework log file is written into. The `filename` is
+        ``filename`` argument specifies the name of the file to write the
+        screenshot into. If no ``filename`` is given, the screenshot is saved
+        into file `selenium-screenshot-{index}.png` under the directory where
```
pekkaklarck: It's good that single backticks have been replaced with double backticks in arguments because resulting formatting looks better and there's no risk for accidental linking. I think also `selenium-screenshot-{index}.png` should get the same treatment. Alternatively it could be formatted to be italics my surrounding it with `_`.

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -17,49 +17,85 @@ def __init__(self):
...
+        | Should Be Equal | ${dir1} | ${OTHER_DIR}${/}other-1-name.png |
+        | File Should Exist | ${OTHER_DIR}${/}some-other-name-1.png |
+        | Should Be Equal | ${dir2} | ${OTHER_DIR}${/}some-other-name-1.png |
+        | File Should Exist | ${OTHER_DIR}${/}other-2-name.png |
+        | Should Be Equal | ${dir3} | ${OTHER_DIR}${/}other-2-name.png |
+
+        Example 3:
+        | Capture Page Screenshot | ${OTHER_DIR}${/}sc-{index:06}.png |
+        | File Should Exist | ${OTHER_DIR}${/}sc-000001.png |
+
```
pekkaklarck: Extra empty line.

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -88,13 +124,25 @@ def _restore_screenshot_directory(self):
 
     def _get_screenshot_paths(self, filename):
         if not filename:
-            self._screenshot_index += 1
-            filename = 'selenium-screenshot-%d.png' % self._screenshot_index
+            filename = 'selenium-screenshot-{index}.png'
+            index = self._get_screenshot_index(filename)
+            filename = filename.format(index=index)
+        elif '{index' in filename:
```
pekkaklarck: This `elif` could just be `if` and the two lines above it could be removed.

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -88,13 +124,25 @@ def _restore_screenshot_directory(self):
...
+            index = self._get_screenshot_index(filename)
+            filename = filename.format(index=index)
+            filename = filename.replace('/', os.sep)
         else:
             filename = filename.replace('/', os.sep)
```
pekkaklarck: There's quite a bit of duplication above. At least the `elif` could be changed to `if` and two lines above it removed, but perhaps the code could be simplified to this:

``` python
    def _get_screenshot_paths(self, filename):
        if not filename:
            filename = 'selenium-screenshot-{index}.png'
        else:
            filename = filename.replace('/', os.sep)
        filename = filename.format(index=self._get_screenshot_index(filename))
```

The above code behaves differently if the path has something like `{foo}` on it, but the original doesn't handle it either if there's also `{index`. I wouldn't care about that here at all, I'd just document that if the path contains `{` or `}`, they must be doubled.

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -88,13 +124,25 @@ def _restore_screenshot_directory(self):
...
+        link = robot.utils.get_link_path(path, logdir)
         return path, link
+
+    def _get_screenshot_index(self, filename):
+        try:
+            self._screenshot_index[filename] += 1
+            return self._screenshot_index[filename]
+        except KeyError:
+            self._screenshot_index[filename] = 1
+            return 1
```
pekkaklarck: Cleaner way to write the above:

``` python
if filename not in self._screenshot_index:
    self._screenshot_index[filename] = 0
self._screenshot_index[filename] += 1
return self._screenshot_index[filename]
```

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -17,49 +17,85 @@ def __init__(self):
...
         if persist is False:
             self._screenshot_path_stack.append(self.screenshot_root_directory)
             # Restore after current scope ends
-            utils.events.on('scope_end', 'current', self._restore_screenshot_directory)
+            utils.events.on('scope_end', 'current',
+                            self._restore_screenshot_directory)
 
         self.screenshot_root_directory = path
 
     def capture_page_screenshot(self, filename=None):
```
pekkaklarck: Hmm, I don't see any reason why this couldn't and shouldn't be `filename='selenium-screenshot-{index}.png'`. It would make the default more explicit, simplify documentation, and avoid on `if` in the code.

src/Selenium2Library/keywords/_screenshot.py
aaltat: Went for `Starting from version 1.8, if ...`

src/Selenium2Library/keywords/_screenshot.py
aaltat: Done

src/Selenium2Library/keywords/_screenshot.py
aaltat: When for italic with `_{index}_` in code side.

src/Selenium2Library/keywords/_screenshot.py
aaltat: Done, with underscore.

src/Selenium2Library/keywords/_screenshot.py
aaltat: Done

src/Selenium2Library/keywords/_screenshot.py
aaltat: Went for a sightly different option and simplified the code even more. Did document the `{` must be now doubled and did write test for it.

Looking the code for new eyes, I do not understand why I did created that peace of....

src/Selenium2Library/keywords/_screenshot.py
aaltat: Done

src/Selenium2Library/keywords/_screenshot.py
aaltat: Done

aaltat:
@pekkaklarck Thanks for good review.

just-be-dev:
I just read through it and everything looks good to me. You've got my :+1: to merge whenever you feel this is ready.

aaltat:
If @pekkaklarck is also happy, then I am also ready for merge

aaltat:
Squashed all to commits to one, because the middle commits where not meaningful alone.

aaltat:
Hmm some other test did fail, can I somehow push Travis to rerun the tests?

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -17,49 +17,96 @@ def __init__(self):
...
+        the path will be restored at the end of the currently executing scope.
         """
         path = os.path.abspath(path)
         self._create_directory(path)
         if persist is False:
             self._screenshot_path_stack.append(self.screenshot_root_directory)
             # Restore after current scope ends
-            utils.events.on('scope_end', 'current', self._restore_screenshot_directory)
+            utils.events.on('scope_end', 'current',
+                            self._restore_screenshot_directory)
```
pekkaklarck, line 32: Are these changes here to avoid lines longer than 80 chars? That is OK in general, but such changes should be submitted separately. Preferably so that the whole module is gone through in one go. In minor cases it's also a bit questionable is the slightly enhanced PEP-8 compatibility worth an extra change in the version history.

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -17,49 +17,96 @@ def __init__(self):
...
 
         self.screenshot_root_directory = path
 
-    def capture_page_screenshot(self, filename=None):
+    def capture_page_screenshot(
+            self, filename='selenium-screenshot-{index}.png'):
```
pekkaklarck: This indentation looks a bit weird. I would typically have at least `self` on the same rows as the `def`. If that causes the next argument line to be slightly over 80 chars I wouldn't care.

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -17,49 +17,96 @@ def __init__(self):
...
+        does the ``filename`` contain _{index}_. See example 1 and 2 for more
+        details.
+
+        The _{index}_ uses Python
+        [https://docs.python.org/2/library/string.html#string-formatting|
+        String Formatting] and it is possible to format the _{index}_ with
+        [https://docs.python.org/2/library/string.html#grammar-token-width|width].
+        The example 3 uses width to format the _{index}_ to desired length.
+        The width is also the only tested and supported option for _{index}_.
+        Other formatting options may also work but are not tested/supported.
```
pekkaklarck: I think this documentation goes too much into details and the comment about other options not being supported is a bit strange. I see absolutely no reason why they wouldn't work, and the example in fact already sets also the fill character. I would use something simpler like this:

> The `{index}` is replaced with the actual index by using Python's [str.format](https://docs.python.org/2/library/stdtypes.html#str.format) method, and it can be formatted using the standard [format string syntax](https://docs.python.org/2/library/string.html#format-string-syntax). The example 3 shows this by setting the width and the fill character.

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -17,49 +17,96 @@ def __init__(self):
...
+
+        Example 3:
+        | Capture Page Screenshot | ${OTHER_DIR}${/}sc-{index:06}.png |
+        | File Should Exist | ${OTHER_DIR}${/}sc-000001.png |
+
+        Example 4
+        | Capture Page Screenshot | ${OUTPUTDIR}${/}braces-{{index}}.png |
+        | File Should Exist | ${OUTPUTDIR}${/}braces-{index}.png |
+        | Capture Page Screenshot | ${OUTPUTDIR}${/}brace-{{index.png |
+        | File Should Exist | ${OUTPUTDIR}${/}brace-{index.png |
```
pekkaklarck: I don't think this example is needed. I doubt anyone really needs `{}` in the name. Because that is so unlikely and the [referenced Python docs](https://docs.python.org/2/library/string.html#format-string-syntax) anyway explain that `{}` must be doubled, I think we don't need to mention it in our docs at all.

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -87,14 +134,17 @@ def _restore_screenshot_directory(self):
...
-        logDir = self._get_log_dir()
-        path = os.path.join(screenshotDir, filename)
-        link = robot.utils.get_link_path(path, logDir)
+        filename = filename.format(
+            index=self._get_screenshot_index(filename))
+        filename = filename.replace('/', os.sep)
+        screenshotdir = self._get_screenshot_directory()
+        logdir = self._get_log_dir()
+        path = os.path.join(screenshotdir, filename)
+        link = robot.utils.get_link_path(path, logdir)
```
pekkaklarck, line 133: This should actually be replaced with `os.path.relpath`. It's new in Python 2.6 and works identically than Robot's older `utils.get_link_path`.

src/Selenium2Library/keywords/_screenshot.py
```diff
@@ -17,49 +17,85 @@ def __init__(self):
...
+
         """
         path, link = self._get_screenshot_paths(filename)
         self._create_directory(path)
 
         if hasattr(self._current_browser(), 'get_screenshot_as_file'):
-          if not self._current_browser().get_screenshot_as_file(path):
-              raise RuntimeError('Failed to save screenshot ' + filename)
+            if not self._current_browser().get_screenshot_as_file(path):
+                raise RuntimeError('Failed to save screenshot ' + link)
```
pekkaklarck, line 92: I commented above that unrelated PEP-8 changes typically should be submitted separately. Indentation fix to code that is anyway touched is definitely fine. I might even remove the unnecessary empty lines around this if/else block...

src/Selenium2Library/keywords/_screenshot.py
just-be-dev: I disagree. Maybe just linking to that section of the documentation would be enough, but you never know what a user will attempt to do. It at least needs to be mentioned.

src/Selenium2Library/keywords/_screenshot.py
just-be-dev, line 32: I mostly agree with what @pekkaklarck said. We really should keep changes focused. With that being said, I'm not necessarily against this staying. I'll defer to you two on that decision.

src/Selenium2Library/keywords/_screenshot.py
just-be-dev, line 133: :+1: Agreed.

src/Selenium2Library/keywords/_screenshot.py
just-be-dev, line 92: I'd just leave this part as is.

src/Selenium2Library/keywords/_screenshot.py
aaltat, line 32: Yes, modifications where done do void too long lines. I can can remove it from the commit if it's needed.

src/Selenium2Library/keywords/_screenshot.py
aaltat: This is one of the [PEP-8 recommendations](https://www.python.org/dev/peps/pep-0008/#indentation)

```
def long_function_name(
        var_one, var_two, var_three,
        var_four):
    print(var_one)
```

But I do not feel strongly about it, it complies to PEP-8 also when done like this:

```
def capture_page_screenshot(self,
                            filename='selenium-screenshot-{index}.png'):
```

Done

src/Selenium2Library/keywords/_screenshot.py
aaltat: True. Too many details, yours is way better.

src/Selenium2Library/keywords/_screenshot.py
aaltat: I agree that it is more or less copy&paste what the Python docs do say. In the other hand, I do feel that not many will actually read the Python docs that carefully (if at all).

Perhaps good solution is to remove the examples and keep the note in the doc. The kw doc is in any case quite big now.

src/Selenium2Library/keywords/_screenshot.py
aaltat, line 92: I did change the filename -> link and in the same time fixed the PEP-8. I was thinking same as @pekkaklarck to remove the empty lines around if/else but was thinking that it should be done in separate commit or not. And then did not do ti.

src/Selenium2Library/keywords/_screenshot.py
aaltat: Done

aaltat, line 133: It almost work identically. One of the test do fail, because, if the `${TEMPDIR} ==  c:\users\user_name\appdata\local\temp` but the `${OUTPUTDIR} ==
D:\workspace\robotframework-selenium2library\test\results` then os.path.relpath fails on error: 
`ValueError: path is on drive c:, start on drive D:`

```
  File "<string>", line 2, in capture_page_screenshot
  File "D:\workspace\robotframework-selenium2library\src\Selenium2Library\keywords\keywordgroup.py", line 15, in _run_on_failure_decorator
    return method(*args, **kwargs)
  File "D:\workspace\robotframework-selenium2library\src\Selenium2Library\keywords\_screenshot.py", line 87, in capture_page_screenshot
    path, link = self._get_screenshot_paths(filename)
  File "D:\workspace\robotframework-selenium2library\src\Selenium2Library\keywords\_screenshot.py", line 135, in _get_screenshot_paths
    link = os.path.relpath(path, logdir)
  File "C:\Python27\lib\ntpath.py", line 528, in relpath
    % (path_prefix, start_prefix))
```

Should I leave it as it was, try to fix the error in this commit or change the to use `os.path.relpath` in another commit. I am on favor of the last one.

src/Selenium2Library/keywords/_screenshot.py
pekkaklarck, line 32: No need to remove. Unrelated changes just add noise that makes the review a bit harder. Now we all already know why the changes are here and removing them wouldn't win much.

src/Selenium2Library/keywords/_screenshot.py
pekkaklarck: Ok, didn't remember that recommendation. Both formats are fine. Also, I don't think 80 char line length is so important rule that it should mess up other formatting.

src/Selenium2Library/keywords/_screenshot.py
pekkaklarck: Keeping the note is fine. I still think the example isn't that useful because this is so rare case, but I definitely don't feel strongly about it.

src/Selenium2Library/keywords/_screenshot.py
pekkaklarck, line 92: From reviewers point of view, removing unnecessary empty lines is typically fine because understanding the change takes practically no time. Same with indentation fixes, especially when done for the code touched anyway.

src/Selenium2Library/keywords/_screenshot.py
pekkaklarck, line 133: Oh, I seriously thought these two functions would be identical, but on a closer look there seem to be also other small differences. Please keep it like it is. Due to differences it is unlikely that `get_link_path` will be removed from Robot anyway.

test/acceptance/keywords/screenshots.robot
```diff
@@ -38,11 +39,42 @@ Capture page screenshot to custom root directory
...
+    Capture Page Screenshot    ${OUTPUTDIR}${/}screenshot-and-index${/}format-{index:06}-name.png
+    File Should Exist    ${OUTPUTDIR}${/}screenshot-and-index${/}format-000001-name.png
+    File Should Exist    ${OUTPUTDIR}${/}screenshot-and-index${/}format-000002-name.png
+
+Capture page screenshot with braces should now be doupled
+        ${file} =    Capture Page Screenshot    ${OUTPUTDIR}${/}screenshot-and-index${/}brackets-{{index}}-name.png
+        File Should Exist    ${OUTPUTDIR}${/}screenshot-and-index${/}brackets-{index}-name.png
+        Should Be Equal    ${file}    ${OUTPUTDIR}${/}screenshot-and-index${/}brackets-{index}-name.png
+        ${file} =    Capture Page Screenshot    ${OUTPUTDIR}${/}screenshot-and-index${/}brackets-{{index-name.png
+        File Should Exist    ${OUTPUTDIR}${/}screenshot-and-index${/}brackets-{index-name.png
```
pekkaklarck: Newline missing at the end of the file.

pekkaklarck:
I added few more line notes. Please take a look at them @aaltat and fix ones that you see worth fixing. Would you @zephraph like take a final look at this afterwards and merge?

just-be-dev:
Sure thing.

test/acceptance/keywords/screenshots.robot
aaltat: Done

aaltat:
I did fix it based on the comments from @pekkaklarck. @zephraph could you do review also.

test/acceptance/keywords/screenshots.robot
```diff
@@ -38,11 +39,42 @@ Capture page screenshot to custom root directory
...
+    ${file2} =    Capture Page Screenshot    ${OUTPUTDIR}${/}screenshot-and-index${/}some-other-name-{index}.png
+    ${file3} =    Capture Page Screenshot    ${OUTPUTDIR}${/}screenshot-and-index${/}other-{index}-name.png
+    File Should Exist    ${OUTPUTDIR}${/}screenshot-and-index${/}other-1-name.png
+    Should Be Equal    ${file1}    ${OUTPUTDIR}${/}screenshot-and-index${/}other-1-name.png
+    File Should Exist    ${OUTPUTDIR}${/}screenshot-and-index${/}some-other-name-1.png
+    Should Be Equal    ${file2}    ${OUTPUTDIR}${/}screenshot-and-index${/}some-other-name-1.png
+    File Should Exist    ${OUTPUTDIR}${/}screenshot-and-index${/}other-2-name.png
+    Should Be Equal    ${file3}    ${OUTPUTDIR}${/}screenshot-and-index${/}other-2-name.png
+
+Capture page screenshot whith two index do not cause error
```
just-be-dev: There's a typo here. Plus it's worded a little weird. Maybe it should be something like

```
Capturing a page screenshot with two indexes should not cause an error
```

Or something, idk. Regardless, it could definitely be reworded.

just-be-dev:
@aaltat, have you ran this locally to make sure that the screenshots really do show up?

test/acceptance/keywords/screenshots.robot
aaltat: True, wording is not correct and your example is better. But should I remove the hole test and what you mean: worked weird?

aaltat:
I did manualy verify that screen captures are visible in the log.

test/acceptance/keywords/screenshots.robot
aaltat: Changed to name

test/acceptance/keywords/screenshots.robot
```diff
@@ -38,11 +39,42 @@ Capture page screenshot to custom root directory
...
+    Should Be Equal    ${file2}    ${OUTPUTDIR}${/}screenshot-and-index${/}some-other-name-1.png
+    File Should Exist    ${OUTPUTDIR}${/}screenshot-and-index${/}other-2-name.png
+    Should Be Equal    ${file3}    ${OUTPUTDIR}${/}screenshot-and-index${/}other-2-name.png
+
+Capturing a page screenshot with two indexes should not cause an error
+    ${file} =    Capture Page Screenshot    ${OUTPUTDIR}${/}screenshot-and-index${/}two-{index}-in-{index}-name.png
+    File Should Exist    ${OUTPUTDIR}${/}screenshot-and-index${/}two-1-in-1-name.png
+    Should Be Equal    ${file}    ${OUTPUTDIR}${/}screenshot-and-index${/}two-1-in-1-name.png
+
+Capture page screenshot whith index formatting
```
just-be-dev: Typo. `whith` -> `with`.

test/acceptance/keywords/screenshots.robot
```diff
@@ -38,11 +39,42 @@ Capture page screenshot to custom root directory
...
+    Capture Page Screenshot    ${OUTPUTDIR}${/}screenshot-and-index${/}format-{index:06}-name.png
+    Capture Page Screenshot    ${OUTPUTDIR}${/}screenshot-and-index${/}format-{index:06}-name.png
+    File Should Exist    ${OUTPUTDIR}${/}screenshot-and-index${/}format-000001-name.png
+    File Should Exist    ${OUTPUTDIR}${/}screenshot-and-index${/}format-000002-name.png
+
+Capture page screenshot with braces should now be doupled
```
just-be-dev: Perhaps rename this to `Capture page screenshot with escaped braces`. Current title doesn't make sense to me.

just-be-dev:
Found two more small typos. Otherwise, it looks good to me. Fix those and I'll merge it in. 

Thanks for your patience!

test/acceptance/keywords/screenshots.robot
aaltat: Done

test/acceptance/keywords/screenshots.robot
aaltat: Done

aaltat:
Fixed based on @zephraph comments. 

I think, this is normal iterative development process. I did have an idea, or actually problem to solve. Made a first attempt to fix it and honesty it was quite horrible attempt. Then there was new idea from the review process, which was better, but caused lot of problems/corner cases. And on last attempt, we did nail it down quite well. 

Also I am used to review process, it is nice way to share the information between people. It also yields to better results (and this is fine example of one), like better functionality, cleaner code and so one.

pekkaklarck:
Agreed.

just-be-dev:
:+1:

PR #514, opened 2015-09-22
Get size will return size of element as (width,height)

SergiuTudos:


just-be-dev:
Make sure you add acceptance tests too.

SergiuTudos:
it was added:
test/acceptance/keywords/elements.robot

Get Element Size
    ${width}  ${height}=  Get Element Size  link=Link
    Should be True  ${height} > ${0}
    Should be True  ${width} > ${0}
    Run Keyword And Expect Error  Could not determine size for 'non-existent'  Get Element Size  non-existent

aaltat:
The last keyword does not test right thing. It test that if element is not found error is received. Could you write tests:
Element that is hidden, no idea what result actually is.
Element that has no size, is there one?
Is it possible to make the retrieving the size fail?

Also refactoring the code would be nice but I will post those comments later.

src/Selenium2Library/keywords/_element.py
```diff
@@ -296,7 +296,21 @@ def get_horizontal_position(self, locator):
...
-
+ 
+    def get_element_size(self, locator):
+        """Returns width and height of element identified by `locator`.
+            
+        The height abd height and width is returned in pixels,as an integer.
+        Fails if a matching element is not found.
+        
+        """
+        element = self._element_find(locator, True, False)
```
aaltat, line 307: Perhaps the required argument should be set to True. Then `_element_find` will raise ValueError with better error message. Also then lines 308 - 309 are not needed

src/Selenium2Library/keywords/_element.py
```diff
@@ -296,7 +296,21 @@ def get_horizontal_position(self, locator):
...
+        if element is None:
+           raise AssertionError("Could not determine size for '%s'" % (locator))
+        width = element.size['width']
+        height = element.size['height']
+        return (width, height)
```
aaltat, line 312: You could write lines 310 - 312 in one line: `return element.size['width'], element.size['height']`

src/Selenium2Library/keywords/_element.py
```diff
@@ -296,7 +296,21 @@ def get_horizontal_position(self, locator):
         if element is None:
             raise AssertionError("Could not determine position for '%s'" % (locator))
         return element.location['x']
-
+ 
+    def get_element_size(self, locator):
+        """Returns width and height of element identified by `locator`.
+            
+        The height abd height and width is returned in pixels,as an integer.
```
aaltat, line 303: Is is said somewhere that it is pixels and will always be integers? I did try to find it, but could not confirm from anywhere.

test/acceptance/keywords/elements.robot
```diff
@@ -69,3 +69,9 @@ Get Vertical Position
     Should Be True    ${pos} > ${0}
     Run Keyword And Expect Error    Could not determine position for 'non-existent'
     ...    Get Horizontal Position    non-existent
+
+Get Element Size
+	${width}  ${height}=  Get Element Size  link=Link
+	Should be True  ${height} > ${0}
+	Should be True  ${width} > ${0}
+	Run Keyword And Expect Error  Could not determine size for 'non-existent'  Get Element Size  non-existent
```
aaltat, line 77: 1 I think line 77 is not needed
2 Write test for where element has size put it is hidden. 

Example:

```
| Element Should Not Be Visible | locator=here |
| ${width} | ${height} = | Get Element Size | locator=here | 
| Verify | ${width} | ${height} | # No idea what value actually will be | 
```

3 Write test for element that should not have size. Example for a `<div></>` which does not contain anything
4 Add new line in the end.

If the test 2&3 brings any unexpected situations, it should be worth of documenting those issues.

src/Selenium2Library/keywords/_element.py
SergiuTudos, line 303: Did not hear about other values

aaltat, line 303: Because it is not defined anywhere. I would just say: 
`The element width and height is returned.`

PR #519, opened 2015-10-01
Add 2 table keywords

qitaos:
I think these 2 table keywords are useful for users.
- 'get table rows': Return the rows of the table.
- 'get table cols at row':Return the columns of the table in one row.

I add acceptance testcase for them.

There is only one problem: 
- If table has nested table in one row, 'get table cols at row' will count the nested table together on the row.


aaltat:
Purpose of your keywords is to count the rows or columns in a table? Is there a reason why you could not use [Get Matching Xpath Count](http://robotframework.org/Selenium2Library/doc/Selenium2Library.html#Get%20Matching%20Xpath%20Count) keyword?

qitaos:
@aaltat I feel these two very convenient for users.

just-be-dev:
This seems rather confusing to me. `Get Table Rows` tells me that I'm going to be retrieving the actual elements, but you're returning the number of elements. In this case it should be `Get Table Row Counts` or something along those lines, but even still, I'm not sure that this gives enough value to be added. 

If this was modified to do what it actually sounds like it should and returned the actual elements then it would likely be a good addition. You could use the returned elements to get the count.

aaltat:
Thank you from the update, but the previous comment from zephraph is still a valid one. Just counting the cells is doesn't bring value that would justify to merge the pull request.

When thinking about getting the index of row/col/cell the same can be done by using xpath count https://msdn.microsoft.com/en-us/library/ms256103(v=vs.110).aspx I agree that the xpath may come complex but normally playing with index is not a good practice.

Generally speaking handling tables can be complex thing. I also have written few keywords for tables and if I exclude the few cases where I don't care where the thing X is in the table, I have come to a conclusion that the the best approach for me is to read the table values in a list of list (also known as a table) from the page. And return that list of list to Robot Framework side, as a single variable, for verification. Thank way user can read the values one time from the table and use other Robot Framework keywords to perform the checks on the table data. In this way the user can perform check as many times they want, for a single read from the table and the checks can be as complex as they want.

aaltat:
1) This PR contains to many changes and therefore is not going to be merged in the master.

2) If I did read correctly, my previous previous comment is still valid.

Because of the above closing the PR.

PR #527, opened 2015-11-03
Allow locator strategy of xpath= for matching xpath keywords.

### LINKED ISSUE #526: Allow for specifying xpath locator strategy when using matching xpath keywords
We shouldn't penalize users for having the leading xpath locator strategy "xpath=" in the locator that is passed to the keywords `Get Matching Xpath Count` and `Xpath Should Match X Times`. At my company, for example, it is general practice to create variables for the locators specifying the strategy. Thus when using the matching xpath count keywords we have to make a separate variable. Instead we can simply accept either and give a meaningful error otherwise; something like

``` python
if locator.startswith('//')
    locator = 'xpath=' + locator

if locator.startswith('xpath=')
    # use locator
else
    # raise an menaingful error ... "locator should be xpath and start with xpath= or //"...
```

emanlove:
Fixes #526.

### DIFF
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -599,7 +599,7 @@ return !element.dispatchEvent(evt);
 
     # Public, xpath
 
-    def get_matching_xpath_count(self, xpath):
+    def get_matching_xpath_count(self, xpath_):
         """Returns number of elements matching `xpath`
 
         As of release 1.8 `Get Matching Xpath Count` allows the
@@ -612,20 +612,17 @@ return !element.dispatchEvent(evt);
         If you wish to assert the number of matching elements, use
         `Xpath Should Match X Times`.
         """
-        if xpath.startswith('//'):
-            xpath = "xpath=" + xpath
+        if not xpath_.lower().lstrip().startswith('xpath='):
+            xpath_ = "xpath=" + xpath_
         
-        if xpath.startswith('xpath='):
-            count = len(self._element_find(xpath, False, False))
-            return str(count)
-        else:
-            raise ValueError("Element locator must be xpath and start with xpath= or //")
+        count = len(self._element_find(xpath_, False, False))
+        return str(count)
 
-    def xpath_should_match_x_times(self, locator, expected_xpath_count, message='', loglevel='INFO'):
-        """Verifies that the page contains the given number of elements located by the given `locator`.
+    def xpath_should_match_x_times(self, xpath_, expected_xpath_count, message='', loglevel='INFO'):
+        """Verifies that the page contains the given number of elements located by the given `xpath_`.
 
         As of release 1.8 `Xpath Should Match X Times` allows the
-        argument `locator` to start with the xpath= locator strategy.
+        argument `xpath_` to start with the xpath= locator strategy.
 
         Example:
         | Xpath Should Match X Times | //div[@id='sales-pop'] | 1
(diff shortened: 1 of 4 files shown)


aaltat:
Should the _element.py line 624 `locator` argument be changed back to `xpath` ? Because the argument must still be xpath and it will not accept other supported locator strategies.

Regardless of the argument name the argument name should match in lines 602 and 624. Other than that, look's good to me.

src/Selenium2Library/keywords/_element.py
```diff
@@ -602,41 +602,54 @@ def page_should_not_contain_image(self, locator, message='', loglevel='INFO'):
...
+        
+        if xpath.startswith('xpath='):
+            count = len(self._element_find(xpath, False, False))
+            return str(count)
+        else:
+            raise ValueError("Element locator must be xpath and start with xpath= or //")
 
-    def xpath_should_match_x_times(self, xpath, expected_xpath_count, message='', loglevel='INFO'):
-        """Verifies that the page contains the given number of elements located by the given `xpath`.
+    def xpath_should_match_x_times(self, locator, expected_xpath_count, message='', loglevel='INFO'):
```
emanlove: I left the argument in `Get Matching Xpath Count` as __xpath__ but changed the argument in `Xpath Should Match X Times` to __locator__. Here I changed it so one wouldn't get an Robot Framework positional argument error. If, as in one of the acceptance tests, you wrote 

``` RobotFramework
  Xpath Should Match X Times xpath=//input[@type="text"] ${1}
```

and the argument was __xpath__, Robot Framework would think you were placing a named/optional argument, __xpath__, before a positional argument of __expected_xpath_count__.

They don't have to be the same. For `Get Matching Xpath Count` since there is only one argument, which I thought __xpath__ was more descriptive, I left it. To avoid the unintentional error I described above I changed the argument to __locator__ under `Xpath Should Match X Times`.

emanlove:
@aaltat Not sure about

> Should the _element.py line 624 locator argument be changed back to xpath ?

Line 624 doesn't match up to the code changes and doesn't contain xpath nor locator? What do you mean here?

src/Selenium2Library/keywords/_element.py
ombre42: If you want to avoid an issue with the named argument processing of RF, I would suggest calling the argument xpath_, rather than locator.
Adding a trailing underscore is the PEP 8 suggestion for when there is a conflict with Python Keywords.
BuiltIn's Sleep and Get Time keywords use a trailing underscore on the time argument.

src/Selenium2Library/keywords/_element.py
```diff
@@ -602,41 +602,54 @@ def page_should_not_contain_image(self, locator, message='', loglevel='INFO'):
...
 
         If you wish to assert the number of matching elements, use
         `Xpath Should Match X Times`.
         """
-        count = len(self._element_find("xpath=" + xpath, False, False))
-        return str(count)
+        if xpath.startswith('//'):
+            xpath = "xpath=" + xpath
+        
+        if xpath.startswith('xpath='):
```
ombre42: The detection of the xpath locator prefix is both case- and space-sensitive. This is not consistent with S2L.

src/Selenium2Library/keywords/_element.py
```diff
@@ -602,41 +602,54 @@ def page_should_not_contain_image(self, locator, message='', loglevel='INFO'):
...
         `loglevel` arguments.
         """
-        actual_xpath_count = len(self._element_find("xpath=" + xpath, False, False))
+        if locator.startswith('//'):
+            locator = "xpath=" + locator
+        
+        if locator.startswith('xpath='):
+            actual_xpath_count = len(self._element_find(locator, False, False))
+        else:
+            raise ValueError("Element locator must be xpath and start with xpath= or //")
```
ombre42: This is a backwards incompatible change. Previously, the XPath did not have to start with //. For example:
`${count}= | Get Matching Xpath Count | (//a)`
I doubt many tests out there use parentheses in this context, though.
Another possible breakage is leading space in the XPath. That is working now, but will fail after this change.

emanlove:
@aaltat Somehow I read that as line 642. Reworked as @ombre42 suggested.

src/Selenium2Library/keywords/_element.py
```diff
@@ -599,44 +599,51 @@ def page_should_not_contain_image(self, locator, message='', loglevel='INFO'):
...
+        Example:
         | Xpath Should Match X Times | //div[@id='sales-pop'] | 1
-        Incorrect:
         | Xpath Should Match X Times | xpath=//div[@id='sales-pop'] | 1
 
         See `Page Should Contain Element` for explanation about `message` and
         `loglevel` arguments.
         """
-        actual_xpath_count = len(self._element_find("xpath=" + xpath, False, False))
+        if not xpath_.lower().lstrip().startswith('xpath='):
```
ombre42, line 634: I think this is good enough. But wanted to point out that the prefixes as seen in UT test_find_with_sloppy_prefix wouldn't all work here. lstrip would not remove spaces in the prefix or before the equals sign. ElementFinder._parse_locator partitions on = and strips both parts. The prefix is not space-sensitive because it does a lookup in a dict that ignores space.
I think this is good enough but wanted to point out that its still not the same.

emanlove, line 634: Just a note that I think my code should be fixed as @ombre42 pointed out before you merge @aaltat. I should be able to look at it sometime this week or if anyone else does please feel free.

tminakov, line 634: Just to think we are that close to support locators with the strategy in, I'm getting sick of removing it from my locator vars to verify the correct number of elements are present :smile:  
All joking aside, what is needed to get the PR merged? This will be a really helpful improvement; thanks a lot in advance :+1:

aaltat:

aaltat:
This is actually done as part of #949 and therefore I am closing this PR. And actual implementation is done in #972

PR #528, opened 2015-11-04
Get Element Size.

SergiuTudos:


src/Selenium2Library/keywords/_element.py
```diff
@@ -297,6 +297,21 @@ def get_horizontal_position(self, locator):
...
+        """Returns width and height of element identified by `locator`.
+
+        The element width and height is returned.
+        Fails if a matching element is not found.
+        """
+        element = self._element_find(locator, True, False)
+        if element is None:
+            raise AssertionError("Element with locator '%s' was not found or is not visible" % (locator))
+
+        if element.size['width'] * element.size['height'] == 0:
```
aaltat: I am not sure, is it a good idea to raise an error if size is zero? Selenium seems to return zero size, if the element is hidden and I am not totally sure is it a good idea to change that idea. But raising an error follows Selenium2Library general idea. So we need discuss which way to go.

SergiuTudos: this is was a request I've been working on. previous version did not raise
any exception and pull request has been rejected.

emanlove: I don't believe I contributed to the earlier discussion about zero height/width but I am thinking the same as @aaltat that we don't want to error out on a zero width or height. Here, if the element is found, I would return the width and height and let the user interpret the result. I googled "can element have zero width or height" and it looks like there are legitimate reason for the result to be zero and even possible for either one to be zero and the other non zero thus failing on the

```
if width*height == 0 then error
```

check. Thus I would leave it up to the user to decide what is an error or not. Just my 2 cents...

SergiuTudos: I think I'll agree. will change the test and the code itself

SergiuTudos:
Code is ready. just accept it :)

src/Selenium2Library/keywords/_element.py
```diff
@@ -297,6 +297,18 @@ def get_horizontal_position(self, locator):
             raise AssertionError("Could not determine position for '%s'" % (locator))
         return element.location['x']
 
+    def get_element_size(self, locator):
+        """Returns width and height of element identified by `locator`.
+
+        The element width and height is returned.
+        Fails if a matching element is not found.
+        """
+        element = self._element_find(locator, True, False)
```
aaltat: It could be useful to put the `required` argument as `True`. In that way one does not need the lines 307 and 308. Also it quarentees unified error message.

src/Selenium2Library/keywords/_element.py
```diff
@@ -297,6 +297,18 @@ def get_horizontal_position(self, locator):
...
+        if element is None:
+            raise AssertionError("Element with locator '%s' was not found or is not visible" % (locator))
+
+        return (element.size['width'], element.size['height'])
```
aaltat: This returns tuple, which is not desirable. Write the line without parentheses.

aaltat:
We are getting quite close. Squashing all the separate commits to a single commit is desirable. Details can be example found from here: http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit

SergiuTudos:
BINGO

aaltat:
I think this is good to go.

src/Selenium2Library/keywords/_element.py
```diff
@@ -297,6 +297,16 @@ def get_horizontal_position(self, locator):
             raise AssertionError("Could not determine position for '%s'" % (locator))
         return element.location['x']
 
+    def get_element_size(self, locator):
+        """Returns width and height of element identified by `locator`.
+
+        The element width and height is returned.
+        Fails if a matching element is not found.
```
ombre42, line 304: Would be nice if there was an example like with [Get Window Size](http://robotframework.org/Selenium2Library/doc/Selenium2Library.html#Get%20Window%20Size).

pekkaklarck, line 304: Agreed.

src/Selenium2Library/keywords/_element.py
```diff
@@ -297,6 +297,16 @@ def get_horizontal_position(self, locator):
...
+        """
+        element = self._element_find(locator, True, True)
```
pekkaklarck, line 306: Passing `True` or `False` as a positional argument is always a bit confusing. Would be better to name the arguments like `something=True`.

src/Selenium2Library/keywords/_element.py
```diff
@@ -297,6 +297,16 @@ def get_horizontal_position(self, locator):
...
+
+        return element.size['width'], element.size['height']
```
pekkaklarck, line 308: It might be a good idea to return a [namedtuple](https://docs.python.org/2/library/collections.html#collections.namedtuple) instead of a normal tuple. It would allow this kind of usage:

```
${size} =    Get Element Size    ${locator}
Should Be Equal    ${size.height}    42
Should Be Equal    ${size.width}    0
```

Also, the empty line above the `return` statement is unnecessary.

PR #535, opened 2015-11-17
Adds ignoring "--dryrun" to rebot to allow getting the report and log.

HelioGuilherme66:
Changes a flag to True to allow running unit tests in Fedora 22.

### DIFF
--- a/test/run_tests.py
+++ b/test/run_tests.py
@@ -67,10 +67,7 @@ def process_output(args):
         call(['python', os.path.join(env.RESOURCES_DIR, 'statuschecker.py'),
              os.path.join(env.RESULTS_DIR, 'output.xml')])
     rebot = 'rebot' if os.sep == '/' else 'rebot.bat'
-    try:
-        args.remove('--dryrun')
-    except TypeError:
-        pass
+    if '--dryrun' in args: args.remove('--dryrun')
     rebot_cmd = [rebot] + [ arg % ARG_VALUES for arg in REBOT_ARGS ] + args + \
                 [os.path.join(ARG_VALUES['outdir'], 'output.xml') ]
     print('')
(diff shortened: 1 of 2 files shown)


test/run_tests.py
```diff
@@ -67,6 +67,10 @@ def process_output(args):
         call(['python', os.path.join(env.RESOURCES_DIR, 'statuschecker.py'),
              os.path.join(env.RESULTS_DIR, 'output.xml')])
     rebot = 'rebot' if os.sep == '/' else 'rebot.bat'
+    try:
+        args.remove('--dryrun')
+    except TypeError:
+        pass
```
pekkaklarck: First of all, `list.remove` raises `ValueError`, not `TypeError`, if the removed item doesn't exist. Secondly, it would be shorted and cleaner to use:

``` python
if '--dryrun' in args:
    args.remote('--dryrun')
```

HelioGuilherme66:
Please consider this small change of code in S2L acceptance test cases. I would like to run the tests with `--dryrun` and other combinations of robot parameters, and the verify the log and report.

aaltat:
Is this related to the #536 ?

HelioGuilherme66:
Yes it is. I did not remember to have the issue. But the conversation drifted into RF. This is a change only to S2L. (of course I can have this only in my fork, but it would be harmless if on master)

aaltat:
Should we open the original issue or a new one and could you explain why this is needed. And what is the problem you are trying to solve?

aaltat:
Can we close this issue?

HelioGuilherme66:
Merged with master and is ready to be merged.

pekkaklarck:
You should not merge changes from master to your PRs unless changes actually affect your code. Merging creates totally unnecessary commit history.

pekkaklarck:
This would fit into the scope of PR #710, but I fail to see the benefits of the change. There are lot of tests that contain log level validations done by StatusChecker. All those would fail in the dry-run mode.

Could you @HelioGuilherme66 clarify your use cases for `--dryrun`? In general it would be better to first open an issue where we can discuss does a change make sense in the first place.

HelioGuilherme66:
I wanted to have --dryrun working to confirm that new acceptance tests were OK. I also would like to have the Report and Log generated for other experiments.
This PR is going to be closed, because argument parsing logic was changed.

PR #538, opened 2015-11-19
Let's try to get only one element screenshot using PIL

SergiuTudos:
I think we can add a new dependency to the project and introduce new ability for this library to do some sort of visual regression testing. In order to do that a method GET ELEMENT SCREENSHOT is required. Other Image comparison methods will be added later.

### DIFF
--- a/src/Selenium2Library/keywords/_screenshot.py
+++ b/src/Selenium2Library/keywords/_screenshot.py
@@ -1,5 +1,6 @@
 import robot
 import os, errno
+from PIL import Image
 
 from Selenium2Library import utils
 from keywordgroup import KeywordGroup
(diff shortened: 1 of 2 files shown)


boakley:
This is a cool idea. However, I think a dependency on PIL is a bad idea.
PIL development seems to be dead. The most recent version is from 2009.

Pillow seems like the better choice, but even then, that's a relatively
heavyweight dependency for something only a few people will likely use. It
seems to me that "visual regression testing" should be a separate library.
Perhaps it could inherit from Selenium2Library so you wouldn't need to
include both.

HelioGuilherme66:
Hi,

I also like this feature. It was in my TODO list.
I do think I have seen this function in the Selenium library.

Cheers,
Hélio

aaltat:
There are tools that does the image comparison out of the box and you can use your existing S2L for driving the test. Example https://applitools.com and there are plenty of other tools available, open source and commercial.

I agree with @boakly that it should be a separate library, which can be used along with S2L.

SergiuTudos:
Applitools is not free. That's why I want to do quite the same in an open
source library

aaltat:
Have you example looked at: https://github.com/xebia/VisualReview

emanlove:
There is already a whole library that does annotated,cropped and enhance screens hots including focusing on a single element.

See https://github.com/datakurre/robotframework-selenium2screenshots

emanlove:
The image comparison portion is actually really difficult to do as there is a lot of different things one must consider if to make the tests stable. At the recent Selenium Conference, Adam Carmi from Appitools [gave a talk](https://www.youtube.com/watch?v=CHUuLdkFfm0) about visual testing. He gives a good overview of all the image processing problem one encounters while trying to do this.

One of the tools Adam lists is [Huxley](https://github.com/facebookarchive/huxley) which is no longer supported. I don't know why Facebook/Instagram abandoned it but image comparison is really out of scope for this library. And let's not reinvent the wheel with advanced screenshot functionality and reuse and enhance [Asko's robotframework-selenium2screenshots library](https://github.com/datakurre/robotframework-selenium2screenshots).

[I am closing this pull request as per the arguments given. if you feel strongly that I am wrong go ahead an reopen this. I just don't want someone to accidentally click merge]

PR #549, opened 2015-12-11
Changed docstrings

GLMeece:
Changed documentation for set_window_position and get_window_position.

Note - after my commit, I noticed I somehow wiped the grave accents (`) on the`x`and`y` in the docstring. That should be fixed before the final commit.

### DIFF
--- a/src/Selenium2Library/keywords/_browsermanagement.py
+++ b/src/Selenium2Library/keywords/_browsermanagement.py
@@ -260,7 +260,7 @@ class _BrowserManagementKeywords(KeywordGroup):
         return self._current_browser().set_window_size(width, height)
 
     def get_window_position(self):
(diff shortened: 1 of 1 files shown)


src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -269,13 +269,13 @@ def get_window_position(self):
         return position['x'], position['y']
 
     def set_window_position(self, x, y):
-        """Sets the position `x` and `y` of the current window to the specified values.
+        """Sets the position x and y of the current window (relative to the left and top of the screen) to the specified values.
```
aaltat, line 272: Because x and y are arguments, could you write it like this: `x` and `y`

Issue #479, opened 2015-08-05
Python 3 support


davidjb:
Are there any plans for this library to look at supporting Python 3?  The biggest blocker appears to be the fact that robotframework itself isn't yet Python 3.x compatible (though there is a Py3 fork available on PyPI; https://github.com/robotframework/robotframework/issues/1506).

For this package, though, it looks as though there'd need to some changes around the packaging (ez_setup.py), and minor changes to the syntax (exception handling, printing, relative imports), and the big one -- having a conditional statement in `setup.py` to choose between `robotframework` and `robotframework-python3`; all other dependencies appear Python 3 compatible.  With only those changes required, supporting 2.7 and 3.x seems reasonably straightforward, though I'm not intimately familiar with the code base.

Thoughts?

emanlove:
Feel free to submit a pull request. My only concern is that we maintain support for Python 2.7 (and possibly 2.6).

pekkaklarck:
Python 3 support would be great. I agree with @emanlove that Python 2.7 must be supported also in the future. Python 2.6 is not too much extra work on top of that (as long as dependencies support it) but because it is [officially retired](https://www.python.org/download/releases/2.6.9/) I'm not sure is that worth any effort. We haven't yet decided are we going to support it with Robot Framework itself in RF 3.0 anymore.

Another decision to make is what is minimum Python 3 version to support. I'm thinking about Python 3.3 because it supports `u'foo'` syntax (PEP-414) that makes it easier to write Python 2 compatible code.

aaltat:
I guess we should somehow discover the dependencies which each external RF library requires. For Selenium2Library, we depend on decorator, selenium, docutils and RF itself. 

The only external library that does not clearly say it supported 3.X version is docutils, but it claims to be Python 3 compatible. Decorator has best support for Python version, from 2.6 to all the way to 3.5. Selenium also nice support Python 2.6, 2.7, 3.2 and 3.3. 

But I guess that the bottom line is that change for Python 3.X is coming and the RF core them should lead the change and the rest of us will/should follow. So choose wisely. But perhaps this would be place for a wider audience and should be discussed with the whole user group people. 

Personally I have not coded with Python 3.X or read anything that would allow me to form an opinion to any direction from Python 3.X. 

But @zephraph, I think that support for Python 3 requires a major version change, should we change the label to 2.0?

pekkaklarck:
[...]

PR #556, opened 2015-12-24
Makes Selenium2Library compatible with both Python 3.4 and 2.7

HelioGuilherme66:
Tested in Linux x64.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,21 @@
 language: python
 sudo: false
 python:
-  - "2.6"
   - "2.7"
-install:
+  - "3.4"
+before_install:
+  - "./.travis_python_dep.sh"
+install: true
   #- pip install pep8 --use-mirrors
   #- pip install pyflakes --use-mirrors
-  - pip install robotframework
-  - pip install selenium
-  - pip install decorator
-  - pip install docutils
-  - pip install future
-  - pip install mockito
+  # Installation is done at before_install
+  ##- pip install robotframework==3.0b1
+  ##- pip install selenium
+  ##- pip install decorator
+  ##- pip install docutils
+  ##- pip install future
+  # Mockito is only used when in Python 2.7
+  ##- pip install mockito
   #- "if [[ ${BROWSER} == 'ie' ]]; then pip install ie_only_pkg; fi"
 before_script:
   - "export DISPLAY=:99.0"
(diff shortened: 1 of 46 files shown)


pekkaklarck:
If the idea is to keep the code base Python 2 compatible, using 2to3 is not a good idea. Converting code manually and using `future` or `six` as the compatibility layer works much better. For more details see e.g. http://python-future.org/. 

In my opinion we should first decide together what approach to use when porting. That covers deciding what compatibility layer to use (`future`, `six`, create own, Robot's `utils`) and do we want the code to look more like Python 2 or Python 3.

HelioGuilherme66:
I did use `future` after analysis of Robot Framework code. The 2to3 converter surprised me for its efficiency, and I think is a very good starting point. I go now see why Travis is failing.

HelioGuilherme66:
I fixed the problem. There is a launcher for testserver depending if we are on Python 2 or 3. (forgot to add the files). Now the Travis failing should be related with timeout when starting server (locally runs fine).

pekkaklarck:
I totally disagree about 2to3. It has done lot of stupid changes to the code. I'm strongly -1 on using it. Much better do the conversion manually from the start than trying to fix mess 2to3 has created. Manually going through the code also makes it possible to enhance code otherwise. All that is going to help with future maintenance.

aaltat:
Because many other projects have abandoned 2to3, for good reasons, I think we should not repeat their mistakes. Also I agree that code does not look good.

I have not read Robot `utils` code, but my first gut feeling would be to go with `future`. But because it's just a gut feeling, I don't feel strongly about it.

Because Python 3 is the future (on could even argue that it's a present), I think code must look like Python 3.

pekkaklarck:
A big problem with using `robot.utils` as Python 2/3 compatibility layer is that it requires RF 3.0. I agree that `future` or a little more lower level `six` is a good choise.

aaltat:
I think we should update the #479, when we agree on something. Also setting up CI to test against Python 3 would be good too. And dropping 2.6 would be nice too, in my personal opinion. I think we should have some sort of consensus, how we are going to do this.

pekkaklarck:
Dropping Python 2.6 support is a good idea. It cost us way too many hours with RF 3.0.  RF 3.1 will drop it.

HelioGuilherme66:
Closing in favor of #564 (better commits/PR changes).

PR #561, opened 2016-01-01
Code clean up print

HelioGuilherme66:


pekkaklarck:
It seems to me that most of the touched files should actually be removed altogether. I don't think separate packaging scripts etc. are needed at all. If some of the scripts actually are needed, it is questionable do they need to support Python 3 at least immediately.

pekkaklarck:
Other PRs have already fixed prints and/or removed unnecessary files altogether.

PR #562, opened 2016-01-01
Makes Selenium2Library compatible with Robotframework 3.0

HelioGuilherme66:
It stays compatible with RF 2.9.2.


aaltat:
I think this is not anymore valid, @HelioGuilherme66 should we close the PR?

HelioGuilherme66:
Closing as it is no longer necessary.

PR #563, opened 2016-01-01
Code clean up except

HelioGuilherme66:

PR #559, opened 2015-12-31
Changes to Travis settings to allow installation dependent on Python …

HelioGuilherme66:
…versions.

Already has some settings for Python 3.4, but disabled.

PR #564, opened 2016-01-02
Makes Selenium2Library compatible with RobotFramework 3.0 and Python 3.4, 2.7 and 2.6

HelioGuilherme66:
This PR replaces #556 and contains PRs #559, #561, #562 and #563. Please merge them in order and look at commits to see changes grouped by code modification (not always ;).


doc/generate.py
```diff
@@ -1,14 +1,17 @@
...
+
+from __future__ import print_function
 from os.path import join, dirname
+
 try:
     from robot.libdoc import libdoc
 except:
     def main():
-        print """Robot Framework 2.7 or later required for generating documentation"""
+        print('"""Robot Framework 2.7 or later required for generating documentation"""')
```
just-be-dev: Let's remove all the extraneous quotes here.

just-be-dev:
Have you looked through the 3.4 errors? There seems to be a lot of them... Lots of elements couldn't be found.

doc/generate.py
HelioGuilherme66: I left the quotes there because I think that it is intended to print a comment.

HelioGuilherme66:
In my last commit  (8e95b3f) I changed the way how python is called, by using the version suffix. There are problems with Python3.4 ../http/server.py code. That is why the Travis is failing with Python3.4. You may verify that the demo starts the server correctly (and also other tests I have).
This is the only blocking point for us to have a green Travis build in Python3.4.

pekkaklarck:
I'm sorry but this PR is so big and touches so many places that I don't have time to review and comment it in detail. Some general comments:
- In my opinion scripts related to packaging, demo, etc. should not be touched at this point. It seems to me most of them are not used anymore and should just be removed. If some are still needed, they don't urgently need Python 3 support.
- Travis configuration changes should get their own PR. I don't know too much about Travis and canot merge PRs touching it without knowing what change do.
- As I have commented earlier, I would like to see separate PRs fixing individual Python 3 compatibility issues without affecting Python 2 compatibility. Two easy topics to start are:
  - Exceptions: `except Exception, err` -> `except Exception as err`.
  - Imports: `from internal import Stuff` -> `from .internal import Stuff`. This PR could also contain general cleanup to imports to make them more PEP-8 compatible. It shouldn't contain any other code changes.

aaltat:
I think Pekka comment is still a valid one. And also the commit history is awful.

HelioGuilherme66:
This PR is the accumulated commits. We should analyze them in order, starting with PR #559, then after that one is merged, we go for #561, and so on. The latest commits in this one, could be discarded as they are essentially merges from master.

pekkaklarck:
Most of the changes done here is already done by other PRs. This was too big PR with too many changes. to begin with.

PR #568, opened 2016-01-11
Edge browser compatibility

jfx:
Add browser Edge in Open Browser keyword

### DIFF
--- a/src/Selenium2Library/keywords/_browsermanagement.py
+++ b/src/Selenium2Library/keywords/_browsermanagement.py
@@ -23,7 +23,8 @@ BROWSER_NAMES = {'ff': "_make_ff",
                  'htmlunitwithjs' : "_make_htmlunitwithjs",
                  'android': "_make_android",
                  'iphone': "_make_iphone",
-                 'safari': "_make_safari"
+                 'safari': "_make_safari",
+                 'edge': "_make_edge"
                 }
 
 class _BrowserManagementKeywords(KeywordGroup):
@@ -85,6 +86,7 @@ class _BrowserManagementKeywords(KeywordGroup):
         | android          | Android       |
         | iphone           | Iphone        |
         | safari           | Safari        |
+        | edge             | Edge          |
 
 
         Note, that you will encounter strange behavior, if you open
@@ -157,7 +159,7 @@ class _BrowserManagementKeywords(KeywordGroup):
         | Create Webdriver            | PhantomJS    | service_args=${service args}              |                         |
 
         Example for Robot Framework < 2.8:
-        | # debug IE driver |                   |                  |                                | 
+        | # debug IE driver |                   |                  |                                |
         | ${kwargs}=        | Create Dictionary | log_level=DEBUG  | log_file=%{HOMEPATH}${/}ie.log |
         | Create Webdriver  | Ie                | kwargs=${kwargs} |                                |
         """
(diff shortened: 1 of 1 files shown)


aaltat:
Looks good for me. 

There is one thing to discus, usually we have been very careful to raise the minimum required selenium version, but I can not anymore recall the reason for it. @emanlove, or someone, do you remember why? I did check the [CHANGES](https://github.com/SeleniumHQ/selenium/blob/master/py/CHANGES) for selenium and Edge support was made in 2.47.0, would this justify to raise the minimum selenium version?

emanlove:
In my opinion, No, it would not justify raising the minimum selenium version. The reason for not raising the minimum, in general, is to not force people to install only the latest. Not knowing the individual and particular circumstances our end users are in, it gives them the freedom to self choice what works for them. There could be very legitimate reason unbeknownst to us for setting the version at which they choose.

A good criteria or reason for bumping the minimum version of selenium would be if the change in our library without a change in the minimum selenium version would cause an import error for all users. Yes, if one wants to use Edge then one needs some minimum version of selenium but not everyone needs or uses Edge. One might even say its an edge case. (Sorry, couldn't resist). I would much rather see some code that checks selenium version or existence of webdriver.edge and errors out if not there then to raise the minimum version.

emanlove:
Also you should not make changes directly to doc/Selenium2Library.html. Where I think you want to make to the individual docstrings which in this case would be under the `[Create Webdriver](https://github.com/robotframework/Selenium2Library/blob/master/src/Selenium2Library/keywords/_browsermanagement.py#L126)`, `[Open Browser](https://github.com/robotframework/Selenium2Library/blob/master/src/Selenium2Library/keywords/_browsermanagement.py#L59)`, and possibly the `[Switch Browser](https://github.com/robotframework/Selenium2Library/blob/master/src/Selenium2Library/keywords/_browsermanagement.py#L180)` keywords. This needs to be corrected before this pull request is reconsidered.

drobota:
I've already added edge in #1388

drobota:
I have added Edge  in https://github.com/robotframework/Selenium2Library/pull/552

jfx:
Hello,
No change was made directly in doc/Selenium2Library.html but was generated when I run build_dist.py to test it. Anyway I revert it. I revert also the minimum version of Selenium and check it in _browsermanagement.py.

SteveHarrison82:
dear all,

if edge is available in master, it would be of great help. Thank you all

divfor:
I tested it and found it cannot be closed by 'Close All Browsers'. should add unit test to avoid this.

ssaurabh61:
Hi, Is it possible to get Edge browser available on master. It would help us a lot.

drobota:
@ssaurabh61 as a work around you can install package from any branch that you need
see http://stackoverflow.com/questions/20101834/pip-install-from-github-repo-branch
Moreover if you use selenium grid or services like saucelabs, browserstack etc. you can overwrite browserName original capability using `desired_capabilities` parameter.
What is more I added EDGE support in another pr https://github.com/robotframework/Selenium2Library/pull/552
I tested it and it works well.

aaltat:
Sorry for a really long delay. Would you still be interested to get this merged? If you are, would it be possible you to [squash commits](https://github.com/robotframework/Selenium2Library/blob/master/CONTRIBUTING.md#squashing-commits) and then this could be ready to go.

jfx:
Ok I shall do this at the end of week.

aaltat:
Thank you. We are planning to make minor release soon and this fits nicely in.

jfx:
Commits are squashed

aaltat:
Thank you, I will try to make the release next week. But now I concentrate on FRA - POR football game.

Issue #531, opened 2015-11-09
New keyword Element Attribute Should Be


manycoding:
I need to check user created attributes and it'd nice if I have this keyword:
    Element Attribute Should Be    //input[contains(@id, "id")]@aria-disabled     ${data.prerequisites}

Instead of:
    ${value}=    Get Element Attribute    //input[contains(@id, "id")]@aria-disabled
    Should Be Equal    ${value}    ${data.prerequisites}

aaltat:
I think this could be useful. I think that one should define the attribute which they are checking, example something like:
`|Element Attribute Should Be | xpath=//input | key=value |`

As a workaround one can do this with `Page Should Contain` keyword and xpath. Example:
`| Page Should Contain Element | xpath=//input[@id="I'd" and @aria-disabled="True"] |` But having own keyword for this makes it more clearer what the keyword is actually testing, because xpath can come quite complex.

Would you like to make a pull request to implement this?

manycoding:
Forgot to mention the attribute itself.
As for the pull request - I'll think about it.

emanlove:
Just a word of encouragement to you @manycoding as it is always great to get more contributors. If you need any help just ask.

drobota:
https://github.com/robotframework/Selenium2Library/pull/570

php-coder:
> https://github.com/robotframework/Selenium2Library/pull/570

Here is the working link: #570

aaltat:
That is because project changed name and direct links with old name got broken.

PR #570, opened 2016-01-20
Add keyword as requested in #531

drobota:


drobota:
Does anybody can see build log?

HelioGuilherme66:
I can see the output (don't know how to access logs).
There are two tests failing. Are they passing in your machine?

```
Element Attribute Should Be                                           | PASS |
Element Attribute Should Be Should Raise Error If Attribute Incorrect | FAIL |
Keyword 'BuiltIn.Run Keyword And Expect Error' expected at least 2 arguments, got 1.
Element Attribute Should Be Should Raise Error If Attribute Value ... | FAIL |
Keyword 'BuiltIn.Run Keyword And Expect Error' expected at least 2 arguments, got 1.
Acceptance.Keywords.Elements :: Tests elements                        | FAIL |
13 critical tests, 11 passed, 2 failed
13 tests total, 11 passed, 2 failed
```

drobota:
@HelioGuilherme66 Thanks for results, I have no access to logs too. I will try to run them

divfor:
I suggest the args should be "locator@attr_name" and "attr_value", which keeps similar usage with existing keyword "Get Element Attribute":
Source:
Selenium2Library <test library>
Arguments:
[ attribute_locator ]
Return value of element attribute. 
`attribute_locator` consists of element locator followed by an @ sign and attribute name, for example "element_id@class".

aaltat:
One thing came to mind, should we use kwargs for the attribute? It creates dependency for newer version of Robot Framework, but if I recall correctly there is other keyword that has similar dependency. @HelioGuilherme66 was it one of the keywords you did make? But in any case I think it's OK, but needs to be properly documented. 

By using kwargs, one could check multiple attribute values in same time. Something like this:

```
| Element Attribute Should Be | xpath=//path/here | key1=value1 | key2=value2|
```

src/Selenium2Library/keywords/_element.py
```diff
@@ -680,6 +679,31 @@ def remove_location_strategy(self, strategy_name):
...
         self._element_finder.unregister(strategy_name)
 
+    def element_attribute_should_be(self, locator, attr):
+        """Checks whether element has attribute with value
+
+        Usage Example:
+        | Element Attribute Should Be | xpath=//input | key=value |
+        """
+        self._info("Checking attributes '%s' for element '%s'" % (
+                                                    attr, locator))
```
aaltat: Empty lines are not needed, delete them all places in new code.

drobota:
@aaltat I will update keyword to make compatible to work with more the one pair of attributes

src/Selenium2Library/keywords/_element.py
```diff
@@ -680,6 +679,31 @@ def remove_location_strategy(self, strategy_name):
...
+        try:
+            key, value = attr.split("=", 1)
+        except ValueError:
+            raise AssertionError(
+                    "Attribute %s does not match pattern key=value" % attr)
+
+        el = self._element_find(locator, True, True)
+        retval = el.get_attribute(key)
+
+        if retval is None:
```
aaltat: This needs thinking, is different error message required if attribute value is empty. Keyword will fail also in 703 if this if is removed. But it prevents to check is attribute value empty. Did you have some use case for this?

drobota: It is just extra error that element does not have attribute. I wanted prevent case like 
| Element Attribute Should Be |  css=div |  text=${None}  |
bc when element does not have attribute get_attribute returns None.
To check whether attribute is empty it should be 
| Element Attribute Should Be |  css=div |  text=${EMPTY}  |
and for checking attributes like disabled and autofocus it should be 
| Element Attribute Should Be |  css=input |  autofocus=true |

aaltat: The ${EMPTY} and ${None} are, at least for me, the same thing, perhaps different sides of the same coin. For me they both mean the same thing, nothing. Also I would not like to limit the users to use specific kind of variable, ${EMPTY}, just to verif  that attribute don't have value.

Also from the code maintainability point of view, doing `if retval is None:` causes a problem because it's against [idiomatic Python truth value testing](http://python.net/~goodger/projects/pycon/2007/idiomatic/handout.html#testing-for-truth-values). 

The later is not that important if there is a good usecase for preventing using `key=${None}` to check that attribute don't have value. Could you define a use case where it's beneficial to prevent the None?

ombre42: I don't see a problem with testing for retval is None. That is the part of the contract of the `get_attribute` method - see [API docs](http://seleniumhq.github.io/selenium/docs/api/py/webdriver_remote/selenium.webdriver.remote.webelement.html#selenium.webdriver.remote.webelement.WebElement.get_attribute).
If you called this keyword like this:
`| Element Attribute Should Be | css=div | text=${None} |`
Then the method would receive u'text=None' as the value for `attr` and only pass if the attribute value was 'None'. RF would serialize ${None} into the string 'None'. If the key and value were separate arguments or as keyword arguments, then None could be passed as literally ${None}, but that is not how it was implemented.

src/Selenium2Library/keywords/_element.py
```diff
@@ -680,6 +679,31 @@ def remove_location_strategy(self, strategy_name):
...
+            raise AssertionError(
+                    "Element %s does not have attribute %s" % (locator, key))
+
+        if retval != value:
```
ombre42: Just an FYI. Per the API docs, if the value is 'true' or 'false', the return value of `get_attribute` is True or False (bool). But **the docs are wrong**. So a simple equality test here is fine.

aaltat:
Closing die inactivity.

PR #578, opened 2016-02-17
New keyworkd(Wait for Select Window) is added.

CVE-Defender:
Waits for a popup window to appear and load up
The difference with "Select Window" is "timeout" parameter.

hk.oh@piolink.com

### DIFF
--- a/src/Selenium2Library/keywords/_browsermanagement.py
+++ b/src/Selenium2Library/keywords/_browsermanagement.py
@@ -1,4 +1,5 @@
 import os
+import time
 import robot
 from robot.errors import DataError
 from selenium import webdriver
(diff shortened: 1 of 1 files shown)


aaltat:
Sorry for a long delay, but are you still interested to get the PR merged into the master?

If you are, could you also raise an issue. In the issue, could you describe the problem which you want to solve. Examples in Robot Framework data are usually really useful. I don't clearly see the problem which you are trying to solve and some clarification is needed, for me, to understand the problem. Also issue helps us to track stuff better.

Also the PR must be updated so that it doesn't have conflicts with the master. Also follow pep008 is a good guideline. Documentation for the keyword is required and writing tests for the keyword is a must. But all the PR stuff can wait for the issue and it's description.

aaltat:
Closing because of inactivity

PR #583, opened 2016-03-03
modify to locate "content-desc<0x20>"  element appropriately

lucyking:
modify to locate some element with blank postfix:
`... accessibility_id="value    "`


lucyking:
if no `"` wrap,
`criteria = locator_parts[2].strip()` will strip the blank at the end of locator_parts[2].
eg:
pic0
pic
the criteria's value is "我的优惠券" after parse,that can not locate the "我的优惠券 "  in practice.

aaltat:
Is there issue opened for this PR? If there is not could you open one and explain what is the problem you are trying to solve?

lucyking:
@aaltat done.

Issue #129, opened 2012-10-08
TextFields with "onChange" js-event needs to be filled twice to values be available


Hi-Fi:
Environment: Wicket-based form, which validates text inputs
Problem: When "Input text" is ran first time, it empties the field (and value attribute), and causes event to fire (at this case submitting the form). When (in same command from Robot Framework) command tries to add text, it can't find element anymore. 

Workaround is to make Input text first inside "Run keyword and ignore error", but this just seems quite stupid way (plus it generates extra screenshot/page source in log).

Second time the event isn't fired, which causes need for following action to be performed twice (first launches the event, and second time puts the correct value to dropdown's selected).

Hi-Fi:
For some reason Selenium's "clear" fires onChange event, and prevents running the Keyword 'til end. 

Workaround in Selenium2library is to find the element again after clear.

j1z0:
when you clear an element the onChange event will occur.  That is correct.  This doesn't prevent the next statement from happening unless you application specifically handles onChange and then changes the focus or something of the element.

However I'm not sure your particular use case is best handled in Selenium2Library as for the vast majority of people onChange event doesn't shift focus / submit form or whatever.

I think Juho you may have better luck wrapping Selenium2Library and creating your own custom library on top of it so you can handle the specific logic of your application.

Cheers,
Jeremy

Hi-Fi:
I actually have my own version of library in use. Just thought that this might be a thing that someone else will also struggle with, if using Wicket and filtering with onChange making form submit.

aaltat:
There has not been activity for this issue for few years and we are planning to close the old issues if they are not anymore valid ones. If you feel that this is still valid issue, could you send an update. If update is not received in a month, we will assume that this is not anymore an issues and we will close this issue.

Hi-Fi:
I'll try to create a test page and check this. At least code looks like same than that time.

aaltat:
Could you also share information how did you handle the problem in your own library?

Hi-Fi:
Created test page: http://fiddle.jshell.net/xsxd0p57/1/show/

Normally onChange() is launched only when I move out from field (by clicking something/pressing tab/filling other field). But if I put test 2 times to one field, onChange() prints empty result.

RF test case:

```
*** Settings ***
Library           Selenium2Library

*** Test Cases ***
OnChange Test
    Open Browser    http://fiddle.jshell.net/xsxd0p57/1/show/light/    ${BROWSER}
[...]

PR #590, opened 2016-04-01
#129 On_change issue

ghost:
#129 On_change was called twice, once when the field was cleared, and another with the new value.

Now the previous value is not cleared but replaced for the new one, so the on_change is called just once with the last value


aaltat:
This more a problem with the selenium than S2L and therefore closing the PR.

Issue #592, opened 2016-04-15
Deprecation warning from Selenium when using `Select/Unselect Frame`


ghost:
Select Frame / Unselect Frame keywords are using deprecated switch_to_frame / switch_to_default_content

actual warning message:
/usr/lib/python2.6/site-packages/selenium/webdriver/remote/webdriver.py:536: DeprecationWarning: use driver.switch_to.frame instead
/usr/lib/python2.6/site-packages/selenium/webdriver/remote/webdriver.py:542: DeprecationWarning: use driver.switch_to.default_content instead

proposed fix:
change in keyword Select Frame 
self._current_browser().switch_to_frame(element)  into  self._current_browser().switch_to.frame(element) 

change in keyword Unselect Frame 
self._current_browser().switch_to_default_content()
self._current_browser().switch_to.default_content()

ghost:
there are some more locations which are using switch_to :
_element.py:        browser.switch_to_frame(element)
_element.py:        browser.switch_to_default_content()
_element.py:        browser.switch_to_default_content()
_element.py:            browser.switch_to_frame(frame)
_element.py:            browser.switch_to_default_content()
_formelement.py:            alert = self._current_browser().switch_to_alert()
_javascript.py:            alert = self._current_browser().switch_to_alert()
_javascript.py:            alert = self._current_browser().switch_to_alert()

as indicated fix is to replace underscore (_) with dot (.)   after the text switch_to

ghost:
Fix deprecation warning switch_to_frame #599

PR #593, opened 2016-04-16
 fix for deprecation warning for select_frame / unselect_frame #592 

ghost:
fix for deprecation warning for select_frame / unselect_frame #592


ghost:
Apparently the current master branch has 2 errors and 3 warnings for unrelated tests ( these test also fail without this fix / pull request )

aaltat:
Thank you from the pull request. The S2L minimum required selenium version is quite low and could you check if it needs raising? If needs to be raised we need provide some warning to the users and perhaps decide when this would be best to be done.

ghost:
The deprecation change for selenium is in commit " lukeis committed on 19 Mar 2014 "
https://github.com/SeleniumHQ/selenium/commit/8fc478b09ea3c76800d09ab6681f9c3895b1bbc4

the first selenium release after that commit is 2.41.0 ( 23 march 2014 ) so that will be the version to use ( instead of 2.32.0 ).  I assume in the meantime most of us have upgraded ( it;s 2 years ago ).

aaltat:
@emanlove what you think, should we raise the minimum required of selenium? Or should we make a try\cathc?

ghost:
I would like to redo this pull request on a different branch ( not on ekasteel:master )

emanlove:
I'll look to see if there is another pull request as @ekasteel noted, but I would prefer a try-catch block over updating the selenium version. In addition we could provide a warning about eventual need to upgrade.

ghost:
the new pull request is #599...

PR #597, opened 2016-04-29
Fixes `Unselect From List` due to Selenium changes.

HelioGuilherme66:
Since Selenium version 2.53.0 deselect_by_value/label returns boolean, and
Exception if element is not found.
This fix keeps compatibility with versions <=2.52.0.


divfor:
@emanlove could you merge this PR to master? Otherwise all other PRs will get "The Travis CI build failed"

just-be-dev:
I haven't been active in a while, but this change looks good to me. Tests are indeed passing again, so that's swell. I'm going to merge.

PR #599, opened 2016-05-04
Fix deprecation warning switch_to_frame

ghost:
Fix deprecation warning switch_to_frame
since selenium release 2.41.0 this should be switch_to.frame

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ setup(name         = 'robotframework-selenium2library',
                      ],
       install_requires = [
 							'decorator >= 3.3.2',
-							'selenium >= 2.32.0',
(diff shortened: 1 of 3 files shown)


aaltat:
Sorry for a long delay in the answer but are you interested to get this to the master? If you are this requires some thoughts.

1) Would you happen to remember, in which version the API was changed?
2) We support selenium starting from 2.32.0 and if the API was changed after version, we should not directly use the new API, instead we should do try/expect block and expect proper error.

aaltat:

ghost:
no, the changes are minor and can be redone after the architecture changes

ghost:
Close pull request (not merged)

Issue #609, opened 2016-06-07
Add a new keyword for uploading multple files


VitoAlbano:
At the moment the input element with "multple" attribute cannot be tested with the keyword "Choose File".

I've created a custom keyword which takes in input a list of files and works on Chrome ( needs to be test on Firefox).

aaltat:
I haven't encountered this type of element before. Would you have some example webpage somewhere which we could try out? 

How would the keyword work then? Any usage ideas, keyword name and so on?

VitoAlbano:
Here it is an example page where you can try the element : http://www.w3schools.com/tags/tryit.asp?filename=tryhtml5_input_multiple

The keyword could be as easy as choose multiple files where you can pass the locator, maybe a base path ( to avoid to put the absolute path for all the files) and a list of file names.

There is a known issue opened : SeleniumHQ/selenium-google-code-issue-archive#2239 which talks about it, and searching around it seems that actually there is only a way to do this.

i'll paste here my code (it sure can be improved but it actually works :) ) : 

```
def choose_multiple_files(self, locator, basePath="",  *files):
    fileList = []
    for file in files:
        path_to_image = os.path.join(basePath, file)
        if not os.path.isfile(path_to_image):
            raise AssertionError("File '%s' does not exist on the local file system" % path_to_image)
        fileList.append(path_to_image)
    selenium2Lib = BuiltIn().get_library_instance('Selenium2Library')
    selenium2Lib._element_find(locator, True, True).send_keys("\n".join(fileList))
```

aaltat:
Closing due inactivity

lovelytxm:
Thank you so much for sharing.

PR #611, opened 2016-06-10
Added new keyword for multiple files upload

VitoAlbano:
This is my proposal for manage via Selenium2Library the <input type="files" multiple> element which allows you to select multiple files instead of only one.

Referral to issue : #609

### DIFF
--- a/src/Selenium2Library/keywords/_formelement.py
+++ b/src/Selenium2Library/keywords/_formelement.py
@@ -1,5 +1,6 @@
 import os
 from keywordgroup import KeywordGroup
+import logging
 from selenium.common.exceptions import WebDriverException
 
 class _FormElementKeywords(KeywordGroup):
@@ -180,8 +181,8 @@ class _FormElementKeywords(KeywordGroup):
                         % file_path)
         self._element_find(locator, True, True).send_keys(file_path)
 
-    def choose_multiple_files(self, locator, basePath="",  *files):
-        """Inputs the `files` into file input field found by `locator`.
+    def choose_multiple_files(self, locator,browser="chrome", basePath="", *files):
+        """Inputs the "files" into file input field found by `locator`.
         This Keyword extend the Choose Files option to allow user to select multiple files
         (input type=files multiple).
         The file paths need to be specified for each file.
@@ -189,8 +190,8 @@ class _FormElementKeywords(KeywordGroup):
         The file specified with `file_path` must be available on the same host
         where the Selenium Server is running.
         Example:
-        | Choose Multiple Files | my_upload_field | /home/user/files/  |  trades.csv,trades.xsl,trades.doc |
-        | Choose Multiple Files | my_upload_field |                    | /folder/trades.csv,/folder1/trades.xsl,/folder3/trades.doc |
+        | Choose Multiple Files | my_upload_field | chrome  | /home/user/files/ | trades.csv | trades.xsl | trades.doc |
+        | Choose Multiple Files | my_upload_field | firefox |                   | /folder/trades.csv | /folder1/trades.xsl | /folder3/trades.doc |
         """
         fileList = []
         for file in files:
@@ -198,7 +199,11 @@ class _FormElementKeywords(KeywordGroup):
             if not os.path.isfile(path_to_image):
                 raise AssertionError("File '%s' does not exist on the local file system" % path_to_image)
             fileList.append(path_to_image)
-        self._element_find(locator, True, True).send_keys("\n".join(fileList))
+            if browser not in ("googlechrome", "gc", "chrome"):
+                self._element_find(locator, True, True).send_keys(path_to_image)
+        if browser in ("googlechrome", "gc", "chrome"):
+            logging.info(fileList)
+            self._element_find(locator, True, True).send_keys("\n".join(fileList))
 
     def input_password(self, locator, text):
         """Types the given password into text field identified by `locator`.
--- a/src/Selenium2Library/keywords/_formelement.py
+++ b/src/Selenium2Library/keywords/_formelement.py
@@ -180,6 +180,26 @@ class _FormElementKeywords(KeywordGroup):
                         % file_path)
         self._element_find(locator, True, True).send_keys(file_path)
 
+    def choose_multiple_files(self, locator, basePath="",  *files):
+        """Inputs the `files` into file input field found by `locator`.
+        This Keyword extend the Choose Files option to allow user to select multiple files
+        (input type=files multiple).
+        The file paths need to be specified for each file.
+        Alternative is use basePath argument and then only files name.
+        The file specified with `file_path` must be available on the same host
+        where the Selenium Server is running.
+        Example:
+        | Choose Multiple Files | my_upload_field | /home/user/files/  |  trades.csv,trades.xsl,trades.doc |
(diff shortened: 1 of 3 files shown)


src/Selenium2Library/keywords/_formelement.py
```diff
@@ -180,6 +180,26 @@ def choose_file(self, locator, file_path):
                         % file_path)
         self._element_find(locator, True, True).send_keys(file_path)
 
+    def choose_multiple_files(self, locator, basePath="",  *files):
+        """Inputs the `files` into file input field found by `locator`.
```
aaltat: On documentation following pep257[1] is required. Also following pep008 is required, unless the is good reason to not to follow the pep. Good reason not to follow the pep are the examples in keyword documentation. 

[1] https://www.python.org/dev/peps/pep-0257/
[2] https://www.python.org/dev/peps/pep-0008/

src/Selenium2Library/keywords/_formelement.py
```diff
@@ -180,6 +180,26 @@ def choose_file(self, locator, file_path):
...
+        This Keyword extend the Choose Files option to allow user to select multiple files
+        (input type=files multiple).
+        The file paths need to be specified for each file.
+        Alternative is use basePath argument and then only files name.
+        The file specified with `file_path` must be available on the same host
```
aaltat, line 190: `file_path` -> ``file_path``

src/Selenium2Library/keywords/_formelement.py
```diff
@@ -180,6 +180,26 @@ def choose_file(self, locator, file_path):
                         % file_path)
         self._element_find(locator, True, True).send_keys(file_path)
 
+    def choose_multiple_files(self, locator, basePath="",  *files):
```
aaltat: Would it be better to always provide only the absolute file paths? It would make the keyword usage. Also in this case, it could be considered that should we  could combine the two keywords to one `choose_file` keyword.

src/Selenium2Library/keywords/_formelement.py
```diff
@@ -180,6 +180,26 @@ def choose_file(self, locator, file_path):
...
+        """Inputs the `files` into file input field found by `locator`.
+        This Keyword extend the Choose Files option to allow user to select multiple files
+        (input type=files multiple).
+        The file paths need to be specified for each file.
+        Alternative is use basePath argument and then only files name.
+        The file specified with `file_path` must be available on the same host
+        where the Selenium Server is running.
+        Example:
+        | Choose Multiple Files | my_upload_field | /home/user/files/  |  trades.csv,trades.xsl,trades.doc |
```
aaltat: I am pretty sure that in the example `trades.csv,trades.xsl,trades.doc` is not correctly formatted. I think it should be 

```
| Choose Multiple Files | my_upload_field | /home/user/files/  |  trades.csv | trades.xsl | trades.doc |
```

Also fix the example in line 193.

aaltat:
Write also test, new keywords are not accepted without proper tests.

aaltat:
Unfortunately the contributing guidelines are not yet in master, which would have helped you to awoid many of the comments, but for mean time they can be found from here: https://github.com/robotframework/Selenium2Library/pull/509/files

VitoAlbano:
Thanks for the support , i'll add all the necessary things :)

src/Selenium2Library/keywords/_formelement.py
```diff
@@ -180,6 +180,26 @@ def choose_file(self, locator, file_path):
...
+        | Choose Multiple Files | my_upload_field |                    | /folder/trades.csv,/folder1/trades.xsl,/folder3/trades.doc |
+        """
+        fileList = []
+        for file in files:
+            path_to_image = os.path.join(basePath, file)
+            if not os.path.isfile(path_to_image):
+                raise AssertionError("File '%s' does not exist on the local file system" % path_to_image)
+            fileList.append(path_to_image)
+        self._element_find(locator, True, True).send_keys("\n".join(fileList))
```
drobota: This wont work with marionette and other browser see https://github.com/mozilla/geckodriver/issues/98.

aaltat: Good point, instead of using `\n` it should use multiple `send_keys`

VitoAlbano: Already added a change for that :) sorry its still WIP but i think i'll submit the new version today.

Also, actually the multiple file load via geckodriver doesn't trigger "onChange()" action.

I've opened a bug so they are working on it :
https://github.com/mozilla/geckodriver/issues/101

aaltat:
Sorry for delay but I am middle of my house renovation and my PC is in a box (somewhere). Therefore I am not able to perform a proper review anytime soon. I hope that you can wait for while or if some other admin would be able to look at your pull request.

VitoAlbano:
No problem , i'm still working to figure out the best way to check the browser inside the keyword without passing this information from the outside.

src/Selenium2Library/keywords/_formelement.py
```diff
@@ -180,6 +181,30 @@ def choose_file(self, locator, file_path):
...
+        fileList = []
+        for file in files:
+            path_to_image = os.path.join(basePath, file)
+            if not os.path.isfile(path_to_image):
+                raise AssertionError("File '%s' does not exist on the local file system" % path_to_image)
+            fileList.append(path_to_image)
+            if browser not in ("googlechrome", "gc", "chrome"):
+                self._element_find(locator, True, True).send_keys(path_to_image)
+        if browser in ("googlechrome", "gc", "chrome"):
+            logging.info(fileList)
```
ombre42, line 205: Why import logging? Use self._info like elsewhere in this class.

VitoAlbano, line 205: My bad. I've left it inside but my idea was to remove all of the logging, in fact i don't know how much sense has to print the filelist.
Thanks :)

aaltat:
@VitoAlbano are you still interested to get this to master? If you are, I could perform a proper review but in the mean time could you fix the acceptance test so that they pass.

VitoAlbano:
@aaltat Yes  i'm in, sorry tough time at work :), i'll try to make the acceptance test pass this weekend

aaltat:
Also rebasing with master is a good idea, because there has been quite a lot changes. Also look intointo the ombre42 comments.

VitoAlbano:
Ok i manage to understand what is the problem, if i'm getting everything ok the multiple send keys is something that only gecko driver has implemented and Selenium webdriver is still working on it (https://github.com/SeleniumHQ/selenium/issues/3096).
i'll try to re-update my environment to check if the acceptance test works with a newer version of Selenium/Firefox (so it can use geckodriver). As far as i can see the Selenium2Library Travis build uses firefox 45.4.0 just pre-gecko (that starts from version 46 if i remember correctly). On chrome it seems to run everything fine.

VitoAlbano:
@aaltat due the fact that it seems that on FF we won't have soon this kind of chance, should i change the PR to allow it only for Chrome? so changing the test to make it run only for chrome acceptance and removing the multiple send_keys part .

aaltat:
We can exclude the latest FF but solution should work with IE, Chrome, Safari and older version of FF.

febinthomas04:
try this:

Selenium2Library.input text         xpath    file1 \n file2 \n file3


worked for all my cases :)

aaltat:
Closing due inactivity

PR #613, opened 2016-06-13
Fixes #262 by updating keyword documentation

### LINKED ISSUE #262: Document that jQuery is not provided by Selenium2Library
If that so, please update the keyword document to remove the information about jquery & sizzle

aaltat:
Updates Wxecute Javascript keyword the documentation to mention that javascript is not provided by the Selenium2Library

### DIFF
--- a/src/Selenium2Library/__init__.py
+++ b/src/Selenium2Library/__init__.py
@@ -78,6 +78,10 @@ class Selenium2Library(
     This can be fixed by changing the locator to:
     | Click Link    default=page?a=b
 
+    Please note that jQuery is not provided by Selenium2Library
+    and if there is need to use jQuery locators, the system 
+    under test must provide the jQuery library.
+
     *Using webelements*
     ------------------
     Starting with version 1.7 of the Selenium2Library, one can pass an argument


HelioGuilherme66:
Well, you have a typo ;).
(you have "not" instead of "note")

## "Please **note** that JavaScript ..."

How about expanding to:
"Please note that JavaScript (or any derivative/wrapper, like for example JQuery) ..."

aaltat:
Good catch @HelioGuilherme66, fixed.

aaltat:
There are some many wrappers... Perhaps it would be best only to say javascript?

aaltat:
Moved documentation from the keyword to library level, based on ed comments on #262

src/Selenium2Library/__init__.py
```diff
@@ -78,6 +78,10 @@ class Selenium2Library(
     This can be fixed by changing the locator to:
     | Click Link    default=page?a=b
 
+    Please note that JavaScript is not provided by Selenium2Library and
+    if there is need to use JavaScript must be provided by the system
+    under test.
+
```
emanlove, line 84: Not JavaScript but jQuery. Slightly rewording it might say

> Please note that jQuery is not provided by Selenium2Library and if there is need to use jQuery locators, the system under test must provide the jQuery library.

or

> In regards to the jQuery locator stratergy, please note that jQuery is not provided by Selenium2Library. If the test uses jQuery locators, the system under test must provide the jQuery library.

aaltat:
I Agree with @emanlove  and changed the text based on his comments.

aaltat:
Are we happy with this pull request and it be merged to the master?

PR #552, opened 2015-12-16
Add ability to pass custom capabilities to local browsers

drobota:
Issue #550

PR #623, opened 2016-07-22
Improve behavior of setting desired_capabilities

matthewvalimaki:
Remove forceful Firefox profile definition.
Add ability to remove default desired_capabilities definitions.

Signed-off-by: Matthew Valimaki matthew.valimaki@gmail.com

### DIFF
--- a/src/Selenium2Library/keywords/_browsermanagement.py
+++ b/src/Selenium2Library/keywords/_browsermanagement.py
@@ -540,10 +540,9 @@ class _BrowserManagementKeywords(KeywordGroup):
 
         if profile_dir:
             if profile_dir != "":
-            profile = webdriver.FirefoxProfile(profile_dir)
+                profile = webdriver.FirefoxProfile(profile_dir)
         else:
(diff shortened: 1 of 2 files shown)


matthewvalimaki:
A bit more background:
When Selenium node is on headless Alpine Linux Firefox would refuse to start up because of how Selenium2Library set `firefox_profile`. Selenium2Library always sets `firefox_profile` without giving me the option to not to set it. Likewise `marionette: false` was being set again giving me no way to remove it altogether. Yes, I could do `marionette: true` but that is not what I'm looking for, I want complete removal of the value.

The changes I've introduce do two things:
1) do not enforce Firefox profile. Validate the `ff_profile_dir` if defined.
2) allow removal of key-values from `desired_capabilities`

aaltat:
Could you create an issue about the problem and describe the problem which you are trying to solve. Issue helps us to track the status of the project.

aaltat:
As a side note, would the #552 also solve this problem?

aaltat:
Closing due inactive.

PR #627, opened 2016-08-08
Fixes #257 by removing the docutils dependency

### LINKED ISSUE #257: Remove unnecessary `docutils` dependency
This library currently has following dependencies:

```
decorator >= 3.3.2
selenium >= 2.32.0
robotframework >= 2.6.0
docutils >= 0.8.1
```

Three first are actually used by the library, but docutils is only used for compiling README.rest and other reStructuredText files to HTML by `doc/generate_readmes.py` and `doc/buildhtml.py`. This is just a developer/packager activity and thus users don't really need docutils for anything.

I would like to remove the docutils dependency to make it easier to document what dependencies are needed in manual installation (see #255). Avoiding an unnecessary download when using pip wouldn't obviously hurt either.

I'm not certain is compiling reST to HTML really needed in general. Users looking at source probably are fine with *.rst source files and in GitHub (and PyPI #254) these files are formatted automatically. If the compilation isn't needed anymore, then the need for docutils obviously goes away. If there is a use case for the compilation, I would still be more than willing to remove the docutils dependency from users. If we want to be helpful for the developer running the aforementioned compilation scripts, we can change them to import docutils like this:

```
try:
    import docutils
except ImportError:
    raise ImportError('Importing `docutils` failed. Install it with `pip install docutils`.')
```

aaltat:
The docutils is only needed for Selenium2Library development,
therefore removed the dependency from setup.py. Also added meaningfull
error message to development tools, if the docutils is not installed.

### DIFF
--- a/doc/buildhtml.py
+++ b/doc/buildhtml.py
@@ -26,13 +26,16 @@ import os
 import os.path
 import copy
 from fnmatch import fnmatch
-import docutils
-from docutils import ApplicationError
-from docutils import core, frontend, utils
-from docutils.error_reporting import ErrorOutput, ErrorString
-from docutils.parsers import rst
-from docutils.readers import standalone, pep
-from docutils.writers import html4css1, pep_html
+try:
+    import docutils
+    from docutils import ApplicationError
+    from docutils import core, frontend, utils
+    from docutils.error_reporting import ErrorOutput, ErrorString
+    from docutils.parsers import rst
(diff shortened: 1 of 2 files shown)


pekkaklarck:
Looks good. Could add separate `requirements-dev.txt` for pip that only contains `docutils`, but clear error message is good too. Especially when we actually may not really need that `buildhtml.py` script.

Issue #634, opened 2016-08-22
Cleanup build instructions and scripts


aaltat:
The Python build and release has changed since the original Selenium2Library build instructions and scripts has been written. Cleanup the scripts to support modern ways.

Included tasks:
- [x] Better build docs
- [x] Clean up `setup.py`
- [x] Remove `ez_setup` (done as separate issue #646)
- [x] Clean up `MANIFEST.in`
- [x] ~Clean up~ Remove `setup.cfg`
- [x] Better build scripts (probably based on Invoke like with RF itself)
- [x] Remove references to already deleted `doc` directory
- [x] Remove references to already deleted `demo` directory

pekkaklarck:
_Edited by @zephraph to move checklist to the ticket description for more visibility_

just-be-dev:
Hey @pekkaklarck, I moved your checklist up to the description that way it can be seen from the issues list.

emanlove:
I would recommend we also upgrade the test runner (which could be another ticket).

pekkaklarck:
Agreed. I would assume Python 3 adds some new requirements for test runners too. This is also something where we can look for inspiration (or steal code) from RF. If there are lot of very similar needs, we may want to extract some build and test scripts to a separate shared project. Anyway, enhancing test runners definitely needs a separate task.

pekkaklarck:
All checkboxes in the description checked.

PR #641, opened 2016-08-26
Update ez_setup.py newest version 0.26 to make it more compatible for all python versions

eino-makitalo:
If this packages is still needeed it's better to be newest version supporting all python versions.
ez_setup.py from https://bootstrap.pypa.io/ez_setup.py - python 2/3


pekkaklarck:
I've been thinking we should just delete ez_setup.py. See #634.

eino-makitalo:
Yes. I read #634, thats better plan.  It's better to close this.

PR #647, opened 2016-09-05
Remove ez_setup.py and references to it in setup.py.

### LINKED ISSUE #646: Remove `ez_setup`
`ez_setup` module has been used to automatically install `setuptools` as part of S2L installation. Nowadays we can expect everyone to already have `pip` and `setuptools` and thus can remove `ez_setup` altogether. This also avoids us updating `ez_setup.py` to Python 3 compatible version.

pekkaklarck:
Fixes #646.


aaltat:
Looks good to me.

PR #649, opened 2016-09-05
Unify print() usage (#684)

aaltat:


pekkaklarck:
First of all, I think it would be better to have one PR to fix all `print` statements.

Secondly, the change in this PR doesn't work correctly without `from __future__ import print_function`. In Python 2 using `print()` will print `()` without the aforementioned import.

aaltat:
Ah, so true. Then I will add the import

aaltat:
Because print statement is used mainly on the development tools, that would put the `future` installation requirement to the `requirements-dev.txt`. But if need to use the `future` in places which are used by the library itself then the `future` installation requirement should go to `requirements.txt`. I do not see that far, that I could say for sure for which one it should be put. Because not the need is only on the dev tools, should we put the installation requirement only in the `requirements-dev.txt`?

pekkaklarck:
`from __future__ import print_function` works out-of-the-box since Python 2.6. In fact, all `__future__` imports are handled by Python itself.

aaltat:
Closed on favor of #684

PR #651, opened 2016-09-05
Fix imports

HelioGuilherme66:
Using futurize for Python 2.7 absolute imports, and PEP8 style for imports.


pekkaklarck:
This PR has all kinds of unrelated changes. For example, when looking at the "Files changed" tab, the first change I see is removal of CONTRIBUTING.md. There are also 26 individual commits. Most likely this PR is based on an existing PR or branch. Whatever the reason, there is no possibility this PR could be merged.

PR #654, opened 2016-09-06
Fix remaining print

HelioGuilherme66:
Some left over print were not changed to print().


pekkaklarck:
This PR contains lot of totally unrelated changes. Could you @HelioGuilherme66 please review PR contents in a bit more detail before you submit them? It's time taking for others to review them when they are big and contain changes that shouldn't be there in the first place.

PR #655, opened 2016-09-06
Unit Tests will not run on Fedora without this PR

HelioGuilherme66:
Setting include_root_package_name to True, will allow running unit tests in Fedora 22, and all other O.S.

### DIFF
--- a/test/run_unit_tests.py
+++ b/test/run_unit_tests.py
@@ -5,7 +5,7 @@ from Selenium2Library import utils
 
 def run_unit_tests(modules_to_run=[]):
     (test_module_names, test_modules) = utils.import_modules_under(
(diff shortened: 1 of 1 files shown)


aaltat:
That whole test runner logic would need refactoring. There should and are cleaner way to import stuff for testing. If your current code base does not allow it, we should do something about the code base then. But it looks that this did not break anything, tested with Ubuntu and Windows 7.

pekkaklarck:
I already rewrote the unit test runner. Got sidetracked with vendored mockito and other weirdness but I have a PR ready soon.

aaltat:
After refactoring is this still required?

pekkaklarck:
No, the code changed here will be removed altogether. Closing this.

Issue #620, opened 2016-07-07
Drop Python 2.6 support


pekkaklarck:
Python 2.6 hasn't been officially supported since October 2013 and Robot Framework is going to drop support for in the next major release. There's no need for Selenium2Library to spent development and CI resources to support it anymore in its major release either.

For more details and reasoning see robotframework/robotframework#2276.

pekkaklarck:
In practice this change would mean that we don't include Python 2.6 in our forthcoming CI system and stop testing it in general. I don't think we need to explicitly check is someone using Python 2.6 or try to go through the code to check for places where we could use Python 2.6 features. This issue is here mainly to act as note to be added to the release notes.

aaltat:
The only thing left from this is to documentation and fixing the`setup.py`

aaltat:
Also check `ÌNSTALL.rst` doc

aaltat:
Done in many commits.

PR #662, opened 2016-09-06
Clean up run_unit_tests.py.

pekkaklarck:
Finding unit tests is rewritten by using test discovery functionality
provided by the unittest module in Python 2.7 and newer. This means
unit tests cannot be executed with Python 2.6 anymore, but we've
decided to drop its support anyway (#620). Tests won't be run with
Python 2.6 on Travis anymore either.

Also removed utils related to loading modules from S2L.utils. This
unit test related code didn't belong under the main project in the
first place.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 sudo: false
 python:
-  - "2.6"
   - "2.7"
 install:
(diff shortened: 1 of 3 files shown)


aaltat:
Tested and also looks OK.

PR #668, opened 2016-09-23
Libdoc updates - link to project and fix 404

ombre42:
The first hit on the U.S. English Google is to the libdoc, which is a dead end. So I thought it would be good to link to the project page. Also I found a 404 on the user guide caused by RF migrating to GitHub.


src/Selenium2Library/__init__.py
```diff
@@ -19,7 +19,10 @@ class Selenium2Library(
     _WaitingKeywords,
     _AlertKeywords
 ):
-    """Selenium2Library is a web testing library for Robot Framework.
+    """Selenium2Library is a web testing library for Robot Framework. This
+	document is about using Selenium2Library. For information about
+	installation, support, and more please visit the
+	[https://github.com/robotframework/Selenium2Library|project page].
```
pekkaklarck, line 27: Good idea to add this information, but it should be a separate paragraph indented same amount as other paragraphs. For more details see PEP-257: https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

src/Selenium2Library/__init__.py
```diff
@@ -130,8 +133,8 @@ class Selenium2Library(
 
     All timeouts can be given as numbers considered seconds (e.g. 0.5 or 42)
     or in Robot Framework's time syntax (e.g. '1.5 seconds' or '1 min 30 s').
-    For more information about the time syntax see:
-    http://robotframework.googlecode.com/svn/trunk/doc/userguide/RobotFrameworkUserGuide.html#time-format.
+    For more information about the time syntax see the
+	[http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#time-format|Robot Framework User Guide].
```
pekkaklarck: Good catch! Also good idea to use the link format supported by Libdoc nowadays. The leading extra spaces (or is that a tab?) should be removed, though, and the last line aligned same way as the previous one.

pekkaklarck (commented):
Good enhancements. Please see my line notes for few things that could be enhanced.

src/Selenium2Library/__init__.py
kormbrek-rally: thanks for the feedback. even though it is docstring, it exists to generate the library documentation, so I did not think about the normal conventions. I will update it.

src/Selenium2Library/__init__.py
ombre42: I dunno how that happened but I will fix it. and will work on keeping my multiple github accounts straight :)

aaltat:
From my mobile screen all looks OK, but have to look from a bigger screen, because last time I did missed the line 137 problem from the mobile screen.

ombre42:
New laptop. Did not realize I was using tabs instead of 4 spaces. Silly me. Hope to contribute more meaningfully again in the future. This PR was just getting my feet wet after being out of the pool for so long. @pekkaklarck or other contributors please try the squash during merge feature - great addition to GitHub. @pekkaklarck thanks again for your attention to detail. In then end it really pays off as it is all to easy to build up technical debt.

pekkaklarck:
Thanks for the PR @ombre42 and looking forward for more!

Tabs/spaces issues occur very easily unless you configure the editor to always use spaces (or the editor does that by default). Luckily Python 3 is more strict about mixing tabs/spaces and such errors cannot occur anymore with it. Python 2 also has `-tt` option to issue errors about inconsistent tab usage. Could we add it to our test scripts somewhere?

The new squashing functionality on GitHub is great. @aaltat, please merge the PR once you had a chance to look at it on a bigger screen.

Issue #673, opened 2016-10-07
Support locating elements using element class


aaltat:
Decide should selenium find_by_class exposed as selector strategy. Based on discussion on user group: https://groups.google.com/forum/m/#!topic/robotframework-users/2gzzmGAhofY

emanlove:
I vote yes. In my mind the most basic concept for the Selenium2Library is it exposes and mirrors the functionality of the underlying Selenium Python WebDriver. As this is one of the base find_by method it makes sense that it should be within the core locator strategies.

aaltat:
I am thinking to the same direction. Also it is something which is not that hard to do.

Gaurang033:
if no one has started working on it, i can take this up.

aaltat:
Feel free to make a pull request. Remember to write tests and update the library keyword documentation on the locator strategies.

Gaurang033:
Created the pull request #584. Please check.

aaltat:
Please make a separate pull request to resolve this issue.

Gaurang033:
created a separate pull request as suggested.  #674

aaltat:
Thank you for the update.

aaltat:
Done in https://github.com/robotframework/Selenium2Library/pull/674

PR #674, opened 2016-10-14
Added method to find element by class name #673

Gaurang033:
Added a new locator strategy find by class name and unit test case for it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+addons:
+  firefox: latest-esr
 language: python
 sudo: false
 python:
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ install:
   #- pip install pep8 --use-mirrors
   #- pip install pyflakes --use-mirrors
   - pip install robotframework
-  - pip install selenium
+  - pip install selenium==2.53.6
   - pip install decorator
   - pip install mockito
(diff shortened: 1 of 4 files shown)


aaltat:
I did restart the Travis build because it seems that the error is not related to your chances.

PR must also update the library documentation in here https://github.com/robotframework/Selenium2Library/blob/master/src/Selenium2Library/__init__.py Update the chapter which describes the locator strategies.

Usually it would be good idea to write acceptance test also. It would help to detect if the selenium interface changes.

aaltat:
The Travis build did fail again. Would you mind looking what the actual failure seems to be?

Gaurang033:
Hi,

I just checked the logs, it seems like geckodriver execute path is not set,
Following is the error message. All the tests are passing locally.
WebDriverException: Message: 'geckodriver' executable needs to be in PATH.

aaltat:
Oh darn. In that case it seems that Travis did update the Firefox to a version which needs the geckodriver. Therefore it needs some work from the admin side. Have to take a look and fix it asp. Let's see can I find time later today.

HelioGuilherme66:
@aaltat I have already fixed Travis build. I have restarted the build of this PR.

aaltat:
The pull request needs to rebased locally with master and push again to upstream.

PR #678, opened 2016-10-18
Update travis conf

drobota:
@aaltat, @HelioGuilherme66, @pekkaklarck  please, have a look.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,24 +2,51 @@ addons:
   firefox: latest-esr
 language: python
 sudo: false
-python:
-  - "2.7"
 install:
-  #- pip install pep8 --use-mirrors
-  #- pip install pyflakes --use-mirrors
-  - pip install robotframework
-  - pip install selenium==2.53.6
-  - pip install decorator
-  - pip install mockito
-  #- "if [[ ${BROWSER} == 'ie' ]]; then pip install ie_only_pkg; fi"
+  - pip install .
+  - pip install -r requirements.txt
+  - pip install -r requirements-dev.txt
 before_script:
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
 script:
-  - "firefox --version"
-  - "python test/run_tests.py python $BROWSER --noncritical known_issue_-_travisci"
+  - "if [[ -n ${SELENIUM_VERSION+x} ]]; then pip install -U selenium==${SELENIUM_VERSION}; fi"
+  - "python test/run_tests.py python \"$BROWSER\""
 env:
-  matrix:
-  - BROWSER=firefox
-  #- BROWSER=ie # will be added when we have completed SauceLabs CI
-  #- BROWSER=safari # will be added after #247 merged
+  global:
+    - secure: "QyY0j8C+hVhsGLv4YU0UMvLm5wBDg6xT5J0Ee77C/JAQ6vcdCMpTfWI8oDxTZb7pBhpDkYuNe747yCYF5PrdBYQbO08FilxctReuJfMXcH44rZC5q49mUmPamkvg1LkzIWuC1s9uKnjlvyDT5ojuPXZl2Vn0gIhSIg+yOyBkM1U="
+    - secure: "HJaJTYauaBZODMDkbnRVsiCk+t2uVtztkkJ5V4dXSnE1uIBms2PVbufQwb+YNvzHkGheLtpnt/hLM1HhmSFgUtOCf+eumQ6kaLR2KRVEysXwWJzgZoLBidNBOy5rvAIsc9JJbDaIgYVGzyaJGTP9AMb8CyvIPE//lH10sf+GB7g="
+matrix:
+  include:
+    - os: linux
+      env:
+        - BROWSER=firefox
+        - SELENIUM_VERSION=2.53.6
+      addons:
+        firefox: "47.0.1"
+      python: "2.7"
+    - os: linux
+      env:
+        - BROWSER=chrome
+      addons:
+        sauce_connect: true
+      python: "2.7"
+    - os: linux
+      env:
+        - BROWSER=firefox
+        - SELENIUM_VERSION=2.53.6
+      addons:
+        sauce_connect: true
+      python: "3.4"
+    - os: linux
+      env:
+        - BROWSER="internet explorer"
+      addons:
+        sauce_connect: true
+      python: "2.7"
+    - os: linux
+      env:
+        - BROWSER=safari
+      addons:
+        sauce_connect: true
+      python: "2.7"
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
 decorator >= 3.3.2
-selenium >= 2.32.0
+selenium
 robotframework >= 2.6.0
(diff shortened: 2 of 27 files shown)


ombre42:
Increasing minimum version of selenium and also pegging to a specific version? This does not just affect CI but also where this library is installed or did I miss something?

drobota:
Now selenium 2 library can only work with old Firefox driver which works only in firefox up to version 47.0.1 and selenium 2.53.6. We might need drop supporting it and switch to marionette(https://github.com/mozilla/geckodriver)

ombre42:
Understandable but pip can handle having a minimum and maximum version (I.e. range of compatible versions). Usually we try to give flexibility to the end user. Also I think a big change like selenium version should be mentioned in commit message and PR message IMHO.

aaltat:
It should be possible to modify existing functionality in Open Browser keyword to support Firefox with and without the marionette. Also it should be possible to make the change in such manner that the it doesn't cause changes to the keyword signature.

I am thinking how to handle the change with the grid side. Should the keyword make some automatic setting with the desired capabilities or not. I am starting to think that we should not make any magic when user set the desired capabilities.

Therefore I am against changing the minimum required selenium version. Also the selenium 3 should be only drop in replacement for selenium 2 and it doesn't cause changes to the selenium API.

Of course we should recommended people to use latest selenium but we should not force it.

.travis.yml
```diff
@@ -2,24 +2,42 @@ addons:
...
+  - "python test/run_tests.py python \"$BROWSER\""
 env:
-  matrix:
-  - BROWSER=firefox
-  #- BROWSER=ie # will be added when we have completed SauceLabs CI
-  #- BROWSER=safari # will be added after #247 merged
+  global:
+    - secure: "QyY0j8C+hVhsGLv4YU0UMvLm5wBDg6xT5J0Ee77C/JAQ6vcdCMpTfWI8oDxTZb7pBhpDkYuNe747yCYF5PrdBYQbO08FilxctReuJfMXcH44rZC5q49mUmPamkvg1LkzIWuC1s9uKnjlvyDT5ojuPXZl2Vn0gIhSIg+yOyBkM1U="
+    - secure: "HJaJTYauaBZODMDkbnRVsiCk+t2uVtztkkJ5V4dXSnE1uIBms2PVbufQwb+YNvzHkGheLtpnt/hLM1HhmSFgUtOCf+eumQ6kaLR2KRVEysXwWJzgZoLBidNBOy5rvAIsc9JJbDaIgYVGzyaJGTP9AMb8CyvIPE//lH10sf+GB7g="
+matrix:
```
aaltat, line 19: Should we add some run with Python 3 too. Example have to runs with Chrome one with Python 2.7 and another one with Python 3.

requirements.txt
```diff
@@ -1,3 +1,3 @@
 decorator >= 3.3.2
-selenium >= 2.32.0
+selenium == 2.53.6
```
aaltat: Despite my previous comments, I think this change is OK, specially with the selenium 3. Supporting old version of selenium would need extra runs from the CI and in the current model, we do not have resource to provide the support and we do not have resources to pay for the CI time (or actually pay for the cloud service to get all the browsers running)

.travis.yml
```diff
@@ -2,24 +2,42 @@ addons:
...
-  - "firefox --version"
-  - "python test/run_tests.py python $BROWSER --noncritical known_issue_-_travisci"
+  - "python test/run_tests.py python \"$BROWSER\""
 env:
-  matrix:
-  - BROWSER=firefox
-  #- BROWSER=ie # will be added when we have completed SauceLabs CI
-  #- BROWSER=safari # will be added after #247 merged
+  global:
+    - secure: "QyY0j8C+hVhsGLv4YU0UMvLm5wBDg6xT5J0Ee77C/JAQ6vcdCMpTfWI8oDxTZb7pBhpDkYuNe747yCYF5PrdBYQbO08FilxctReuJfMXcH44rZC5q49mUmPamkvg1LkzIWuC1s9uKnjlvyDT5ojuPXZl2Vn0gIhSIg+yOyBkM1U="
```
aaltat, line 17: Are the Sauce Labs hash keys to access the account. You have the clear passwords saved in somewhere? Perhaps we should distribute the password with wider audience but not put the passwords in git repo.

Is there reason why the are two secure lines?

src/Selenium2Library/keywords/_element.py
```diff
@@ -363,9 +365,11 @@ def click_element_at_coordinates(self, locator, xoffset, yoffset):
...
         self._info("Click clicking element '%s' in coordinates '%s', '%s'." % (locator, xoffset, yoffset))
         element = self._element_find(locator, True, True)
-        #self._element_find(locator, True, True).click()
-        #ActionChains(self._current_browser()).move_to_element_with_offset(element, xoffset, yoffset).click().perform()
-        ActionChains(self._current_browser()).move_to_element(element).move_by_offset(xoffset, yoffset).click().perform()
+        ActionChains(
+            self._current_browser()).\
+            move_to_element(element).\
+            move_by_offset(xoffset, yoffset)\
+            .click().perform()
```
aaltat: It could be cleaner to write: 
        action = ActionChains(self._current_browser())
        action.move_to_element(element)
        action.move_by_offset(xoffset, yoffset)
        action.click().perform()

src/Selenium2Library/keywords/_element.py
```diff
@@ -587,7 +590,6 @@ def mouse_down_on_image(self, locator):
 
     def page_should_contain_image(self, locator, message='', loglevel='INFO'):
         """Verifies image identified by `locator` is found from current page.
-
         See `Page Should Contain Element` for explanation about `message` and
```
aaltat, line 596: I think empty line should have not been removed.

test/acceptance/create_webdriver.robot
```diff
@@ -7,10 +7,11 @@ Library           Collections
 Create Webdriver Creates Functioning WebDriver
     [Documentation]    LOG 2:1 INFO REGEXP: Creating an instance of the \\w+ WebDriver
     ...    LOG 2:4 DEBUG REGEXP: Created \\w+ WebDriver instance with session id (\\w|-)+
+    [Tags]  Known Issue Chrome    Known Issue Internet Explorer    Known Issue Safari
```
aaltat, line 10: Should we document the tags we use, so that people contributing could also use them right?

aaltat (changes requested):
One think which I do not fully grasp with Travis, where are those browsers opened?  Are they opened by Travis or are they used with SauceLabs? Where those remote_url and the desired_capabilities are coming from?

requirements.txt
aaltat: But instead `selenium == 2.53.6` should we say `selenium >= 2.53.6`

.travis.yml
drobota: Yes, these are encoded login and password for sauce labs. They can be securely stored in repo because they can be decoded only in travis env.

.travis.yml
drobota: I can add it, but does s2l compatible with py3?

.travis.yml
aaltat: OK, this is good to go then.

.travis.yml
aaltat: No it is not yet compatible. But the aim for the 2.0 release, which hopefully is the next release, is to make it compatible with Python 3.

aaltat (approved):

PR #682, opened 2016-10-24
Let's allow robot to get browser's logs

SergiuTudos:
Add new keyword get_log


src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -331,6 +331,14 @@ def select_window(self, locator=None):
...
         finally:
             self._window_manager.select(self._current_browser(), locator)
+            
+    def get_log(self, log_type):
+        """Get the log for a given selenium log type
+
+        The `log_type` argument defines which logs to get. Possible values are:
+        `browser`, `driver`, `client` or `server`
+        """
+        return self._current_browser().get_log(log_type)
```
HelioGuilherme66, line 342: There is no validation on the values of `log_type`. If we pass `BROWSER` we would have an error.

test/acceptance/keywords/javascript.robot
```diff
@@ -107,3 +107,12 @@ Drag and Drop by Offset
...
     Element Text Should Be    id=droppable    Dropped!
+
+Verify Console Log Can be Caught
+   ${message}  Set Variable   Sample Console Error
+   Execute Javascript  console.error('${message}')
+   ${logs}=  Get Log  browser
+   ${length}=  Get Length   ${logs}
+   ${last pos}=  Evaluate  ${length}-1
+   &{err}=  Set Variable  @{logs}[${last pos}]
+   Should Contain  ${err.message}  ${message}
```
HelioGuilherme66: Missing a newline at the end of file.

test/acceptance/keywords/javascript.robot
```diff
@@ -107,3 +107,12 @@ Drag and Drop by Offset
     Element Text Should Be    id=droppable    Drop here
     Drag and Drop by Offset    id=draggable    ${100}    ${20}
     Element Text Should Be    id=droppable    Dropped!
+
+Verify Console Log Can be Caught
+   ${message}  Set Variable   Sample Console Error
+   Execute Javascript  console.error('${message}')
+   ${logs}=  Get Log  browser
```
HelioGuilherme66, line 115: We should have tests for all possible cases. In this test `browser` would not contain the expected message, "Sample Console Error". (It did show in Firefox console).
Did some trials with `driver` and it got output, but with `client` and `server` I've got zero size ${logs}.

The test is failing at CI, because the log entry you are getting is a localized message about no specification of the file encoding (and not containing the expected message). In my system and in PT, is:
`'A codificação de caracteres do documento HTML não foi declarada. O documento será mostrado com texto ilegível em algumas configurações de navegadores se o documento tiver caracteres fora do intervalo US-ASCII. A codificação de caracteres da página tem de ser declarada no documento ou no protocolo de transferência.'`

test/acceptance/keywords/javascript.robot
HelioGuilherme66: Overall, this may be an useful keyword. Hope you could improve it. Thanks.

src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -331,6 +331,18 @@ def select_window(self, locator=None):
...
+            
+    def get_log(self, log_type):
+        """Get the log for a given selenium log type
+
+        The `log_type` argument defines which logs to get. Possible values are:
+        `browser`, `driver`, `client` or `server`
+        """
+        available_types = ["browser", "driver", "client" , "server"]
+        if log_type not in  available_types:
+            raise  ValueErrorValueError(log_type + " is not a supported type")
```
ombre42: We should not restrict the value of `log_type`. There is a `log_types` attribute on web driver to fetch valid values that makes a call to the driver server ([doc](http://seleniumhq.github.io/selenium/docs/api/py/webdriver_remote/selenium.webdriver.remote.webdriver.html#selenium.webdriver.remote.webdriver.WebDriver.log_types)). This leaves flexibility for other types of logs.

src/Selenium2Library/keywords/_browsermanagement.py
```diff
@@ -331,6 +331,18 @@ def select_window(self, locator=None):
             pass
         finally:
             self._window_manager.select(self._current_browser(), locator)
+            
+    def get_log(self, log_type):
+        """Get the log for a given selenium log type
```
ombre42, line 336: The Python documentation for the web driver method is pretty inadequate. But we should be better. The information should include most of the content from the Javadoc. Also how to enable logging and setting levels would be nice to see.

> Fetches available log entries for the given log type. Note that log buffers are reset after each call, meaning that available log entries correspond to those entries not yet returned for a given log type. In practice, this means that this call will return the available log entries since the last call, or from the start of the session. For more info on enabling logging, look at LoggingPreferences.

ombre42 (commented):
This feature would be useful, but the implementation needs improvement

src/Selenium2Library/keywords/_browsermanagement.py
SergiuTudos: Does anyone have an idea why console.error fails on firefox? everything passes on my local machine

aaltat:

SergiuTudos:
what exactly should I improve. Your help would be valuable

aaltat:
The comment about improving the  keyword documentation is still a valid one.

Also the test makes me wonder, should we convert the message automatically to string? But perhaps it's best to return the log as selenium returns it. 

Also if you want credits for the feature add your self to the changes.rst file

aaltat:
Closing due inactivity

PR #693, opened 2016-11-26
Makes unit tests runner more compatible between operating systems

HelioGuilherme66:
Now calls run_unit_tests.py file instead of the method (this fixes failure in Fedora Core 22).
Get path to python/jython to run unit tests.

### DIFF
--- a/test/run_tests.py
+++ b/test/run_tests.py
@@ -3,10 +3,13 @@ from __future__ import print_function
 import env
 import os
 import sys
+from os.path import abspath, dirname, join
 from subprocess import Popen, call
 from tempfile import TemporaryFile
 
 from run_unit_tests import run_unit_tests
+
+CURDIR = dirname(abspath(__file__))
 
 ROBOT_ARGS = [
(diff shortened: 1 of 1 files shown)


pekkaklarck:
It is a good idea to run unit tests as an external process and not by directly calling `run_unit_tests` method. In the current approach unit tests are always executed by using the Python interpreter that runs `run_tests.py` script.

The implementation to run tests using `sys.executable` doesn't make much sense, though, because it ends up using the exact same interpreter that is used for running `run_tests.py`. There's also a problem that on Jython `sys.executable` may not always be set correctly.

A good way to accomplish the original goal would be using the given `interpreter` to run the unit tests. This should be done after the larger acceptance test cleanup done in PR #710, though. Or perhaps I could add this enhancement into that PR as well.

HelioGuilherme66:
Will need rewrite after merging #716.
(Will use --interpreter value)

HelioGuilherme66:
Closing in favor of #720

PR #694, opened 2016-11-29
Fixes Chromedriver tests under Windows, by @yahman72

HelioGuilherme66:
(from his commit: webdriver process management)


HelioGuilherme66:
Tested under Windows 7 Professional x64 and Python 2.7.12 32bit, selenium 2.53.6, RF 3.0
ChromeDriver  2.25.426923 - Chrome 54.0.2840.99 m - 0 failed tests
InternetExplorerDriver (32-bit) 2.53.1.0 - IE 11.0.31 - 23 failed tests
Firefox 47.0.1 - 0 failed tests
PhantomJS 2.1.1 - 21 failed tests (excluded Choose File test)

aaltat:
There are few minor things to fix, mostly code cleanup and following pep8. But those things are hard to comment with mobile. I will try to find time to comment the changes later. Could you raise an issue about the problem?

test/resources/testserver/testserver.py
```diff
@@ -4,17 +4,24 @@
 import BaseHTTPServer
 import httplib
 import os
-
+import socket
 
 class StoppableHttpRequestHandler(SimpleHTTPServer.SimpleHTTPRequestHandler):
```
aaltat, line 10: Two empty lines before the class definition

test/resources/testserver/testserver.py
```diff
@@ -78,6 +88,32 @@ def serve_forever(self):
         while not self.stop:
             self.handle_request()
 
+class HttpEchoer(socket._fileobject):
```
aaltat, line 94: Two empty lines before the class definition

test/run_tests.py
```diff
@@ -46,8 +46,9 @@ def acceptance_tests(interpreter, browser, args):
 
 def start_http_server():
     server_output = TemporaryFile()
+    #server_output = open("server.log", 'w+', buffering=False)
```
aaltat: If this is commented out, where the line 51 server_output comes.

test/resources/testserver/testserver.py
```diff
@@ -63,12 +70,15 @@ def send_head(self):
         self.send_response(200)
         self.send_header("Content-type", ctype)
         fs = os.fstat(f.fileno())
-        self.send_header("Content-Length", str(fs[6]))
+        # chrome hangs if this server runs on windows
+        # it probably waits for more data -> os.fstat() does not return correct length?
+        # we are responding with HTTP 1.0 = we do not need content-length header
+        #self.send_header("Content-Length", str(fs[6]))
         self.send_header("Last-Modified", self.date_time_string(fs.st_mtime))
+        #self.send_header("Connection", "close")
```
aaltat: If this is not needed, remove it

test/resources/testserver/testserver.py
```diff
@@ -63,12 +70,15 @@ def send_head(self):
...
         self.end_headers()
         return f
 
-
```
aaltat, line 71: Two empty lines before the class definition

aaltat (changes requested):

test/run_tests.py
HelioGuilherme66: That is just an alternative way to keep the server logs (like Jari explained in initial code/Issue).

HelioGuilherme66:
Code is PEP8 compliant now.

test/run_tests.py
aaltat: OK, need to read it from there.

aaltat: I did read and looks OK.

aaltat: But should we remove the comment then

HelioGuilherme66: No, we should not remove the comment. We should add a comment describing that by default the output goes to a temporary file, but toggling the commented lines allows developer to keep the output.

aaltat: I don not like commented code, but adding the doc about the that part is good to go.

HelioGuilherme66: Added comment for clarification.

pekkaklarck:
Commented code makes no sense. If we think server logs are needed, we should write code like that. I think that would be fine, writing them into the same directory as other results shouldn't harm. If we want to make it configurable, we should add a command line option. I doubt it's worth the effort in this case.

pekkaklarck:
Also, the code is not PEP-8 compliant.

pekkaklarck:
I changed run_tests.py to preserve server logs in PR #710. Assuming it is merged, changes to run_tests.py in this PR aren't needed anymore.

aaltat:
Also this PR does not resolve the problem in Windows. The test execution still fails with or without this PR. But I must say that it improves the situation a lot.

HelioGuilherme66:
This PR was valid at the time it was proposed.

aaltat:
It could be that this is some sort of windows problem. If I comment out the `with http_server():` from `run_tests.py` and then run the `python -m SimpleHTTPServer 7000` I face similar issues then with the current http server implementation.

Having the support for running test in windows would be really nice, but it strongly looks like that we need to look some other http server.

HelioGuilherme66:
Closing in favor of #748.
The send_header override was removed.

PR #696, opened 2016-12-02
Adds 'Get Locations' keyword

tx3stn:
As a keyword to return a list for all of the currently opened urls (in the same way titles works).

The urls are exposed by `WebDriverMonkeyPatches.get_current_window_info` so I'm not sure if there was a reason this functionality has been left out previously (I had a look through the issues on here and in the google group but couldn't see anything) or if it just wasn't needed.

I'm using it for testing on a site where the url is the most reliable/consistent information so its really useful for switch window command to make sure you're in the right place. 

I've added an acceptance test and a unit test in line with the other `get_window...` keywords.

Let me know if there's any issues with any of this.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,10 @@
 Release Notes
 =============
 
+1.8.1
+-----
+- Added 'Get Locations' keyword [thaffenden]
+
 1.8.0
 -----
 - Moved keyword documentation to:
--- a/src/Selenium2Library/keywords/_browsermanagement.py
+++ b/src/Selenium2Library/keywords/_browsermanagement.py
@@ -346,6 +346,13 @@ class _BrowserManagementKeywords(KeywordGroup):
         """Returns the current location."""
         return self._current_browser().get_current_url()
 
+    def get_locations(self):
+        """Returns and logs current locations of all windows known to the browser."""
+        return self._log_list(
+            [window_info[4] for window_info in
+             self._window_manager._get_window_infos(self._current_browser())]
+        )
+
     def get_source(self):
         """Returns the entire html source of the current page or frame."""
         return self._current_browser().get_page_source()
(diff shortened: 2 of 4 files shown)


aaltat:
There is currently Get Location keyword, which (if I recall correctly) returns the currently active browser URL. The new keyword name should match the existing one. Or even better, we could enhance the existing keyword to be configurable to return active or all browser URL.

tx3stn:
@aaltat ahh yep, staying consistent with the `get location` keyword probably would have been a good idea.
I can update this PR to do that though.
I can add an argument to `get location` to return all locations (and update tests, doc strings etc).
Do you have any preference on the argument name? I was just thinking `all_locations`.
```python
    def get_location(self, all_locations=False):
        """Returns the current location."""
        if all_locations:
            <code here>
        else:
            return self._current_browser().get_current_url()
```

I can't see any tests using `get location` so I'll add ones for returning the single location and all.

pekkaklarck:
I'd say new Get Locations is better than configuring existing Get Location. Don't feel tol strongly about it, though.

tx3stn:
@pekkaklarck I'm perfectly happy to do either way. 
Is there anyone that might feel strongly? We could take a quick poll before I make the changes?

aaltat:
I don't feel strongly on the matter either. I don't want to keyword number to raise too hing, but in the other hand, having separate keyword might be cleaner to use and provides better readability. Perhaps the end usage matter in this case and having separate keyword is better.

tx3stn:
@aaltat @pekkaklarck I've updated this to `Get Locations`. 
Let me know if you have any thoughts/concerns.

aaltat:
Looks alright from mobile screen, have to look again from bigger screen. If you want, add your name here https://github.com/robotframework/Selenium2Library/blob/master/CHANGES.rst for credits

aaltat:
And opening a issue would be nice, it helps us to keep track on things better.

tx3stn:
@aaltat issue raised #697 
Updated `CHANGES.rst` to include this change and bumped the patch version. 
(I followed the precedent set in the changes file where new keywords had been added under patch increments. As it's new functionality if you want me to change this to a minor version bump that's fine).
Let me know if there's anything else.

Issue #699, opened 2016-12-06
Get parent webelement keyword


tx3stn:
I have an update that I have made locally to get an element's parent that I thought I'd see if anyone had any thoughts or objects before raising a PR.

```python
    def get_parent_webelement(self, locator):
        if isinstance(locator, WebElement):
            locator_element = locator
        else:
            locator_element = self.get_webelement(locator)
        return locator_element.find_element(by="xpath", value="parent::*")
```

I'm currently using this for pages with table like data with numerous input fields, not actually in a table, and I need to be sure the fields being updated are the correct ones, which I can do by getting the parent element (which in this case is the container for the whole row), and then looking for the child fields I need in the parent. 

Happy to raise the PR with tests for this if theres no objections / anyone else will find it useful.

aaltat:
Yes there has been and yes it could be useful. Example in #672. 

Because this is a duplicate, closing this issue.

aaltat:
Reopened because #702 discussion

aksenov007:
As @aaltat [requested](https://github.com/robotframework/SeleniumLibrary/pull/702#issuecomment-352174529), here is a possible acceptance test for this issue. It's the same as in the #702 and I think it is enough to describe how this keyword should work.

```
Get Parent Web Element Passing Element
    [Documentation]  Get parent web element of the element passed
    ${expected_parent}=     Get WebElement  //div[@id="div_id"]
    ${element}=    Get WebElement    //div[@id="div_id"]/a
    ${parent}=  Get Parent WebElement   ${element}
    Should be equal     ${parent}   ${expected_parent}

Get Parent Web Element Passing Locator
    [Documentation]  Get parent web element of the locator passed
    ${expected_parent}=     Get WebElement  //div[@id="div_id"]
    ${parent}=  Get Parent WebElement   //div[@id="div_id"]/a
    Should be equal     ${parent}   ${expected_parent}
```
The only case is worrying me: get parent of the `<html>` web element... Should I write a separate acceptance test for that case or not?

aaltat:
Closing due inactivity.

Issue #672, opened 2016-10-02
Get Child object of a given element


mzbedat:
Hi,
When using the Get WebElements keyword am getting a list of elements that fits my locator.
Each element contains a child object, how can I get the child one?

example:

``` html
<a class="yt-uix-sessionlink spf-link " data-sessionlink="itct=CGEQlDUYAyITCOeh4-OEvM8CFQKkHAodw6QFfSiOHjIGZy1oaWdoWg9GRXdoYXRfdG9fd2F0Y2g" href="/watch?v=5i-94at2SDA" aria-hidden="true">
<div class="yt-thumb video-thumb">
<span class="yt-thumb-simple">
<img height="110" width="196" data-ytimg="1" alt="" src="https://i.ytimg.com/vi/5i-94at2SDA/hqdefault.jpg?custom=true&w=196&h=110&stc=true&jpg444=true&jpgq=90&sp=68&sigh=YqxUUk-Z6HgOYPKBVVuMA9o1iaU" onload=";__ytRIL(this)">
</span>
</div>
<span class="**video-time**" aria-hidden="true">**3:34**</span>
/a>
```

I want to get the "video-time" inside a random a> element, is it possible?

aaltat:
Is there a reason why this does not work:

```
| ${text} = | Get Element Text | xpath=//*[@class="**video-time**"] |
```

mzbedat:
Yes.
I have many "video-time" elements in the page (YouTube Home Page for example) and I want to randomize one element and pick it up only if it's duration is less than a threshold...

for now I used:
`${duration_text}=      Selenium2Library.Get Text   xpath=.//ol[@class='item-section']/li/div/div[1]/div[2]/ul/li[${**random**}]/div/div/div/div[1]/a/span`

But I know that in robotframework-uiautomator library there is a Get Child keyword, so maybe it can be added to Selenium library as well.

emanlove:
The idea of a "Get Child" keyword is interesting. Working with the AngularJS Library and the underlying Protractor/WebDriverJS I see a lot of this (pseudo-code here)

```
child = parent.find_by_('locator')
...
grandchild = child.find_by('another_locator')
```

I have thought about adding an optional parameter to the Get Web Elements keyword passing it a root element from where to search from. I struggled with how we could do this with just a locator. Some sort of "Get Child" keyword might be a valid solution...

aaltat:
[...]

PR #702, opened 2016-12-07
Adds 'Get Parent WebElement' and 'Get Child WebElement' keywords

tx3stn:
Adds keyword to get the parent of the specified element. 
Accepts passing of a web element or a locator to find the web element, so I have added a test for each of those scenarios. 

Raised under issue #699 which was closed as a duplicate of #672 (but that deals with element children not parents). 
As #672 was asking for a `get child webelement` keyword I thought I'd add that in as well as they work in the same way.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -4,6 +4,8 @@ Release Notes
 1.8.1
 -----
 - Added 'Get Locations' keyword [thaffenden]
+- Added 'Get Parent Webelement' keyword [thaffenden]
+- Added 'Get Child Webelement' keyword [thaffenden]
 
 1.8.0
 -----
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -35,6 +35,22 @@ class _ElementKeywords(KeywordGroup):
         """
         return self._element_find(locator, False, True)
 
+    def get_parent_webelement(self, locator):
+        """Returns the parent WebElement for the given locator.
+
+        See `introduction` for details about locating elements.
+        """
+        locator_element = self.get_webelement(locator)
+        return locator_element.find_element(by="xpath", value="parent::*")
+
+    def get_child_webelements(self, locator):
+        """Returns the child WebElements for the given locator.
+
+        See `introduction` for details about locating elements.
+        """
+        locator_element = self.get_webelement(locator)
+        return locator_element.find_elements(by="xpath", value="child::*")
+
     # Public, element lookups
 
     def current_frame_contains(self, text, loglevel='INFO'):
--- a/test/acceptance/keywords/elements.robot
+++ b/test/acceptance/keywords/elements.robot
@@ -32,6 +32,32 @@ More Get Elements
     : FOR    ${checkbox}    IN    @{checkboxes}
     \    Checkbox Should Be Selected    ${checkbox}
 
+Get Parent Web Element Passing Element
+    [Documentation]  Get parent web element of the element passed
+    ${expected_parent}=     Get WebElement  //div[@id="div_id"]
+    ${element}=    Get WebElement    //div[@id="div_id"]/a
+    ${parent}=  Get Parent Webelement   ${element}
+    Should be equal     ${parent}   ${expected_parent}
+
+Get Parent Web Element Passing Locator
+    [Documentation]  Get parent web element of the locator passed
+    ${expected_parent}=     Get WebElement  //div[@id="div_id"]
+    ${parent}=  Get Parent Webelement   //div[@id="div_id"]/a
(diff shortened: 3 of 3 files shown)


aaltat:
Thank you from your contribution, but this was not exactly what I did have in mind with #672

Instead of doing this, the #672 describers to enhance the locator strategy so that it could also accept WebDirver [WebElement](http://selenium-python.readthedocs.io/api.html#selenium.webdriver.remote.webelement.WebElement) and it could be used to search the sibling element(s). 

Something like this: 
```
| ${element} = | Get WebElement | id=table_id |
| Page Should Contain Element | xpath=//td[@class="foobar"] | web_element=${element} | 
```
Then the `Page Should Contain Element` keyword would only look sibling elements under the element identified with id=table_id selector and not from the whole DOM.

Would you be willing to implement such functionality?

tx3stn:
Ok, well #699 wasn't a duplicate of #672 then. 
Also, this should solve what the OP's original request was in #672.
I'm happy to take a look at enhancing the locator strategy, but as the initial intention of my PR was to add a `Get Parent WebElement` keyword I'll do that under a separate pull request.

aaltat:
Yes the #699 was not duplicate, I did reopen the issue. But in the other hand I don't clearly see the use case for this type of keywords. Could you describe what actually is the use case in #699

mikesavino85:
It would be super cool if we could, as an optional argument, specify the child's tag or the inner text of the child.

aaltat:
@mikesavino85 how that is different from using xpath axes: http://www.w3schools.com/xml/xpath_axes.asp

Rhyr:
I'm just now running into a situation where this may actually prove useful, unless I'm missing another method somewhere...

In the page I'm currently automating, we have a data table and associated filters (using React); when inspecting the elements, the only visible difference between one checked and one unchecked is an "::after" within a lower div. That lead me to switch to CSS where I got "css=div[aria-hidden=false] * label > input:not(:checked)[type='checkbox']", which actually does work! Unfortunately, CSS does not support axes (aka selecting  label[/input[@type='checkbox']]) and XPath does not seem to find the 'checked' attribute anywhere!

My intention (once I figure out why this keyword is missing from my install despite updating..) is to Get WebElements using CSS, then :FOR through them with Get Parent WebElement.

As I said, I could be missing something, but figured I would share the use case in case it does indeed help validate the need for this!

aksenov007:
Hi @aaltat ! Can I open another PR for the issues #672 and #699 ? It seems that the author will not finish this one... I want to use existing code and add needed changes for that.

Comments about this request are scattered among the request and the issues, so and I want to check what I need except for the proposed code to close the issues?

As far as I understand there will be two new keywords:
* `Get Parent WebElement` (with the only parameter `locator`)
* `Get Child WebElement` (with two mandatory parameters - `parent_locator` and `child_locator`)

I consider the first one is already already done in this PR, do you agree?
The second one should be changed to fit the code snippet which @emanlove wrote [here](https://github.com/robotframework/SeleniumLibrary/issues/672#issuecomment-252485337), do you agree?

aaltat:
Pull request are always welcomed.  Could you first write few acceptance test in the issues so that we can agree on the design?

mikesavino85:
I know you've already done the work but as an answer to the question I had, I wrote a robot library to pull an element (defined by a number of things) -> all children elements (defined by a number of things). You could call it recursively so if you wanted to go element->child node->children of child node, it would pull them all back and put it in a dictionary for you. 

The reason I wanted this was for a specific use case but also I wanted to write it generically. Specifically, I had elements on the page that included a link and I wanted the data to persist in my test and relate back. In the actual use case, there were a number of locations that had links and I wanted to ensure the next page after clicking on the link corresponded to text in the original element. So, I selected some <tag> and the children a -> href and p (which contained the text I wanted to check). Then, I clicked the link and made sure an h1 tag contained the text in the original p element. 

The reasoning behind this was, obviously, we had templated pages generated from a db and had had some problems making sure they all lined up. We wanted to make sure the links worked as it was a critical path for the business and we obviously didn't want to hard code the checks because they could easily be updated from elsewhere in the application. 

So, I don't think I actually explained my question well, I came up with a solution that worked. Unfortunately, I was let go from that company and no longer have access to the code.

yjagdale (approved):

yjagdale:
Please resolve merge conflicts.

tx3stn:
Took a look at this just now in case it was just a simple rebase but look like files have moved around.

I fortunately don't have to use robot framework any more (and have not for some time, this PR is 2 years old) so have no need for this functionality any more.

Happy for it to be closed if nobody else familiar with the recent changes/current structure wants to pick it up and do the rebase.

aaltat:
PR can be always found, if there is a need, closing by requests.

Issue #706, opened 2016-12-07
Remove the WebDriverMonkeyPatches


aaltat:
It looks like the [WebDriverMonkeyPatches](https://github.com/robotframework/Selenium2Library/blob/master/src/Selenium2Library/webdrivermonkeypatches.py) was done to ease to conversion from SeleniumLibrary to Selenium2Library. Most likely we do not needed at all and we could use the methods from the selenium directly.

ombre42:
I think this is a good idea. But I am pretty sure some people out there are using these methods on `_current_browser()`. So it should be clearly communicated that `get_current_window_info`, etc. is going away.

aaltat:
I think some of the methods will stay, in some form or other, but not in there. At least the ones that will have direct counterpart in selenium side will disappear.

emanlove:
Is there a list of what we will still have and what we will lose by removing it?

aaltat:
Not yet, but we definitely need a one. The #708 changes the monkeypatched methods to their selenium/webdirver counter parts: 
 * get_window_handles() -> window_handles
 * get_current_url() -> current_url
 * get_page_source() -> page_source
 * get_title() -> title

Those where the easy ones. Most likely the rest will in stay in some form or other but it needs some thinking.  But before that thinking I need some sleep.

aaltat:
Done in #708

PR #708, opened 2016-12-07
Removing WebDriverMonkeyPatches methods

aaltat:
Removed get_current_url, get_page_source, get_title and
get_window_handles methods from WebDriverMonkeyPatches. Made the
library to the counterpart from the selenium side directly.
Added acceptance test for Get Title keyword and fixed the unit
test to use the selenium methods.

Todo: Remove the rest of the methods or implement methods somewhere else in the code base.


pekkaklarck:
On a quick look this looks good to be merged. Or do you want to remove other monkeys first?

emanlove:
Question, which I think @aaltat answered in his post under #706, is this backward compatible? Reading through the code it looks like it is but I want to verify that we are indeed still backward compatible.

test/unit/test_webdrivermonkeypatches.py
```diff
@@ -1,69 +0,0 @@
...
-
-    current_window_handle = HANDLE
-
-
-class WebDriverMonkeyPatchesTests(unittest.TestCase):
-
-    def test_window_info_values_are_strings(self):
-        driver = MockWebDriver()
-        when(driver).execute_script(SCRIPT).thenReturn(['id', 'name', 'title', 'url'])
-        info = driver.get_current_window_info()
```
emanlove, line 25: I am be missing something here but it looks like all these unit tests just test get_current_window_info method which I don't see being removed above. I can see removing these unit tests if get_current_window_info is removed but as it is not yet is this premature to remove these unit tests?

aaltat: True, will modify the file accordingly

aaltat: Added file back.

aaltat:
Are we happy and can this be merged?

pekkaklarck:
This doesn't remove all monkeys yet, right?

test/unit/test_webdrivermonkeypatches.py
aaltat: Renamed the file, because removed corresponding monkey patch. Also made the unit test work with the new implementation.

PR #710, opened 2016-12-11
Clean up acceptance test execution

pekkaklarck:
- Use external robotstatuschecker module instead of an old vendored copy.
- Add docstring to run_tests.py and use it in when showing usage on the cli.
- Don't read browser information from environment variables, only use the
  value given from the cli.
- Use robot.rebot_cli for generating logs/reports instead of running rebot
  on the system.
- Preserve test server logs.
- General cleanup here and there.


aaltat:
To be able to do #713, it would be easier to merge this in to the master and base the work on top of this PR.

PR #714, opened 2016-12-15
Clarified `get_matching_xpath_count` returns str

GLMeece:
Updated documentation to include the return type (string) of the XPath count.

Many users would assume that the "count" is an integer. However, since the count value is cast to a string upon return, this is not the case. Since this behavior has been in the code for quite awhile, likely as not users have coded around it. However, for accuracy's sake (as well as for those just learning RF).

### DIFF
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -607,7 +607,7 @@ return !element.dispatchEvent(evt);
     # Public, xpath
 
     def get_matching_xpath_count(self, xpath):
-        """Returns number of elements matching `xpath`
+        """Returns number of elements matching `xpath` as a string.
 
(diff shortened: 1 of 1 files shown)


aaltat:
Do you want to add your name to the CHANGES.rst and could you open a issue?

pekkaklarck:
Simple doc fixes could, in my opinion, be mergen also without issues.

Also, I agree int would be better type but changing would be backwards incompatible. Updating docs is a very good idea. Thanks for PR!

GLMeece:
For formality's sake, I have [created an issue](https://github.com/robotframework/Selenium2Library/issues/715), which includes Pekka's suggestion to have it optionally return an int as a keyword arg (still keeping a string return as the default).

aaltat:
Should we close this one and consentrate on #717

aaltat:
Closed to favor #735

Issue #713, opened 2016-12-14
Change the Travis to run differently the acceptance test


aaltat:
Because there was some unexpected problems with Travis and the accessing the encrypted information (the Sauce Labs credentials) [1] we need to change how we run acceptance tests from the original plan.

 * For each PR, run with Firefox (provided by Travis), selenium 2/3 and Python 2/3. 
 * Create nightly/weekly build to run from master and run by using Sauce Labs by using different browsers.

[1] https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml

aaltat:
We could define the `.travis.yaml`like this: 
```
language: python
sudo: required
dist: trusty
addons:
  apt:
    sources:
      - google-chrome
    packages:
      - google-chrome-stable

install:
  - pip install robotframework
  - pip install future
matrix:
  include:
    - python: "2.7" # For PR and cron
      env:
        - BROWSER=chrome
        - SELENIUM=2.53.6
    - python: "3.3" # For PR and cron
      env:
        - BROWSER=chrome
        - SELENIUM=2.53.6
    - python: "3.3" # For PR and cron
      env:
        - BROWSER=chrome
        - SELENIUM=3.0.2
    - python: "3.3" # For cron
      env:
        - BROWSER=firefox
        - SELENIUM=2.53.6
before_script:
  - "export DISPLAY=:99.0"
  - "sh -e /etc/init.d/xvfb start"
  - pip install selenium==$SELENIUM
script: python run_me.py $BROWSER
```
And then run test runner we could do something like this:
```
from __future__ import print_function
import os
import sys
from selenium import webdriver

print('start')
browser = sys.argv[1]
selenium = os.environ.get('SELENIUM')
print(browser)
print(selenium)
travis_event_type = os.environ.get('TRAVIS_EVENT_TYPE')

if travis_event_type == 'cron':
    print(os.environ.get('SAUCE_ACCESS_KEY'))
    print(os.environ.get('SAUCE_USERNAME'))
    # and other sstuff run with Sauce labs
elif travis_event_type != 'cron' and browser != 'chorme':
    # exit fast if trying to use other browser than chrome
    print(
        'No cron, exting test with selenium-{} and {} from Sauce'.format(
            selenium,
            browser
        )
    )
    sys.exit(0)
else:
    print('Running test for PR with {} using selenium'.format(browser, selenium))
    driver = webdriver.Chrome()
[...]

PR #716, opened 2016-12-17
Modified how Travis CI is used with the project.

aaltat:
Travis does not allow, for securirty reasons, PR buils to access
encrypted data. Project stores the Sacuce Labs credentials in
encrypted format in Travis settings and therefore builds from PR can
not use the browsers from Sauce Labs. Changed CI configuration and
run_tests.py to use Trvis Chrome browser for PR buils. Also configured
Travis corn to test from master periodically and to use different
browsers from the Sauce Labs.

Did take Python argparse in use and therefore passing arbitary
pybot and rebot arguments is not anymore possible. But now it is
possible also to use Sauce Labs from when running test from
local computer.


aaltat:
Perhaps for time being we should test selenium 3 with Python 2. Because we are not yet Python 3 compatible and therefore the build for selenium 3 fails.

test/run_tests.py
```diff
@@ -203,12 +205,13 @@ def sauce_credentials(sauceusername, saucekey):
...
 
     args = parser.parse_args()
-    browser = args.browser.lower()
-    if browser is not 'chrome' and env.TRAVIS_EVENT_TYPE in TRAVIS_PRS:
+    browser = args.browser.lower().strip()
+    if browser != 'chrome' and env.TRAVIS_EVENT_TYPE != 'cron':
         print(
-            'Can not run test with browser {} from SauceLabs\n'
+            'Can not run test with browser "{}" from SauceLabs\n'
             'SauceLabs can be used only when running with corn and from '
```
HelioGuilherme66: Typo. it should be:
`             'SauceLabs can be used only when running with cron and from '`

.travis.yml
```diff
@@ -7,6 +7,11 @@ addons:
       - google-chrome
     packages:
       - google-chrome-stable
+before_install:
+  - wget http://chromedriver.storage.googleapis.com/2.26/chromedriver_linux64.zip
```
HelioGuilherme66:
```
CDVERSION=`curl http://chromedriver.storage.googleapis.com/LATEST_RELEASE`
wget http://chromedriver.storage.googleapis.com/$CDVERSION/chromedriver_linux64.zip
```

test/run_tests.py
```diff
@@ -203,12 +205,13 @@ def sauce_credentials(sauceusername, saucekey):
...
-            'Selenium2Library master banch'.format(browser)
+            'Selenium2Library master banch, but your event type '
```
HelioGuilherme66: Typo. it should be:
`'Selenium2Library master branch, but your event type '`

test/run_tests.py
```diff
@@ -10,26 +10,43 @@
...
+`bin/sc -u YOUR_USERNAME -k YOUR_ACCESS_KEY -i localtunnel`
+
+More details and to downlaod Sauce Connect visit:
+https://wiki.saucelabs.com/display/DOCS/High+Availability+Sauce+Connect+Setup
 
 Examples:
 
   run_tests.py python chrome
-  run_tests.py jython c:\\Python35\\python.exe --test "Click element"
+  run_tests.py jython firefox --suite list
```
HelioGuilherme66: There was a change in the argument passing (mentioned on doc), but the actual code does not look like it has been changed.
Example should be:
`run_tests.py --interpreter jython firefox --suite list`

There are more fixes like on some typos. I'll stop the review here, but will test this file run_tests.py for better review.

HelioGuilherme66 (changes requested):
The major issue is with run_tests.py.
I'll do more testing on it.
If possible amend the text for PR there are some typos.

HelioGuilherme66:
The way it is coded, we can never run tests locally with Firefox, it aborts execution by not being allowed to run in Sauce Labs.

aaltat:
Ah, true. I know how to fix it. Thanks for pointing it out

HelioGuilherme66:
I now created a PR in @aaltat branch with my proposals [1](https://github.com/aaltat/robotframework-selenium2library/pull/1).

test/run_tests.py
aaltat: Done

test/run_tests.py
aaltat: Done

test/run_tests.py
aaltat: Done

.travis.yml
aaltat: Done

aaltat:
Updated `run_tests.py` to support other browsers that chrome locally

test/run_tests.py
```diff
@@ -36,9 +36,9 @@
 
 Examples:
 
-  run_tests.py python chrome
-  run_tests.py jython firefox --suite list
-  run_tests.py python chrome --scusername your_username --sckey account_key
+  run_tests.py chrome
+  run_tests.py --suite list --interpreter jython firefox
```
HelioGuilherme66: Better to have --suite at end because is a list.
run_tests.py --interpreter jython firefox --suite list

HelioGuilherme66 (approved):

HelioGuilherme66:
I still think you could merge my PR in your fork. There are some other minor improvements.
If you merge this one and #720 we will have green builds on Travis.

test/env.py
```diff
@@ -10,10 +10,11 @@
...
 
-TRAVIS = os.environ.get("TRAVIS", None)
-TRAVIS_JOB_NUMBER = os.environ.get("TRAVIS_JOB_NUMBER")
-SAUCE_USERNAME = os.environ.get("SAUCE_USERNAME")
-SAUCE_ACCESS_KEY = os.environ.get("SAUCE_ACCESS_KEY")
+TRAVIS = os.environ.get("TRAVIS", False)
+TRAVIS_EVENT_TYPE = os.environ.get("TRAVIS_EVENT_TYPE", None)
+TRAVIS_JOB_NUMBER = os.environ.get("TRAVIS_JOB_NUMBER", "localtunnel")
+SAUCE_USERNAME = os.environ.get("SAUCE_USERNAME", None)
+SAUCE_ACCESS_KEY = os.environ.get("SAUCE_ACCESS_KEY", None)
```
pekkaklarck: If these are only used by `run_tests.py`, we could consider moving all this there and could probably remove the whole `env.py`. Not really related to this PR, though.

test/run_tests.py
```diff
@@ -10,26 +10,43 @@
...
-  options:      Additional command line options passed to Robot Framework
+  browser:        Any browser supported by the library (e.g. `chrome`or
+                  `firefox`)
+  --interpreter:  Any Python interpreter supported by the library (e.g.
+                  `python`, `jython` or `c:\\Python27\\python.exe`). By
+                  default set to `python`.
+  --suite:        Selects the test suites by name.
+  --scusername:   Username to access Sauce Labs to order browsers when
+                  running test from local computer.
+  --sckey:        Access key for Sauce Labs account.
```
pekkaklarck: Doesn't `argparse` provide help text? No need to have it here in that case. The whole module docstring could be just a simple description of the script, possibly with few examples, that could then be passed to `argparse` as the general description. More comprehensive docs could then be moved to `README.(rst|md)` in this directory.

test/run_tests.py
```diff
@@ -69,12 +85,12 @@ def unit_tests():
         sys.exit(failures)
 
 
-def acceptance_tests(interpreter, browser, options):
+def acceptance_tests(interpreter, browser, suite, sauceusername, saucekey):
```
pekkaklarck: As a style comment, I'd prefer `sauce_username` and `sauce_key` as argument names. If they have default values like `None`, it would be good to add them to the signature.

test/run_tests.py
```diff
@@ -96,68 +112,110 @@ def http_server():
...
                for opt in ROBOT_OPTIONS]
-    if env.TRAVIS:
-        options.extend(get_travis_conf(browser))
-    command = runner + options + cli_options + [env.ACCEPTANCE_TEST_DIR]
-    log_start(command)
+    if sauceusername and saucekey:
+        options.extend(get_sauce_conf(browser, sauceusername, saucekey))
+    command = runner
+    if suite:
+        command += ['-s', suite]
```
pekkaklarck: `'--suite'` would be a little more explicit than `'-s'`.

test/run_tests.py
```diff
@@ -96,68 +112,110 @@ def http_server():
...
     options = [opt.format(browser=browser,
                           py_version=interpreter + sys.version[:3])
                for opt in ROBOT_OPTIONS]
-    if env.TRAVIS:
-        options.extend(get_travis_conf(browser))
-    command = runner + options + cli_options + [env.ACCEPTANCE_TEST_DIR]
-    log_start(command)
+    if sauceusername and saucekey:
+        options.extend(get_sauce_conf(browser, sauceusername, saucekey))
+    command = runner
```
pekkaklarck, line 133: This looks strange. Could either just rename `runner` to `command` originally or use something like

```python
command = runner + (['--suite', suite] if suite else [])
```

test/run_tests.py
```diff
@@ -96,68 +112,110 @@ def http_server():
...
-            )
-        ])
-    return conf
+    subprocess.call(
+        command, env=dict(os.environ, ROBOT_SYSLOG_FILE=syslog)
+    )
 
 
-def log_start(command_list):
+def log_start(command_list, sauceusername, saucekey):
```
pekkaklarck: Could just have the signature as `command_list, *hidden` and then later `for item in hidden`.

test/run_tests.py
```diff
@@ -96,68 +112,110 @@ def http_server():
...
             command = command.replace(hidden, '*' * len(hidden))
     print()
     print('Starting test execution with command:')
     print(command)
 
 
-def process_output(browser, cli_options):
+def get_sauce_conf(browser, sauceusername, saucekey):
+    if browser == 'chrome' and env.TRAVIS:
+        conf = []
```
pekkaklarck: Why not just `return []`? Would avoid the big `else` block below.

test/run_tests.py
```diff
@@ -96,68 +112,110 @@ def http_server():
...
+    return conf
+
+
+def process_output(browser, suite):
     print('Verifying results...')
     output = os.path.join(env.RESULTS_DIR, 'output.xml')
     robotstatuschecker.process_output(output, verbose=False)
     options = [opt.format(browser=browser) for opt in REBOT_OPTIONS]
+    if suite:
+        options += ['-s', suite]
```
pekkaklarck: This shouldn't be needed at all. Enough to filter based on suite when starting execution.

test/run_tests.py
```diff
@@ -96,68 +112,110 @@ def http_server():
...
 
 
+def sauce_credentials(sauceusername, saucekey):
+    if env.TRAVIS:
+        username = env.SAUCE_USERNAME
+        key = env.SAUCE_ACCESS_KEY
+    else:
+        username = sauceusername
+        key = saucekey
+    return username, key
```
pekkaklarck, line 189: Getting configs from `env` and from the command line is really confusing. Also, wouldn't it be better for command line options to have higher precedence if given?

test/run_tests.py
```diff
@@ -96,68 +112,110 @@ def http_server():
...
+
+
 if __name__ == '__main__':
-    if len(sys.argv) < 3 or '--help' in sys.argv:
-        sys.exit(__doc__)
+    parser = argparse.ArgumentParser(description='Library test runner')
+    parser.add_argument(
+        '--interpreter',
```
pekkaklarck, line 199: Should add a short option like `-I` for such a commonly needed option.

test/run_tests.py
```diff
@@ -96,68 +112,110 @@ def http_server():
...
+        help='Interpreter used run the test'
+    )
+    parser.add_argument('browser', help='Browser used in testing')
+    parser.add_argument(
+        '--suite',
+        '-s',
+        help='Selects the test suites by name.'
+    )
+    parser.add_argument(
+        '--scusername',
```
pekkaklarck: `--sauceusername` would be a lot better name. Could add a short option like `-U` if needed.

test/run_tests.py
```diff
@@ -96,68 +112,110 @@ def http_server():
...
+        help='Username to order browser from SaucuLabs'
+    )
+    parser.add_argument(
+        '--sckey',
```
pekkaklarck: `--saucekey` optionally with a short option like '-K` would be better.

test/run_tests.py
```diff
@@ -96,68 +112,110 @@ def http_server():
...
+    )
+
+    args = parser.parse_args()
+    browser = args.browser.lower().strip()
+    if env.TRAVIS and browser != 'chrome' and env.TRAVIS_EVENT_TYPE != 'cron':
+        print(
+            'Can not run test with browser "{}" from SauceLabs\n'
+            'SauceLabs can be used only when running with cron and from '
+            'Selenium2Library master branch, but your event type '
+            'was "{}"'.format(browser, env.TRAVIS_EVENT_TYPE)
```
pekkaklarck: Apparently only Chrome is supported here? Error should mention that too.

test/run_tests.py
```diff
@@ -96,68 +112,110 @@ def http_server():
...
 if __name__ == '__main__':
-    if len(sys.argv) < 3 or '--help' in sys.argv:
-        sys.exit(__doc__)
+    parser = argparse.ArgumentParser(description='Library test runner')
+    parser.add_argument(
+        '--interpreter',
+        default='python',
+        help='Interpreter used run the test'
+    )
+    parser.add_argument('browser', help='Browser used in testing')
```
pekkaklarck: If we only can run tests on Travis with Chrome, having that as the default browser would probably be a good idea. Could then make this an option like `--browser` with a short option `-B`.

test/run_tests.py
```diff
@@ -96,68 +112,110 @@ def http_server():
         serverlog.close()
 
 
-def execute_tests(interpreter, browser, cli_options):
+def execute_tests(interpreter, browser, suite, sauceusername, saucekey):
     runner = [interpreter, '-m', 'robot.run']
```
pekkaklarck: We should support also interpreter with spaces like `py -3` that's very handy on Windows. I guess that can already be given from the command line like `--interpreter "py -3"` but this code won't handle it. Could just use `interpreter.split()` to split the command. `shlex.split(interpreter)` would be safer but I doubt we need it here.

pekkaklarck (commented):
Code changes look good to me in general. I added some comments mainly related to the style. Few things I'd do related to this that aren't necessarily in the scope of this PR:

- Get rid off the `env.py` script entirely. Would be much cleaner to handle all configuration in runner scripts themselves. I doubt there's much acceptance and unit test runners share anyway.

- Add `README.(rst|md)` into the test directory.

- Handle interpreter like `py -3` (added separate comment about this).

test/run_tests.py
aaltat: Yes it does and it's actually quite good help system. Dividing this with readme file and argparse is good idea. At least the argparse help must be enhanced.

aaltat:
I am busy with Xmas preparation, so making the changes will go after Xmas.

Thanks from the comments (in a good way and not in a Finnish sartastic way) and happy holidays to all.

test/run_tests.py
aaltat: The `list` was a bad example, because it refers to `list.robot` file and not to Python list. Changed to example to `javascript`

test/env.py
aaltat: I agree. But  the PR is getting big and perhaps doing it in a separate PR is better and cleaner.

test/run_tests.py
aaltat: Done

test/run_tests.py
aaltat: Done

test/run_tests.py
aaltat: Moved to use https://docs.python.org/2.7/library/argparse.html#partial-parsing which allows to keep the old functional to give random Robot Framework arguments from command line. Therefore this comment is not anymore valid one.

test/run_tests.py
aaltat: Moved to use https://docs.python.org/2.7/library/argparse.html#partial-parsing which allows to keep the old functional to give random Robot Framework arguments from command line. Therefore this comment is not anymore valid one.

test/run_tests.py
aaltat: Done

test/run_tests.py
aaltat: Hmm, I am usually on favor for single return statement for each method. But perhaps this is clearer.

test/run_tests.py
aaltat: This might still valid, but the code has changed quite a look. Please take a new look and send comment if something is needed.

test/run_tests.py
aaltat: Agree. Made a temporally change and will be fixed better when removing `env.py` file in separate fix.

test/run_tests.py
aaltat: Done

test/run_tests.py
aaltat: Done

test/run_tests.py
aaltat: Done

test/run_tests.py
aaltat: Yes, when running tests with Travis against PR, then it is only possible to use Chrome. Improved text.

test/run_tests.py
aaltat: I kind a liked to this way. For me it makes the travis [matrix](https://github.com/robotframework/Selenium2Library/pull/716/files#diff-354f30a63fb0907d4ad57269548329e3R20) easier to read. Also it saves few letter for typing, now it can be used with: 
`run_tests.py chrome`
instead of:
`run_tests.py -B chrome`

test/run_tests.py
aaltat: Not done. I agree that this is useful but did not have time to code this in today.

aaltat:
* Did not yet do ` py -3` because of lack of time. But will do it later. 
 * Added note about the README to the issue #713 
 * Would like to remove the `env.py`with separate PR.

aaltat:
But as @HelioGuilherme66 pointed out and was reminded by this https://github.com/robotframework/robotframework/issues/2502, should we minimum supported Python 3 version by Robot Framework (Python 3.3), should we use something else or should we always live with the latest Python 3 available from Travis? Which, according to Travis documentation is is Python 3.5 (there is some support for Python 3.6 nightly builds, according the documentation)?

test/run_tests.py
aaltat: Done

aaltat:
Added `test/README.rst` and fixed `--interpreter "py -2"`problem.

`BUILD.rst` needs more work but perhaps in separate PR like the `env.py` file.

aaltat:
Decided to remove `env.py` because it was simple operation.

aaltat:
@pekkaklarck would you have time to take a look?

BUILD.rst
```diff
@@ -26,44 +26,10 @@ test/
 
 Unit and Acceptance Tests
 -------------------------
+To run tests give command::
+	python test/run_tests.py chomre
```
pekkaklarck: - chomre -> chrome
- If `run_tests.py` is executable, could just use `test/run_tests.py chrome`
- Should it rather be something generic like `<browser>` instead of `chrome`?

BUILD.rst
```diff
@@ -26,44 +26,10 @@ test/
...
-
-	# Run all tests with Python and Firefox
-	python test/run_tests.py python ff
-	# Run only test suite `javascript` with Jython and Internet Explorer
-	python test/run_tests.py jython ie -s javascript
-
-To run just the unit tests, run::
-
-	python test/run_unit_tests.py
+More details in test/README.rst
```
pekkaklarck: Formatted like `<test/README.rst>`_ this would create a link to the README file when rendered as HTML on GitHub.

test/README.rst
```diff
@@ -0,0 +1,64 @@
+Testing and debugging
+=====================
+Requirements
+------------
+Before running the test, install the dependencies::
+
+    pip install requirements.txt
+    pip install requirements-dev.txt
```
pekkaklarck: Shouldn't there be `-r` too?

test/README.rst
```diff
@@ -0,0 +1,64 @@
...
+  directory
+- A very simple `httpserver.py` which is used to serve the html for tests in
+  `resources/testserver`
+- A collection of simple html files under `resources/html` directory
+- Start-up scripts for executing the tests: `run_tests.py` and
+  `run_unit_tests.py`
+
+To run unit and acceptance tests, run::
+
+    python test/run_tests.py ff|ie|chrome
```
pekkaklarck: Perhaps just `test/run_tests.py <browser>`?

test/README.rst
```diff
@@ -0,0 +1,64 @@
...
+execution as a last arguments to the `run_tests.py` command. Example arguments
+like --test, --suite, --include and --exclude may be used to configure
+acceptance test execution.
+
+Examples::
+
+    python run_tests.py chrome
+    python run_tests.py --interpreter jython firefox --suite javascript
+    python run_tests.py chrome --sauceusername your_username --saucekey account_key --suite javascript
+    py -3 run_tests.py --interpreter "py -2" chrome --suite javascript
```
pekkaklarck: - I'd remove `python` and `py -3` from the beginning altogether.
- Using `--interpreter` hasn't been explained.

test/README.rst
```diff
@@ -0,0 +1,64 @@
...
+
+Using Sauce Labs for acceptance tests
+-------------------------------------
+
+When running test by using browser from Sauce labs, it is required that the
+Sauce Connect is used. The Sauce Connect allows the browser from Sauce Labs
+reach the acceptance test web server. The acceptance test uses tunnel with
+name `localtunnel` and therefore when establishing the Sauce Connect tunnel
+use the following command::
+    bin/sc -u YOUR_USERNAME -k YOUR_ACCESS_KEY -i localtunnel
```
pekkaklarck, line 67: There probably should be an empty row after `::`.

test/run_tests.py
```diff
@@ -85,79 +75,158 @@ def acceptance_tests(interpreter, browser, options):
...
+            'command. Example arguments like --test, --suite, --include and '
+            '--exclude may be used to configure acceptance test execution. '
+            'The Robot Framework command line arguments are not displayed in '
+            'the optional arguments list in the help.\n\n'
+            'Examples:\n\n'
+            'python run_tests.py chrome\n'
+            'run_tests.py --interpreter jython firefox --suite javascript\n'
+            'python run_tests.py chrome --sauceusername your_username '
+            '--saucekey account_key --suite javascript\n'
+            ' '
```
pekkaklarck: Should definitely format this using Python's triplequoted strings. Alternatively could use `__doc__` here.

test/README.rst
```diff
@@ -0,0 +1,64 @@
...
+afterwards using `robotstatuschecker` tool. The tool can be installed using
+`pip install robotstatuschecker` and more information about it can be found
+from: https://github.com/robotframework/statuschecker/. Notice that initially
+some tests fail.
+
+
+It is possible to pass Robot Framework command line arguments to the test
+execution as a last arguments to the `run_tests.py` command. Example arguments
+like --test, --suite, --include and --exclude may be used to configure
+acceptance test execution.
```
pekkaklarck: Do all RF's options work or can you only use some? Earlier versions had a restriction to use only some but is that fixed?

pekkaklarck (approved):
Looks good to me in general. I mainly briefly looked at the docs, though. Check my comments and merge when you think this is ready.

BUILD.rst
aaltat: Done

BUILD.rst
aaltat: Done

test/README.rst
aaltat: Yes and done

test/README.rst
aaltat: Done

test/README.rst
aaltat: In this version all goes, but that might not be always a good idea. Enhanced documentation to recommend few arguments.

test/README.rst
aaltat: Done

test/README.rst
aaltat: Done

test/run_tests.py
aaltat: The `argparse` is making some things hard to present, but perhaps this is more clearer way. Although I do not like the back slash as line continuation, but it is the lesser evil this time.

pekkaklarck (commented):
I'd add module docstring to run_tests.py and then pass description and epilog to argparse as `__doc__.spilitlines()[0]` and `'\n'.join(__doc__.splitlines()[2:])`, respectively. Scripts having docstring is generally nice and it would be handy to reuse it with argparse.

aaltat:
Done

test/README.rst
```diff
@@ -0,0 +1,85 @@
...
+Running test with different interpreter
+---------------------------------------
+
+By default the interpreter is set to `python` and it depends on the  operating
+system settings, which python is used by default. It is possible to run test
+by using different interpreter by using the `--interpreter` command line
+argument.
+
+Starting from version 2.0 onwards the Selenium2Library is tested by using 
+Python 2 and 3. Other interpreters are not test by the development team.
```
HelioGuilherme66: Verb and not name. It should be: "Other interpreters are not test**ed** by the development team."

test/README.rst
```diff
@@ -0,0 +1,85 @@
...
+
+
+Robot Framework command line arguments
+--------------------------------------
+
+It is possible to pass Robot Framework command line arguments to the test
+execution as a last arguments to the `run_tests.py` command. It is recommended
```
HelioGuilherme66: Remove single article. It should be: "... execution as ~~a~~ last arguments to the `run_tests.py` command."

test/run_tests.py
```diff
@@ -1,35 +1,43 @@
...
-Arguments:
+More details and to downlaod Sauce Connect visit:
+https://wiki.saucelabs.com/display/DOCS/High+Availability+Sauce+Connect+Setup
 
-  interpreter:  Any Python interpreter supported by the library (e.g. `python`,
-                `jython`, `c:\\Python27\\python.exe`)
-  browser:      Any browser supported by the library (e.g. `chrome`, `firefox`)
-  options:      Additional command line options passed to Robot Framework
+It is possible to pass Robot Framework command line arguments to the test
+execution as a last arguments to the `run_tests.py` command. It is
```
HelioGuilherme66: Same correction as [in comment](https://github.com/robotframework/Selenium2Library/pull/716/files#r95268607)

aaltat: Done

test/README.rst
aaltat: Done

test/README.rst
aaltat: Done

PR #718, opened 2016-12-18
Fixes installer for both Python 2.7 and 3.x.

HelioGuilherme66:
Replaces problematic execfile by exec.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,8 @@ from setuptools import setup
 
 CURDIR = dirname(abspath(__file__))
 
-execfile(join(CURDIR, 'src', 'Selenium2Library', 'version.py'))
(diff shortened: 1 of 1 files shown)


pekkaklarck:
Why would be `compile` needed here? AFAIK this ought to be enough:

```python
with open(join(CURDIR, 'src', 'Selenium2Library', 'version.py')) as f:
    exec(f.read())
```

HelioGuilherme66:
Modified code as @pekkaklarck suggestion.

aaltat:
Would you mind rebase with the current master. I would like to see that the change works with Python 2 and 3 with Travis.

HelioGuilherme66:
Rebased (tests pass except on python 3.3).

PR #720, opened 2016-12-19
Fix unit tests on Python3 and keep them OK on Python2.7

HelioGuilherme66:
- Formatting of exceptions messages
- Definition of unicode method on Python 3.x

### DIFF
--- a/src/Selenium2Library/keywords/_browsermanagement.py
+++ b/src/Selenium2Library/keywords/_browsermanagement.py
@@ -12,6 +12,10 @@ from Selenium2Library.locators import WindowManager
 
 from .keywordgroup import KeywordGroup
 
+try:
+    unicode
+except NameError:
+    unicode = str
 
 ROOT_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
 FIREFOX_PROFILE_DIR = os.path.join(ROOT_DIR, 'resources', 'firefoxprofile')
--- a/src/Selenium2Library/utils/__init__.py
+++ b/src/Selenium2Library/utils/__init__.py
@@ -2,6 +2,11 @@ from . import events
 from .browsercache import BrowserCache
 from .librarylistener import LibraryListener
 
+try:
+    unicode
+except NameError:
+    unicode = str
+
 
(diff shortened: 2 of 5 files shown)


HelioGuilherme66:
Please merge this one to have Travis builds green again. Later we can optimize the "unicode" in Python 3.x.

pekkaklarck:
The approach to handle unicode is not correct. It perhaps makes unit tests pass but it won't actually work correctly. Not sure yet what's the best solution to this, but committing broken implememtation wouldn't make sense.

Actual unit test fixes could be done already now, but converting exceptions to strings should be done using `str()`. Should actually use `unicode()` to support non-ASCII on Py2, but on unit tests we could just assume everything to be ASCII.

alexandrul:
This approach would be ok?
https://github.com/oxplot/fysom/issues/1
http://www.rfk.id.au/blog/entry/preparing-pyenchant-for-python-3/

aaltat:
Thanks from the link, need to read it with more thought.

HelioGuilherme66:
@pekkaklarck Please see if you agree with the TODAY's change to use `robot.utils` `PY3` and `unic()`.

pekkaklarck:
No, I don't agree with them:

- Using `PY3` directly would create a dependency to RF 3.0 or newer. 
- Should never define `unicode` as a custom method because it won't work with `isinstance`. `unicode = str` is better.
- Imports are not PEP-8 compatible.
- `"{}".format(e)` is a strange way to format an exception. `str(e)` is totally fine in tests. In production code should think about the possibility that `e` may contain non-ASCII characters.

HelioGuilherme66:
I have run out of ideas. Please post solution for the unicode problem.

pekkaklarck:
I'm not sure yet what's the best solution for handling unicode/str/bytes stuff in this library. The solution may in fact be different in different places, most importantly different in the real library code and in unit tests. I hope I have time for Se2Lib next week ,but I'll start from the architecture changes @aaltat and I have discussed.

aaltat:
This issue is already fixed in master, therefore closing the PR.

Issue #719, opened 2016-12-19
Consider string `False` to be false similarly as Robot Framework


aaltat:
RF convenience methods for handling booleans, example 
```
| Some Keyword | False |
```
Is evaluated as False in Robot Framework standard libraries, although the argument is string and according Python it should evaluate as True. Changing the S2L to behave in similar fashion as the Robot Framework standard libraries could be useful.

More details in http://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Boolean%20arguments and some discussion in #717

aaltat:
This is done is #845

aaltat:
The doc update is missing, reopening.

pekkaklarck:
Here's an example from a similar docs from XML library:
http://robotframework.org/robotframework/latest/libraries/XML.html#Boolean%20arguments

The main differences in S2L compared to XML are that also string `NONE` is considered false (this will change in RF/XML too) and that same logic applies also to `None` argument, not only to booleans.

pekkaklarck:
I don't see any reasonable use case where this could cause backwards incompatibility problems.

PR #735, opened 2017-01-22
Adds possibility to define get matching xpath count kw return type

### LINKED ISSUE #715: Support returning int from `Get Matching Xpath Count`
This "feature" has been present AFAIK since about the first check-in. I tripped up over this when attempting to add a value to the return and got a type error.

Two suggested approaches:

1. Change the documentation, per [this pull request](https://github.com/robotframework/Selenium2Library/pull/714).
2. As an optional keyword argument, allow for an `int` return (default would remain `str`). I have not included a code example/pull for that.

aaltat:
Fixes #715

### DIFF
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -622,8 +622,11 @@ return !element.dispatchEvent(evt);
 
     # Public, xpath
 
-    def get_matching_xpath_count(self, xpath):
+    def get_matching_xpath_count(self, xpath, return_str=True):
         """Returns number of elements matching `xpath`
+
+        The default return type is `str`but it can changed to `int` by setting
+        the ``return_str`` argument to Python False.
 
         One should not use the xpath= prefix for 'xpath'. XPath is assumed.
 
@@ -636,7 +639,7 @@ return !element.dispatchEvent(evt);
         `Xpath Should Match X Times`.
         """
         count = len(self._element_find("xpath=" + xpath, False, False))
-        return str(count)
+        return str(count) if return_str else count
 
     def xpath_should_match_x_times(self, xpath, expected_xpath_count, message='', loglevel='INFO'):
         """Verifies that the page contains the given number of elements located by the given `xpath`.
--- a/src/Selenium2Library/keywords/_element.py
+++ b/src/Selenium2Library/keywords/_element.py
@@ -625,7 +625,7 @@ return !element.dispatchEvent(evt);
     def get_matching_xpath_count(self, xpath, return_str=True):
         """Returns number of elements matching `xpath`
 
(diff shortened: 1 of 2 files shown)


src/Selenium2Library/keywords/_element.py
```diff
@@ -622,9 +622,12 @@ def page_should_not_contain_image(self, locator, message='', loglevel='INFO'):
 
     # Public, xpath
 
-    def get_matching_xpath_count(self, xpath):
+    def get_matching_xpath_count(self, xpath, return_str=True):
         """Returns number of elements matching `xpath`
 
+        The default return type is `str`but it can changed to `int` by setting
```
kormbrek-rally: missing space after \`str\`

src/Selenium2Library/keywords/_element.py
```diff
@@ -636,7 +639,7 @@ def get_matching_xpath_count(self, xpath):
         `Xpath Should Match X Times`.
         """
         count = len(self._element_find("xpath=" + xpath, False, False))
-        return str(count)
+        return str(count) if return_str else count
```
kormbrek-rally, line 642: IMHO we should follow [DateTime's](https://github.com/robotframework/robotframework/blob/master/src/robot/libraries/DateTime.py) lead and use a util method to make sure that `str('false')` is treated as `False`, etc.
```
from robot.utils import is_truthy
...
```

test/acceptance/keywords/elements.robot
```diff
@@ -61,6 +62,12 @@ Get Matching XPath Count
     [Documentation]    Get Matching XPath Count
     ${count}=    Get Matching XPath Count    //a
     Should Be Equal    ${count}    19
+    ${count}=    Get Matching XPath Count    //a    ${True}
+    Should Be Equal    ${count}    19
+    Should Be String    ${count}
+    ${count}=    Get Matching XPath Count    //a    ${False}
+    Should Be Equal As Numbers    ${count}    19
```
kormbrek-rally: I think you can just do a direct comparison to an `int` here...
```
Should Be Equal    ${count}    ${19}
```

src/Selenium2Library/keywords/_element.py
aaltat: I agree, but would like to do it same time for all the keywords using booleans. There is #719 for this problem.

test/acceptance/keywords/elements.robot
aaltat: Good point, I will fix it.

src/Selenium2Library/keywords/_element.py
aaltat: Good catch, I will fix it.

PR #744, opened 2017-02-01
Adds `future` dependency.

HelioGuilherme66:


aaltat:
I am not sure do we need this in the requirements file. What problem you are trying to solve with this PR?

HelioGuilherme66:
The `future` (or [futurize](http://python-future.org/futurize.html)) package is to have python 2.7 and 3.x compatibility. There are some methods useful in python 3.x, like the `with_metaclass`, there are types definitions to have simultaneous compatibility. The first important code it brought was the print() function to python 2.6. We can use `future`, or we can get only the methods we need from it (the files requiring it are just a few).

pekkaklarck:
I thought we already discussed what we specifically don't want to use `future`, `six` or any other such module at least as an extra external dependency. I doubt we in the end need them at all, but if we do, we can either vendor a whole module or just copy functions we need. For example, `with_metaclass` is just an few lines of (very interesting) code.

PR #745, opened 2017-02-01
Removes python 2.6 support from build_dist.py.

HelioGuilherme66:
Adds python 3.x support to build_dist.py
Fixes print().


aaltat:
I think we should remove the whole `build_dist.py` because we are taking the https://github.com/robotframework/rellu in use.

pekkaklarck:
Yeah, I already removed it locally.

PR #753, opened 2017-02-01
Convert files with dos2unix

aaltat:
Only run dos2unix tool to these files, no other changes where made.


HelioGuilherme66 (approved):
Long live Unix line-endings! ;)

Issue #751, opened 2017-02-01
Add supported RF version to Travis


aaltat:
Make Travis run test with RF 2.8, 2.9 and 3.0. Latest from all versions

aaltat:
Done in #754

PR #754, opened 2017-02-01
Add RF 3.0.1, 2.9.2 and 2.8.7 to Travis

aaltat:
We must also test with RF version we support. Updated Travis to
run test for different Robot Framework versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,6 +48,6 @@ before_script:
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
   - pip install selenium==$SELENIUM
-  - pip install selenium==$ROBOTFRAMEWORK
+  - pip install robotframework==$ROBOTFRAMEWORK
 script:
   - python test/run_tests.py $BROWSER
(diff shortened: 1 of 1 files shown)


aaltat:
Fixes #751

.travis.yml
```diff
@@ -23,21 +23,31 @@ matrix:
...
     - python: "3.3" # For cron only
       env:
         - BROWSER=firefox
         - SELENIUM=2.53.6
+        - ROBOTFRAMEWORK=3.0.1
 before_script:
   - "export DISPLAY=:99.0"
   - "sh -e /etc/init.d/xvfb start"
   - pip install selenium==$SELENIUM
+  - pip install selenium==$ROBOTFRAMEWORK
```
HelioGuilherme66: This is wrong. It should be:
`pip install robotframework==$ROBOTFRAMEWORK`

HelioGuilherme66 (changes requested):
We should consider using selenium==2.53.6 with firefox-esr-latest and not with chrome. I suspect that newer chrome does not work well with 2.53.6.

.travis.yml
aaltat: Should not play with my hobbies so late in the night and therefore make stupid mistakes like this. Good catch, will fix it.

aaltat: Done

Issue #750, opened 2017-02-01
Not able to switch to another Tab 


manju4ever:
I'm facing difficulty to switch to another tab. So using  **Select Window    window_title**  it's not working.
How do we switch the driver to point to another tab, is there a right way to do this ?

aaltat:
Sometimes opening a new window takes fare bit of time and therefore the `Select Window` fails. This is specially true if you click element, which opens a new window, because the click doesn't know anything about the new window and that it should wait such event. 

As a best solution would be to make keyword that actually waits and select the window, example:
```
| Wait And Select Window | window_title | timeout=10s |
```
Would you like to build such keyword?

As a workaround, using this should work:
```
| Wait Until Keyword Succeeds | 10 | 2 | Select Window | window_title |
```
Please note that I did not check the keyword syntax and it might contain errors?

manju4ever:
Sure thank you @aaltat , will definitely check it out and close it.

aaltat:
Did you get the problem solved?

manju4ever:
We used the select window by url and it worked like a charm.

Select Window    url="http://something.com"

Strategy	Example	Description
title	Select Window | title=My Document	Matches by window title
name	Select Window | name=${name}	Matches by window javascript name
url	Select Window | url=http://google.com	Matches by window's current URL

PR #756, opened 2017-02-07
Makes acceptance test more stable with Travis

### LINKED ISSUE #734: There are random failures with acceptance test when running in  CI + Chrome
In half of the test runs in CI, there is a failure. Most of the failures seems to be related on opening new window. The problem can not be reproduced when running tests on a local PC.

Think how to resolve the problem and take corrective actions.

aaltat:
There is quite often a random failures with test that open new window.
It seem that with Travis is takes little bit of time to open a window
and this fix wait the windows to open before test is performed.

Partially fixes #734

### DIFF
--- a/test/acceptance/keywords/navigation.robot
+++ b/test/acceptance/keywords/navigation.robot
@@ -101,13 +101,13 @@ Target Opens in New Window
     [Documentation]    Target Opens in New Window
     Cannot Be Executed in IE
     Click Link    Target opens in new window
-    Wait Until Keyword Succeeds    5    1    Wait Untill Window Is Open
+    Wait Until Keyword Succeeds    5    1    Wait Until Window Is Open
     Select Window    ${INDEX TITLE}
     Verify Location Is "index.html"
     [Teardown]    Run Keyword If Test Passed    Run Keywords    Close Window    Select Window
 
 *** Keywords ***
-Wait Untill Window Is Open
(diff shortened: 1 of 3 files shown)


test/acceptance/keywords/navigation.robot
```diff
@@ -101,6 +101,13 @@ Target Opens in New Window
...
+    Wait Until Keyword Succeeds    5    1    Wait Untill Window Is Open
     Select Window    ${INDEX TITLE}
     Verify Location Is "index.html"
     [Teardown]    Run Keyword If Test Passed    Run Keywords    Close Window    Select Window
+
+*** Keywords ***
+Wait Untill Window Is Open
+    ${titles} =    Get Window Titles
+    ${status} =    Evaluate    len(${titles}) > 1
+    Should Be True    ${status}
```
pekkaklarck: `Wait Until Window Is Open` sounds like a keyword that could be in the library itself. Also, there's only one `L` in `Until`.

aaltat:
There has been some complaints from users too, example #750 and we should build such `Wait*` keyword for window selection too.

I will fix the typo later today.

aaltat:
Created #757 to create the new keyword

test/acceptance/keywords/navigation.robot
aaltat: Done

PR #765, opened 2017-02-18
Removes unused import from browsermanagement.py

aaltat:


aaltat:
Closed because #770 fixes this one too.

PR #767, opened 2017-02-25
Table should not contain

eL-Prova:
Added option for check for should not contain value in table


aaltat:
Closing due inactivity.

PR #770, opened 2017-03-07
Selenium2Library new architecture

aaltat:
This PR removes the library multi inheritance madness. This change result cleaner code base, which eases the maintenance and hopefully eases accepting the new enhancement to the library. 

The PR also removes the meta classes and the automatic keyword decoration in the keywordgroup object. Also the PR introduces a new context object(which is the library itself). The context object holds information  from the different WebDriver instances and allows methods to register and fetch the different WebDriver instances.


aaltat:
This PR is not yet ready to merged, it has two unit test to be fixed and half of the acceptance test do fail. But I want to store the code also somewhere else than in my laptop and I want other also to take a look and give their input on the matter.

src/Selenium2Library/base.py
```diff
@@ -0,0 +1,25 @@
...
+
+    def log(self, msg, level='INFO', html=False):
+        logger.write(msg, level, html)
+
+    def warn(self, msg, html=False):
+        logger.warn(msg, html)
+
+    def _log(self, message, level='INFO', html=False):
+        level = level.upper()
+        if (level == 'INFO'):
```
pekkaklarck: Parens aren't needed! Python ain't Java! =)

src/Selenium2Library/keywords/cookie.py
```diff
@@ -1,46 +1,58 @@
-from .keywordgroup import KeywordGroup
+from Selenium2Library.robotlibcore import keyword
```
pekkaklarck: I think it would be better to use `from .robotlibrarycore import keyword` here. Makes it easier to change the library name if/when we decide to do it.

src/Selenium2Library/__init__.py
```diff
@@ -184,10 +183,42 @@ def __init__(self,
...
         self.ROBOT_LIBRARY_LISTENER = LibraryListener()
+
+    def run_keyword(self, name, args, kwargs):
+        try:
+            return DynamicCore.run_keyword(self, name, args, kwargs)
+        except Exception as e:
+            print 'run failure here'
+            raise e
+
+    def current_browser(self):
```
pekkaklarck: This should be a property and possibly named simply `browser`. Makes usage from library components nicer when you can say `context.browser.something()` and not `context.current_browser().something()`. If the old architecture has this method, we should leave it for backwards compatibility and consider [deprecating it](https://docs.python.org/2/library/warnings.html#warnings.warn).

src/Selenium2Library/keywords/alert.py
```diff
@@ -1,21 +1,23 @@
 import time
+
 from selenium.common.exceptions import WebDriverException
 from selenium.webdriver.support.ui import WebDriverWait
 from selenium.webdriver.support import expected_conditions as EC
 
-from .keywordgroup import KeywordGroup
+from Selenium2Library.base import Base
+from Selenium2Library.robotlibcore import keyword
```
pekkaklarck: Already commented once, but I think `from .module import something` is better than `from Selenium2Library.module import something`. The former has less noise and require no changes if/when we change the library name.

src/Selenium2Library/keywords/alert.py
```diff
@@ -1,21 +1,23 @@
...
 
 
-class AlertKeywords(KeywordGroup):
+class AlertKeywords(Base):
 
     __ACCEPT_ALERT = 'accept'
     __DISMISS_ALERT = 'dismiss'
```
pekkaklarck: This can nowadays be normal constants named just like `ACCEPT_ALERT`.

src/Selenium2Library/keywords/cookie.py
```diff
@@ -1,46 +1,58 @@
...
+class CookieKeywords(object):
 
+    def __init__(self, ctx):
+        self.ctx = ctx
+
+    @keyword
     def delete_all_cookies(self):
         """Deletes all cookies."""
-        self._current_browser().delete_all_cookies()
+        self.ctx.current_browser().delete_all_cookies()
```
pekkaklarck: Good example where `ctx.browser.delete_all_cookies()` would be cleaner than the current `ctx.current_browser().delete_all_cookies()`.

src/Selenium2Library/keywords/element.py
```diff
@@ -407,10 +439,12 @@ def drag_and_drop(self, source, target):
         Examples:
         | Drag And Drop | elem1 | elem2 | # Move elem1 over elem2. |
         """
-        src_elem = self._element_find(source, True, True)
-        trg_elem = self._element_find(target, True, True)
-        ActionChains(self._current_browser()).drag_and_drop(src_elem, trg_elem).perform()
+        src_elem = self.element_find(source, True, True)
+        trg_elem = self.element_find(target, True, True)
```
pekkaklarck: General comments not related to the architecture change:

- `element_find()` clearly needs default values so that `True, True` doesn't need to be repeated.
- In general Boolean arguments should be given in named format like `something=True` to make it easier to understand the calling code.
- This module is rather big. Should consider splitting it into smaller pieces. Element finding logic could perhaps be moved into context.

Because the above aren't really related to the architecture change, it's probably better not to do it as part of this PR. Could perhaps just add some TODOs.

src/Selenium2Library/keywords/formelement.py
```diff
@@ -1,51 +1,58 @@
...
 
 
-class FormElementKeywords(KeywordGroup):
+class FormElementKeywords(Base):
 
-    # Public, form
+    def __init__(self, ctx):
+        Base.__init__(self)
+        self.ctx = ctx
+        self.element_find = ElementKeywords(ctx).element_find
```
pekkaklarck: This works, but moving the element finding logic into the context or into the base class would probably be cleaner.

pekkaklarck (changes requested):
Looks great in general!

pekkaklarck:
One missing change is removing the `decorator` module dependency from `setup.py`. =)

src/Selenium2Library/keywords/cookie.py
```diff
@@ -1,46 +1,58 @@
-from .keywordgroup import KeywordGroup
+from Selenium2Library.robotlibcore import keyword
 
 
-class CookieKeywords(KeywordGroup):
+class CookieKeywords(object):
```
AlexanderWhitestone: Just for my own understanding:
Why does this class inherit from object, but many other keyword classes inherit from Base?
I see that it isn't necessary since there is no logging for these keywords.  I just want to understand the principal behind the choice.
Is it simplicity over uniformity?

pekkaklarck: Good point. I'd say all lib components should extend same base class. Consistency FTW!

src/Selenium2Library/base.py
aaltat: Copy paste from the old library... Will fix it.

src/Selenium2Library/keywords/cookie.py
aaltat: Good point, consistency rules in this case

aaltat:
I think I will fix the unit test next. The failing unit test needs some thinking because the things that they are also testing are not anymore there. Perhaps I need refactor those parts out to a separate unit test for the browser cache.

src/Selenium2Library/keywords/cookie.py
AlexanderWhitestone: Thanks, glad to help.

src/Selenium2Library/__init__.py
ombre42: A hundred times yes. Please leave `_current_browser()` for a release cycle as it will break a whole lot of outside code. I suggest emitting a warning every time it is invoked. Or perhaps only once.
There are many aspects to Selenium S2L does not expose so a lot of people depend on this private method. In the mean time you can have `_current_browser()` call `browser` or `current_browser()`.

I also agree it would be better suited as an attribute rather than a method.

aaltat: Thank you from reminding this one. We will create a `_current_browser()` for backwards compatibility.  And most likely we must also create `_cache.register()` for the same reason. Now it is `cache.register()`

aaltat:
Only 4 acceptance test fails. Not far to go.

aaltat:
Run on failure fixed. Not totally sure is the this https://github.com/robotframework/Selenium2Library/pull/770/files#diff-d9fdca8cec320726d945c028908fd190R213 best approach, but works nicely. Three more failures to be fixed and then I should refactor the [elementfinder.py](https://github.com/robotframework/Selenium2Library/blob/master/src/Selenium2Library/locators/elementfinder.py)

aaltat:
Also I do not quite understand why this if is here: https://github.com/robotframework/Selenium2Library/blob/master/src/Selenium2Library/keywords/runonfailure.py#L64

aaltat:
All acceptance test now pass. Now it is easier to refactor [elementfinder.py](https://github.com/robotframework/Selenium2Library/blob/master/src/Selenium2Library/locators/elementfinder.py)

aaltat:
This PR is ready for review, it would be good if someone would have time read the code and make sure that changed are good ones. And I know that this PR is quite huge one and not in average sized PR. 

I also think that the [elementfinder.py](https://github.com/robotframework/Selenium2Library/blob/master/src/Selenium2Library/locators/elementfinder.py) refactoring should be done in a separate PR, because this already is a huge PR.

src/Selenium2Library/__init__.py
```diff
@@ -1,25 +1,25 @@
-from .keywords import *
+import warnings
+
+from .utils import BrowserCache
```
pekkaklarck: Alphabetical ordering of imports is recommended by PEP-8. This particular import should be combined with the other `from .utils import` line a little lower.

src/Selenium2Library/__init__.py
```diff
@@ -184,10 +183,54 @@ def __init__(self,
...
         self.set_selenium_timeout(timeout)
         self.set_selenium_implicit_wait(implicit_wait)
         self.register_keyword_to_run_on_failure(run_on_failure)
         self.ROBOT_LIBRARY_LISTENER = LibraryListener()
+
+    def run_keyword(self, name, args, kwargs):
+        try:
+            return DynamicCore.run_keyword(self, name, args, kwargs)
+        except Exception:
+            RunOnFailureKeywords(self)._run_on_failure()
```
pekkaklarck: Using `_private` method is ugly. Is this still work-in-progress? In that case adding `TODO` would be OK at this phase.

src/Selenium2Library/base.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+class Base(object):
+
+    def __init__(self, ctx):
+        self.ctx = ctx
+        self.element_finder = ElementFinder()
+        self._speed_in_secs = 0.0
+        self._timeout_in_secs = 5.0
+        self._implicit_wait_in_secs = 5.0
```
pekkaklarck: Is the plan that different instances can have different values for these times? If not, then they shouldn't be instance attributes. This is important to check!

The `_secs` postfix sounds a little unnecessary too, but I guess it could help separating these from possible millisecond values.

src/Selenium2Library/keywords/alert.py
```diff
@@ -1,21 +1,23 @@
 import time
+
 from selenium.common.exceptions import WebDriverException
 from selenium.webdriver.support.ui import WebDriverWait
 from selenium.webdriver.support import expected_conditions as EC
 
-from .keywordgroup import KeywordGroup
-
+from Selenium2Library.base import Base
+from Selenium2Library.robotlibcore import keyword
```
pekkaklarck, line 8: Should we use `..base` instead of `Selenium2Library.base`? Robot uses the latter but the former would make the library name change easier. This isn't a big deal in general and definitely not in the context of this PR.

src/Selenium2Library/keywords/browsermanagement.py
```diff
@@ -55,18 +50,24 @@ def close_all_browsers(self):
...
-            self._cache.close()
-
-    def open_browser(self, url, browser='firefox', alias=None,remote_url=False,
-                desired_capabilities=None,ff_profile_dir=None):
+        if self.browsers.current:
+            self.debug(
+                'Closing browser with session id {}'.format(
+                    self.browsers.current.session_id
+                )
+            )
```
pekkaklarck: This is a bit verbose coding style, especially for something like debugging. Instead of using 5 lines I'd write it on 2, for example, like this:

```python
self.debug('Closing browser with session id {}'
           .format(self.browsers.current.session_id))
```

As an unrelated note, I'd prefer proper sentences with `.` at the end. Again, not part of this PR.

src/Selenium2Library/keywords/browsermanagement.py
```diff
@@ -556,8 +581,7 @@ def _make_browser(self, browser_name, desired_capabilities=None,
 
         return browser
 
-
-    def _make_ff(self , remote , desired_capabilites , profile_dir):
+    def _make_ff(self, remote, desired_capabilites, profile_dir):
```
pekkaklarck, line 574: +1 for removing unnecessary and PEP-8 incompatible whitespace. Separate commit in master would have been better, though. I do this all the time too and I wouldn't revert these changes.

src/Selenium2Library/keywords/element.py
```diff
@@ -83,7 +87,8 @@ def element_should_not_contain(self, locator, expected, message=''):
         Key attributes for arbitrary elements are `id` and `name`. See
         `Element Should Contain` for more details.
         """
-        self._info("Verifying element '%s' does not contain text '%s'."
+        self.info(
+            "Verifying element '%s' does not contain text '%s'."
                    % (locator, expected))
```
pekkaklarck: Inconsistent indentation. Why not to have `info("Verifying ...` like earlier?

src/Selenium2Library/keywords/element.py
```diff
@@ -141,26 +152,30 @@ def locator_should_match_x_times(self, locator, expected_locator_count, message=
         See `Page Should Contain Element` for explanation about `message` and
         `loglevel` arguments.
         """
-        actual_locator_count = len(self._element_find(locator, False, False))
+        actual_locator_count = len(self.element_find(
+            locator, first_only=False, required=False)
+        )
```
pekkaklarck, line 153: This is really hard to read. I'd use

```python
actual_locator_count = len(self.element_find(locator, first_only=False,
                                             required=False))
```

or

```python
elements = self.element_find(locator, first_only=False, required=False)
actual_locator_count = len(elements)
```

instead. Shortening the local `actual_locator_count` variable to just `count` would probably be a good idea too. Everything might be nicely < 80 chars after that.

src/Selenium2Library/keywords/element.py
```diff
@@ -171,10 +186,11 @@ def page_should_not_contain_element(self, locator, message='', loglevel='INFO'):
         Key attributes for arbitrary elements are `id` and `name`. See
         `introduction` for details about locating elements.
         """
-        self._page_should_not_contain_element(locator, None, message, loglevel)
-
-    # Public, attributes
+        self.page_not_contains_element(
+            locator, message=message, loglevel=loglevel
+        )
```
pekkaklarck: This is the last time I mention this, but I don't think it is generally a good idea to move all arguments into separate row if having them in the same row as the method call makes the line `> 80`. It works nicely in some cases, especially if arguments themselves are long, but in general I consider it a lot better to split just the offending arguments to own row:

```python
self.page_not_contains_element(locator, message=message, 
                               loglevel=loglevel)
```

src/Selenium2Library/keywords/formelement.py
```diff
@@ -245,16 +269,22 @@ def textfield_value_should_be(self, locator, expected, message=''):
...
         """
-        element = self._element_find(locator, True, False, 'text field')
-        if element is None: element = self._element_find(locator, True, False, 'file upload')
+        element = self.element_find(
+            locator, required=False, tag='text field'
+        )
+        if not element:
+            element = self.element_find(
+                locator, required=False, tag='file upload'
+            )
```
pekkaklarck: This is a good example of the thing I promised not to mention again. =)

pekkaklarck (changes requested):
Looks very good in general!

The only serious issue I noticed is how `Base` has instance attributes that I doubt should be per instance. See the comment in the class itself for details.

As a style issue I *really* don't like the idea to split function calls that cause > 78 char lines like

```python
result = some_function_call(
    "All args here", "but some would fit prev line"
)
```

There are places where this enhances readability, but in the above case we just get one extra line and confusing single `)`. Just splitting the offending arguments into a separate line is, in my opinion, a lot better general guideline:

```python
result = some_function_call("Those args that fit are here",
                            "remaining on second line")
```

Splitting all arguments to own line gets especially confusing in nested function calls and with if/for constructs.

src/Selenium2Library/__init__.py
aaltat: Done and checked imports in all changed files.

src/Selenium2Library/__init__.py
aaltat: Fixed

src/Selenium2Library/base.py
aaltat: No all instances must have the same value, that was the way it worked before the refactoring. Those timers where in https://github.com/robotframework/Selenium2Library/blob/master/src/Selenium2Library/keywords/browsermanagement.py#L45 but I moved them to Base. Other place that I was thinking to put them was the library `__init__.py` but can not anymore recall why I did actually put the timers here. Changed to not the instance specific but perhaps `__init__` would be better place after all.

src/Selenium2Library/keywords/alert.py
aaltat: Lets decide this one when we decide should we rename the library.

aaltat:
Í agree on the syntax comments but did not have time to fix those today. Hopefully tomorrow or next day. 

I am not totally satisfied on the solution for the timers, in `Base`, perhaps they should be moved to library `__init__`?

pekkaklarck:
Having config values in the library itself, i.e. in the context object, sounds good to me. If there are lot of them, we could create a separate class that has them all that the library then uses internally.

aaltat:
There are not many parameters which would be needed globally. There are those three timers and is the run on failure already running, so 4 in total.

pekkaklarck:
Do we still need a flag to tell is run-on-failure running?

aaltat:
I think we need. Because run on failure can run any keyword and therefore it does this `BuiltIn().run_keyword(self.ctx._run_on_failure_keyword)`. If the run on failure keyword is Selenium2Library​keyword and it fails too, it causes recursive loop.

There might be other solutions to solve the recursive loop problem, but this is the most backwards compatible way.

aaltat:
Fixed those too verbose coding styles, which I did spot. Also did little bit drive by refactoring and adjusted some spacing issues in the code which I have not touched.

pekkaklarck:
Forgot that run-on-failure is implemented using `BuiltIn.run_keyword`. You are right that it often ends up calling S2L keywords recursively.

aaltat:
It was pretty late when I made the last commit and therefore I might have missed some styling problems. I need to reread the changes one more time so that all is good to go. I will try to do it today or tomorrow. Is there something else that we would need to consider?

pekkaklarck:
Haven't looked at this PR since my full review, but I think this is good to be merged in general. We can fine tune the code also after that.

Just need to decide is there so big need for new minor release with better Firefox support that this should wait. Creating a branch for the minor release afterwards is possible too, though.

pekkaklarck (approved):

aaltat:
That is very good question. I am starting to think that doing Python 3 is more important because one can use the `Create Webdriver` keyword to get Firefox working. And because workaround exist, putting effort to release the Python 3 support feels more important.

pekkaklarck:
I trust your judgement on this.

aaltat:
There is always possibly to make the branch and release from there. Merged to master.

vairamsvsjdo:
@aaltat Thanks for this drive to version 3 support.

Issue #774, opened 2017-03-14
Expose selenium WebElement.value_of_css_property() to Selenium2Library


AlexanderWhitestone:
This method isn't used in any keywords in Selenium2Library right now.

There are good keywords for checking CSS values written here:
https://github.com/axon-id/Selenium2LibraryExtension/blob/master/src/Selenium2LibraryExtension/keywords/__init__.py

aaltat:
Closing due inactivity. If there is an interest to me this happen please open new issue and preferably create also a PR

PR #775, opened 2017-03-14
Css keywords

AlexanderWhitestone:
This keyword was written in a fork (I didn't write it)

Issue #774

I wasn't sure if I should keep the comment about the author or not.
I kept it because it was in the original repo.

### DIFF
--- a/src/SeleniumLibrary/keywords/element.py
+++ b/src/SeleniumLibrary/keywords/element.py
@@ -368,6 +368,24 @@ class ElementKeywords(LibraryComponent):
             raise AssertionError(message)
 
     @keyword
+    def element_css_attribute_should_be(self, locator, prop, expected):
+	"""Verifies the element identified by `locator` has the expected
+	    value for the targeted `prop`.
+	    | *Argument* | *Description* | *Example* |
+	    | locator | Selenium 2 element locator | id=my_id |
+	    | prop | targeted css attribute | background-color |
+	    | expected | expected value | rgba(0, 128, 0, 1) |"""
+
+	self.info("Verifying element '%s' has css attribute '%s' with a value of '%s'" % (locator, prop, expected))
+        element = self.find_element(locator)
+
+	value = element.value_of_css_property(prop)
+	if (value != expected):
+	    raise AssertionError("Element locator '%s' css property '%s' had a \
+            value of '%s' while it should have been '%s'" % (locator, prop,
+                                                             value, expected))
+
+    @keyword
     def get_element_attribute(self, locator, attribute):
         """Returns value of ``attribute`` from element ``locator``.
 


src/Selenium2Library/keywords/element.py
```diff
@@ -55,6 +55,26 @@ def current_frame_should_not_contain(self, text, loglevel='INFO'):
                                  "but it did" % text)
         self._info("Current page should not contain text '%s'." % text)
 
+
+    ######################################################################
+    ## Element Css Attribute Should Be					##
+    ##									##
+    ## Author:		Olivier Verville				##
+    ## Date:		May 24th 2016					##
+    ######################################################################
```
aaltat: Remove lines 58 - 64.

src/Selenium2Library/keywords/element.py
```diff
@@ -55,6 +55,26 @@ def current_frame_should_not_contain(self, text, loglevel='INFO'):
...
+    def element_css_attribute_should_be(self, locator, prop, expected):
+	"""Verifies the element identified by `locator` has the expected
+	    value for the targeted `prop`.
```
aaltat: Doc strings must follow pep256

src/Selenium2Library/keywords/element.py
```diff
@@ -728,6 +748,16 @@ def _get_text(self, locator):
             return element.text
         return None
 
+    def _check_element_css_value(self, locator, prop, expected):
```
aaltat: Because this method is only used in the `Element Css Attribute Should Be` it might be better idea to move the code from here to the keyword.

src/Selenium2Library/keywords/element.py
```diff
@@ -728,6 +748,16 @@ def _get_text(self, locator):
...
+	
+	element = self._element_find(locator, True, True)
+		
+	value = element.value_of_css_property(prop)
```
aaltat: Following pep8 is required, specially on the line length.

src/Selenium2Library/keywords/element.py
```diff
@@ -728,6 +748,16 @@ def _get_text(self, locator):
...
+	if (value != expected):
```
aaltat: Parentheses are not needed.

src/Selenium2Library/keywords/element.py
```diff
@@ -728,6 +748,16 @@ def _get_text(self, locator):
...
+	    raise AssertionError("Element locator '%s' css property '%s' had a "\
```
aaltat: ValueError could be better error class.

aaltat (changes requested):
Fix the comments and write tests. At least write write following tests:
1) Write few test that verify happy path
2) Test where property does not contain expected value
3) Test where element does not contain the CSS property
4) Test where css property is a number. Considering also RF providing the number in different types: 5 versus ${5}
5) Test where CSS property contains multiple elements, like: https://www.w3schools.com/cssref/pr_border.asp

AlexanderWhitestone:
I agree with all the requested changes.  Thanks @aaltat for the review and the feedback.
I will make the recommended changes hopefully soon.

aaltat:
For the new keyword, the code change looks OK. But it is missing the tests, which is not OK.

Also it looks like you used a tool to fix pep8 related errors, which is good thing too. But one pull request should do only a one thing. Having multiple fixes/enhancement in a one pull request makes the review harder. You can leave it for now but could you remember in the future.

AlexanderWhitestone:
I agree, I'll keep that in mind.
I still plan on writing the tests but it will take a bit longer as I needed to get acquainted with the test guidelines and be able to run them.

aaltat:
It's​explained in here: https://github.com/robotframework/Selenium2Library/blob/master/CONTRIBUTING.md#code-contributions and in here: https://github.com/robotframework/Selenium2Library/blob/master/BUILD.rst#unit-and-acceptance-tests

If you are developing in Windows, it might be good idea rebase with master. There is commit in master which fixes the test server problems when running in Windows.

aaltat:
How is the test case writing going on?

AlexanderWhitestone:
I'm unstuck, but I haven't had time to work on them.  Work, school, competitions, and consulting have had me fully booked.

aaltat:
How is the test case writing going on?

AlexanderWhitestone:
I haven't had time to work on them directly, but I have written some tests for other new keywords not yet pulled.  So the learning part is done.  I'm still planning on adding the tests, I just don't know when it will happen.

AlexanderWhitestone:
I discovered some limitations.
If you set a color to a named value, Selenium won't necessarily read it as-is from the style sheet.  It may return the RGB value instead of the named value like 'darkblue.'
So I guess this is not going to work for colors for now.

pekkaklarck:
It could also be the browser transforming color names to RGB values. In that case the actual behavior could also depend on the browser.

aaltat:
Verifying CSS can be complicated and one can run into unexpected problems. But I think it's still a good keyword to have.

The next release is coming soonnish and it would be nice to get this in.

AlexanderWhitestone:
I'm about half done with the tests.  If the color values are browser specific, then I think that will have to be handled by the user of the keyword.  For now I won't test for colors values because it might require browser specific conditions and would be difficult to maintain.

aaltat:
Sounds good to me.

aaltat:
Have you found time to finish this PR?

AlexanderWhitestone:
Sorry for the delay. I just started a new job so I haven't been working on this.

aaltat:
No worries. Do you think that at some point you would have time to finish this PR?

AlexanderWhitestone:
Yes, I will have time at some point.

AlexanderWhitestone:
Do we still want this? I let this PR get really old and I want to know if i should try fixing it.

aaltat:
Welcome back after so long time.

The feature gets asked now and then in the community and therefore the keywords would be good addition to the library. And if I recall correctly, it was just missing few acceptance tests. 

If you have time and will to fix the PR, I am happy to review and merge it. The history should be cleaned tough. We usually do not use merge commit's and history should be clean. Which means that your commits should be on top of the commit's that are already in the master.

AlexanderWhitestone:
Yea I messed up the history while trying to pull in the latest.  I will work on it and do a proper rebase so the commits are on top.

aaltat:
No problems, take your time.

aaltat:
Are you still planning to work on the PR or can we close it?

AlexanderWhitestone:
We can close it.  I haven't been using the library and I'm not long familiar with it unfortunately.

I fixed the history and pushed what I had incase it helps anyone coming in to work on this in the future.

aaltat:
I understand and now worries. Thanks for effort and if there is a need, it can be always found from the GitHub.

PR #788, opened 2017-04-03
Support switching windows and getting some window information on browsers without javascript

davidshepherd7:
I'm using this library with [QtWebDriver](https://github.com/cisco-open-source/qtwebdriver) to test Qt applications. Some of the windows in my application are not actually web-based, so they don't support the `execute_script` command. Currently this makes it impossible to switch between windows because a script is used to get the window title and url.

By using the web driver API instead of a custom `execute_script` we can implement the same functionality without javascript.

### DIFF
--- a/src/Selenium2Library/locators/windowmanager.py
+++ b/src/Selenium2Library/locators/windowmanager.py
@@ -140,7 +140,14 @@ class WindowManager(object):
         raise ValueError(error)
 
     def _get_current_window_info(self, browser):
-        id_, name = browser.execute_script("return [ window.id, window.name ];")
+        try:
+            id_, name = browser.execute_script("return [ window.id, window.name ];")
+        except WebDriverException:
+            # The webdriver implementation doesn't support Javascript so we
+            # can't get window id or name this way.
+            id_ = None
+            name = ''
+
         title = browser.title
         url = browser.current_url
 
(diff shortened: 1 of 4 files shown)


src/Selenium2Library/locators/windowmanager.py
```diff
@@ -139,8 +140,17 @@ def _select_matching(self, browser, matcher, error):
         raise ValueError(error)
 
     def _get_current_window_info(self, browser):
-        id_, name, title, url = browser.execute_script(
-            "return [ window.id, window.name, document.title, document.URL ];")
+        try:
+            id_, name = browser.execute_script("return [ window.id, window.name ];")
+        except WebDriverException:
```
davidshepherd7, line 145: Unfortunately this is the exception type raised, it would be nice if it was something more specific.

davidshepherd7:
It doesn't look like those failing tests have anything to do with my changes.

aaltat:
Write an issue where you explain the problem you are having, It also helps maintainers to keep track of the changes.

By a quick look from a mobile screen, PR looks OK. But I have to give it a closer look from the mobile screen.

The test did stuck for some unknown reason and it happens sometimes in the Travis. I did rerun test and let's see the results.

davidshepherd7:
Hi, I created an issue for it.

It looks like the remaining test failure is a false positive (all the recent builds seem to have the same failure for the same reason).

aaltat:
Thank you. Yes the Python 3 support is not ready and therefore one of the CI jobs fail.

src/Selenium2Library/locators/windowmanager.py
aaltat: It's alright. At least it indicates that some sort of selenium error did happen and it is, in this case, OK to suppress the error.

aaltat (approved):

Issue #701, opened 2016-12-07
Combining `Open Browser`and `Create Webdriver`keywords 


aaltat:
* Take KW args in use
    * `Open Browser` is made in are where kwargs where not available from Robot Framework. The kwargs could allow creating the browser in more generic way
    * could pass the `desired_capabilities` in dictionary. Also should support other configuration options from 
 * One keyword only to handle the opening the browser
It should be possible to move many things from Create Webdriver` keyword to `Open Browser` keyword. Then we should create pubic API to register the browser/driver to S2L side, so that user could create the browser/driver in the python code as they want. 
 * See what selenium API supports
The might be some difficult scenarios with selenium API, which can not be handled with vargars and kwargs. But documenting and creating public API for registering these should be enough. 
 * Think should Open Browser keyword open always keyword or should it reuse existing browser
There is use case where the `Open Browser` keyword could do the same thing as `Go To` keyword if the browser is already open. Think how to do it. 

To better support what `Create WebDriver` keywords enables to do, it might the argument signature of the `Open Keyword` should be changed, or perhaps `Open Browser 2` keywords should be created.
Now the argument signature is:
```
| Open Browser | url | browser=firefox | alias=None | remote_url=False | desired_capabilities=None | |ff_profile_dir=None |
```
But it might better to have: 
```
| Open Browser | url | browser=chrome | alias=None | remote_url=False | desired_capabilities=None | options=None |  service_log_path=None | ff_profile_dir=None |
```

accesslubos:
Hello! Question.
I was just experimenting with headless Chrome. Currently I add that as an argument and then I call the Create Webdriver keyword directly:
```
Call Method    ${chromeOptions}    add_argument    --headless
Create Webdriver    ${global_browserName}    chrome_options=${chromeOptions}
```
Now I noticed there is an easy option to pass in "headlesschrome" as a browser name to the Open Browser keyword which is not available with the Create Webdriver keyword. Would it be possible to combine those two/is this in scope of this ticket?
The only reason why I want to use Create Webdriver instead of Open Browser is specifying download directory and client language setup.

Let me know if I can help!

aaltat:
[...]

PR #797, opened 2017-04-20
Added kwargs for Open Browser (#701)

bdtexas:
Implements the first bullet in #701


aaltat:
Thank you from the pull request, but this is not what I did have in my mind. If you interested for a second try, we could discuss this in issue #701 and try to agree on the design.

PR #799, opened 2017-04-27
Fix typo in run_tests.py

AlexanderWhitestone:
Spell 'supported' correctly on line 229


aaltat:
Thank from spotting the problem and providing PR to fix it. All looks good to go.

PR #803, opened 2017-05-07
Added context object also to TableElementFinder

aaltat:


pekkaklarck:
New `ContextAware` and `LibraryComponent` seem to work well. They could be in a common module, though, and that module could probably be named `base`.

src/Selenium2Library/locators/__init__.py
```diff
@@ -1,12 +0,0 @@
-from .elementfinder import ElementFinder
-from .customlocator import CustomLocator
-from .tableelementfinder import TableElementFinder
-from .windowmanager import WindowManager
```
pekkaklarck: In my opinion exposing different finders and locators directly on the `locators` package level was a good idea. In general I prefer packages exposing their public API in the `__init__` file. After this change imports like

```python
from Selenium2Library.locators import WindowManager
```

need to be changed to

```python
from Selenium2Library.locators.windowmanager import WindowManager
```

and that is definitely not an enhancement.

src/Selenium2Library/locators/__init__.py
```diff
@@ -1,12 +0,0 @@
...
-
-
-__all__ = [
-    "ElementFinder",
-    "TableElementFinder",
-    "WindowManager",
-    "CustomLocator"
-]
```
pekkaklarck, line 12: `__all__` isn't needed and can go.

src/Selenium2Library/librarycomponent.py
```diff
@@ -2,15 +2,16 @@
 from selenium.webdriver.remote.webelement import WebElement
 
 from .locators.elementfinder import ElementFinder
+from Selenium2Library.contextaware import ContextAware
```
pekkaklarck: Somewhat inconsistent to use `.module` with one import and `Selenium2Library.module` with the other.

pekkaklarck:
On a pretty quick look I mostly like these changes. Summary of my thoughts:

- Passing context to `TableElementFinder` works great.
- Splitting `Base` to `ContextAware` and `LibraryComponent` base classes works well too. As I commented above, I'd prefer having these two in a common module (e.g. `base`).
- As I already commented, I don't like not exposing finders/locators directly on the `locator` package level.
- Quite many unrelated changes in this PR.

Issue #809, opened 2017-05-22
Python 3 support: NameError:  'unicode' is not defined when 'Wait Until Page Contains' text


Sushma-Kattinti-Bose:
Python Version:  3.6.1
S2L Version:  3dca1dd

I tried to get the latest master to work with python 3.6.1, but when running the following test I get a stack trace about unicode support (something that changed between python 2 and 3):

```
*** Settings ***
Documentation  This is basic info about the whole suite
Library  Selenium2Library

*** Variables ***

*** Test Cases ***
User must sign in to check out
    [Documentation]  This is some basic info about the test
    [Tags]  Smoke
    Open Browser  http://www.amazon.com  chrome
    Wait Until Page Contains  Your Amazon.com
    Input Text  id=twotabsearchtextbox  Ferrari 458
    Click Button  xpath=//*[@id='nav-search']/form/div[2]/div/input
    Wait Until Page Contains  results for "Ferrari 458"
    Click Link  css=#result_0 a.s-access-detail-page
    Wait Until Page Contains  Back to search results
    Click Button  xpath=//*[@id='add-to-cart-button']
    Wait Until Page Contains  Added to Cart
    Click Link  Proceed to checkout
    Page Should Contain Element  xpath=//*[@id='hlb-ptc-btn-native']
    Sleep  3s
    Close Browser

*** Keywords ***
```

KEYWORD Selenium2Library . Wait Until Page Contains Your Amazon.com
Documentation:	
Waits until `text` appears on current page.
Start / End / Elapsed:	20170522 19:18:18.463 / 20170522 19:18:24.097 / 00:00:05.634
19:18:24.097	FAIL	NameError: name 'unicode' is not defined

aaltat:
This is actually the last thing to be fixed for the Python 3 support. The good news is that that there is a fix in my Selenium2Library fork: https://github.com/aaltat/robotframework-selenium2library/tree/python_3_support?files=1

The unit and acceptance test did pass in my local environment and the code contains bunch of TODO comments which needs to be removed but it should work. If you want, you could give it a try and I would gladly take feedback.

aaltat:
Made a PR: #810 so that it easier to find and contains some code cleanup too.

pekkaklarck:
Isn't there already an open issue about Py3? This should be closed as its dupe. And soon the original can be closed too!

aaltat:
[...]

PR #810, opened 2017-05-23
Python 3 support

aaltat:
Fixes the #809


test/run_tests.py
```diff
@@ -121,7 +121,7 @@ def http_server():
 
 def execute_tests(interpreter, browser, rf_options, sauce_username, sauce_key):
     options = []
-    runner = interpreter.split() + ['-m', 'robot.run']
+    runner = interpreter.split() + ['-m', 'robot']
```
pekkaklarck: This change breaks test execution with RF < 3.0. Need to change back to `robot.run` even though that causes an annoying warning with Python 3.6.

pekkaklarck (changes requested):
Looks good except for the change to acceptance test execution. It would be a bit better to have unrelated cleanup in a separate commit, though.

PR #811, opened 2017-05-26
Refector ElementFinder to better support ctx object.

aaltat:
Also also simplifies where the prefix is determined and which
selenium method is used to find elements

In past //div was defaulted to xpath prefix, because it starts "//".
Now also (// defaults to xpath prefix, because example (//div)[1] is
valid xpath.

### DIFF
--- a/src/Selenium2Library/locators/elementfinder.py
+++ b/src/Selenium2Library/locators/elementfinder.py
@@ -4,10 +4,10 @@ from robot.utils import NormalizedDict
 from Selenium2Library.utils import escape_xpath_value, events
 
 
-class ParserLocator(object):
+class LocatorParser(object):
 
     @classmethod
-    def parse_locator(cls, locator):
+    def parse(cls, locator):
         prefix = None
         criteria = locator
         if locator.startswith('//') or locator.startswith('(//'):
@@ -63,7 +63,7 @@ class ElementFinder(object):
         assert browser is not None
         assert locator is not None and len(locator) > 0
 
-        prefix, criteria = ParserLocator.parse_locator(locator)
+        prefix, criteria = LocatorParser.parse(locator)
         strategy = self._strategies.get(prefix)
         if strategy is None:
             raise ValueError('Element locator with prefix \'{}\' '
--- a/src/Selenium2Library/locators/elementfinder.py
+++ b/src/Selenium2Library/locators/elementfinder.py
@@ -10,7 +10,7 @@ class LocatorParser(object):
     def parse(cls, locator):
         prefix = None
         criteria = locator
-        if locator.startswith('//') or locator.startswith('(//'):
+        if locator.startswith(('//', '(//')):
             prefix = 'xpath'
         else:
             locator_parts = locator.split('=', 1)
--- a/src/Selenium2Library/locators/elementfinder.py
+++ b/src/Selenium2Library/locators/elementfinder.py
@@ -8,19 +8,15 @@ class LocatorParser(object):
 
     @classmethod
     def parse(cls, locator):
-        prefix = None
-        criteria = locator
         if locator.startswith(('//', '(//')):
             prefix = 'xpath'
+            criteria = locator
+        elif '=' not in locator:
+            prefix = 'default'
+            criteria = locator
         else:
-            locator_parts = locator.split('=', 1)
-            if len(locator_parts) == 2:
-                prefix = locator_parts[0]
-                criteria = locator_parts[1].strip()
-            elif len(locator_parts) == 1:
-                prefix = 'default'
-                criteria = locator
-        return prefix, criteria
+            prefix, criteria = locator.split('=', 1)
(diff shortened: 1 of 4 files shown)


src/Selenium2Library/locators/elementfinder.py
```diff
@@ -4,6 +4,25 @@
 from Selenium2Library.utils import escape_xpath_value, events
 
 
+class ParserLocator(object):
```
pekkaklarck: `LocatorParser` sounds a much better than `ParserLocator` IMHO. Also, I'd prefer helper class like this to be below the main class of this module.

src/Selenium2Library/locators/elementfinder.py
```diff
@@ -4,6 +4,25 @@
...
+
+    @classmethod
+    def parse_locator(cls, locator):
```
pekkaklarck: - What benefits does making this a class method bring?
- Since this is a method in a locator parser class, it could be named simply `parse`. Namespaces are a honking good idea and all that.

src/Selenium2Library/locators/elementfinder.py
```diff
@@ -4,6 +4,25 @@
...
+        prefix = None
+        criteria = locator
+        if locator.startswith('//') or locator.startswith('(//'):
```
pekkaklarck: `startswith()` accept a tuple of possible matches like `locator.startswith(('//', '(//'))`.

src/Selenium2Library/locators/elementfinder.py
```diff
@@ -23,19 +42,33 @@ def __init__(self):
...
-        self._strategies = NormalizedDict(initial=strategies, caseless=True, spaceless=True)
+        self._strategies = NormalizedDict(initial=strategies, caseless=True,
+                                          spaceless=True)
         self._default_strategies = list(strategies.keys())
+        self._key_attrs = {
+            None: ['@id', '@name'],
+            'a': [
+                '@id', '@name', '@href',
+                'normalize-space(descendant-or-self::text())'
+            ],
```
pekkaklarck: A bit verbose formatting, especially compared to how the earlier entry is on one line only. #nitpick

src/Selenium2Library/locators/elementfinder.py
```diff
@@ -23,19 +42,33 @@ def __init__(self):
...
 
-        (prefix, criteria) = self._parse_locator(locator)
-        prefix = 'default' if prefix is None else prefix
+        prefix, criteria = ParserLocator.parse_locator(locator)
         strategy = self._strategies.get(prefix)
         if strategy is None:
-            raise ValueError("Element locator with prefix '" + prefix + "' is not supported")
-        (tag, constraints) = self._get_tag_and_constraints(tag)
+            raise ValueError('Element locator with prefix \'{}\' '
+                             'is not supported'.format(prefix))
```
pekkaklarck: - Why not double quotes around the string to avoid escaping the internal single quotes?

- As a very minor style thing, instead of using `self._strategies.get()` and validating that result isn't `None`, I'd use

   ```
    try:
        strategy = self._strategies[prefix]
    except KeyError:
        raise ValueError(...)
   ```

   or

   ```
    if prefix not in self._strategies:
        raise ValueError(...)
    strategy = self._strategies[prefix]
    ```

src/Selenium2Library/locators/elementfinder.py
```diff
@@ -4,6 +4,25 @@
...
+        if locator.startswith('//') or locator.startswith('(//'):
+            prefix = 'xpath'
+        else:
+            locator_parts = locator.split('=', 1)
+            if len(locator_parts) == 2:
+                prefix = locator_parts[0]
+                criteria = locator_parts[1].strip()
+            elif len(locator_parts) == 1:
+                prefix = 'default'
+                criteria = locator
```
pekkaklarck: Something like this would probably be a bit easier to understand:

```python
if locator.startswith(('//', '(//')):
    return 'xpath', locator
if '=' not in locator:
    return 'default', locator
prefix, locator = locator.split('=', 1)
return prefix, locator.strip()
```

pekkaklarck (approved):
Looks good to me in general. Please look at my line notes, though. Also, was there a possibility that some other piece of the code could use the same locator parsing logic?

src/Selenium2Library/locators/elementfinder.py
aaltat: Done

src/Selenium2Library/locators/elementfinder.py
aaltat: 1) The ` @classmethod` allows to use the parser method without creating instance, like `LocatorParser.parse(locator)`
2) Agreed and done

src/Selenium2Library/locators/elementfinder.py
aaltat: I did not know that, good idea. Done

PR #817, opened 2017-06-02
Accept all text-like input fields in Textfield-related keywords

### LINKED ISSUE #546: HTML5 specialized text fields not recognized as text fields
HTML5 introduced several new input types, several of which are specialized text inputs including email,  number, url, search, and tel. These input types are not recognized by `Page Should Contain Textfield`, `Page Should Not Contain Textfield`, `Textfield Should Contain`,  or `Textfield Value Should Be`.
The cause is that these keywords use the constraint 'text field' and ElementFinder will filter out any input elements that do not have type 'text'. Since these inputs are still text inputs, they should be treated as such. The number type may render a spinbox control in some browsers, but still behaves like a text-type input.

Here is a test that should pass, but fails with:
Value of text field 'name=email' should have been 'selenium@xxx.yyy' but was 'None'

```
Open Browser    http://www.the-art-of-web.com/html/html5-form-validation/   gc
Input Text  name=email  selenium@xxx.yyy
Textfield Value Should Be   name=email  selenium@xxx.yyy
```

With minor modifications to elementfinder, I was able to get the above to pass.

This is being entered as a result of this: http://stackoverflow.com/questions/34169390/how-to-find-value-of-email-field-using-selenium2library

leods92:
This fixes #546

### DIFF
--- a/src/Selenium2Library/locators/elementfinder.py
+++ b/src/Selenium2Library/locators/elementfinder.py
@@ -253,9 +253,10 @@ class ElementFinder(ContextAware):
         if not element.tag_name.lower() == tag:
             return False
         for name in constraints:
-            if isinstance(constraints[name], list) and element.get_attribute(name) not in constraints[name]:
-                return False
-            if not element.get_attribute(name) == constraints[name]:
+            if isinstance(constraints[name], list):
+                if element.get_attribute(name) not in constraints[name]:
+                    return False
+            elif not element.get_attribute(name) == constraints[name]:
                 return False
         return True
 
--- a/src/Selenium2Library/locators/elementfinder.py
+++ b/src/Selenium2Library/locators/elementfinder.py
@@ -242,7 +242,7 @@ class ElementFinder(ContextAware):
         if not element.tag_name.lower() == tag:
             return False
         for name in constraints:
-            if type(constraints[name]) is list and element.get_attribute(name) not in constraints[name]:
+            if isinstance(constraints[name], list) and element.get_attribute(name) not in constraints[name]:
                 return False
             if not element.get_attribute(name) == constraints[name]:
                 return False
--- a/src/Selenium2Library/locators/elementfinder.py
+++ b/src/Selenium2Library/locators/elementfinder.py
@@ -187,7 +187,7 @@ class ElementFinder(ContextAware):
             key_attrs = self._key_attrs[None]
         xpath_criteria = escape_xpath_value(criteria)
         xpath_tag = tag if tag is not None else '*'
-        xpath_constraints = ["@%s='%s'" % (name, constraints[name]) for name in constraints]
+        xpath_constraints = self._get_xpath_constraints(constraints)
         xpath_searchers = ["%s=%s" % (attr, xpath_criteria) for attr in key_attrs]
         xpath_searchers.extend(self._get_attrs_with_url(key_attrs, criteria))
         xpath = "//%s[%s%s(%s)]" % (
@@ -198,6 +198,17 @@ class ElementFinder(ContextAware):
         )
         return self._normalize_result(
             self.browser.find_elements_by_xpath(xpath))
+
+    def _get_xpath_constraints(self, constraints):
+        xpath_constraints = []
+        for name in constraints:
+            value = constraints[name]
+            if isinstance(value, list):
+                contraint = "@%s[. = '%s']" % (name, "' or . = '".join(value))
+            else:
+                contraint = "@%s='%s'" % (name, value)
+            xpath_constraints.append(contraint)
+        return xpath_constraints
 
     def _get_tag_and_constraints(self, tag):
         if tag is None:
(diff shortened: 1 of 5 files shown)


src/Selenium2Library/locators/elementfinder.py
```diff
@@ -192,6 +194,8 @@ def _element_matches(self, element, tag, constraints):
         if not element.tag_name.lower() == tag:
             return False
         for name in constraints:
+            if type(constraints[name]) is list and element.get_attribute(name) not in constraints[name]:
+                return False
```
pekkaklarck: - Why is this change needed?
- Should use `usinstance`, not `tyoe`

pekkaklarck:
I'm +1 for this enhancement but tests are needed and this should be mentioned in docs too.

aaltat:
And raise issue also from the problem, so that release notes are properly created.

pekkaklarck:
@aaltat There's already #546 about this.

aaltat:
Ah, I did not remember that one, thanks Pekka.  

In any case, the changes needs to be tested. In in unit and/or acceptance level. Also the change breaks the existing unit test which is not acceptable. 

For the code side, I do not have other comments than Pekka did made. Could you fix the testing side related problems and lets see how it continues from there.

src/Selenium2Library/locators/elementfinder.py
leods92: > Why is this change needed?

Currently constraints only accept 1 value per element attribute and the change involves the creation of a list of values that should be considered acceptable.

> Should use usinstance, not tyoe

Will do the change. Never coded in Python so I really don't know what's best. :) Thanks for the review! 👍

leods92:
@aaltat 

> In any case, the changes needs to be tested. In in unit and/or acceptance level. Also the change breaks the existing unit test which is not acceptable.

> Could you fix the testing side related problems and lets see how it continues from there.

For sure! When I opened the pull request I was busy at work. I was hoping somebody experienced would take it over. In any case, I have some time know and I'm gonna take the time to learn a bit about testing in Python. If I cannot successfully finish it, I'll let you guys know.

leods92:
@pekkaklarck 

Could you please have a second look?
I tried to make tests diversified by testing 2 additional input types: `email` and `url`.
I believe they are the most used text-like input types after `text`.

I hope that's enough, if it's not, please let me know how I can improve. :)

Thanks!

pekkaklarck:
I'm on vacation without my laptop next week (and currently trying to get everything packed). I can take a quick look at the code now, but hopefully @aaltat has time for a full review.

src/Selenium2Library/locators/elementfinder.py
```diff
@@ -151,6 +151,17 @@ def _find_by_default(self, browser, criteria, tag, constraints):
...
+    def _get_xpath_constraints(self, constraints):
+        xpath_constraints = []
+        for name in constraints:
+            value = constraints[name]
+            if isinstance(value, list):
+                contraint = "@%s[. = '%s']" % (name, "' or . = '".join(value))
+            else:
+                contraint = "@%s='%s'" % (name, value)
+            xpath_constraints.append(contraint)
+        return xpath_constraints
```
pekkaklarck: This seems to be totally fine, but if there would be `_get_xpath_constraint`, looping could be handled by a list comprehension similarly as earlier:

```python
xpath_constraints = [self._get_xpath_constraint(name, value)
                     for name, value in constraints.items()]
```

src/Selenium2Library/locators/elementfinder.py
```diff
@@ -192,7 +205,10 @@ def _element_matches(self, element, tag, constraints):
         if not element.tag_name.lower() == tag:
             return False
         for name in constraints:
-            if not element.get_attribute(name) == constraints[name]:
+            if isinstance(constraints[name], list):
+                if element.get_attribute(name) not in constraints[name]:
+                    return False
+            elif not element.get_attribute(name) == constraints[name]:
```
pekkaklarck: Using `!=` would be more clear than `not` and `==`.

leods92: I agree but I thought you Python guys did that (based on the line removed above). Change made! :)

src/Selenium2Library/locators/elementfinder.py
leods92: Made this change too. It's much better. I was just afraid of creating a new method because all other methods in the file are relatively long. I really like having smaller and more methods. ^^

aaltat:
I can take a look at this week.

aaltat:
And I was sick last week, I will try to find time this week.

aaltat:
By looking from mobile screen, this looks OK. I need to give second look from a bigger screen.

In mean while, could you rebase with master and create a PR which does not contain conflicts?

aaltat:
All looks OK to me. Would you mind sorting out the conflict and it can be merged to the master?

leods92:
@aaltat On it! Sorry it took me so long to reply. Weekend's here now! 💃

leods92:
@aaltat Rebased and Travis is green! ✅ 😄

aaltat:
No problem. I will try to take a look later and if all is well then merge to master.

PR #818, opened 2017-06-03
Updated documentation for locators

aaltat:
Also mentions that xpath is supported without using the xpath prefix.

### DIFF
--- a/src/Selenium2Library/__init__.py
+++ b/src/Selenium2Library/__init__.py
@@ -51,16 +51,23 @@ class Selenium2Library(DynamicCore):
 
     *Using locators*
     ---------------
-    By default, when a locator value is provided, it is matched against the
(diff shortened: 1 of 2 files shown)


aaltat:
Also added acceptance tests.

pekkaklarck:
Doc changes look great. Should mention that `(//` is supported only by next version, though, but the problem is specifying what version and library name to use. SL will be 3.0, but should S2L be 3.0 or 2.0? If both are 3.0, we could say "supported by library version 3.0 or newer" or something similar.

aaltat:
Perhaps we create SL and S2L as 3,0? It would solve the problem nicely.

pekkaklarck:
Yeah, v. 3.0 with both sounds good to me.

PR #819, opened 2017-06-03
Updating minimum required versions of Python and selenium.

aaltat:
Also contains documentation updates.


aaltat:
There is `easy_install` in the documentation, do we need to keep that one too in the docs?

pekkaklarck:
Definitely no need to mention easy_install. Don't think proxy stuff is that important here either. Could link to Robot's more comprehensive docs about that and other more advanced pip/installation topics. Or to pip's own docs.

aaltat:
OK, I will delete the `easy_install` chapter. Then there is only `pip` and source install options.

Linking to other documents is also good idea. It would be nice to know is the Jython supported in S2L, but not in this release.

Issue #822, opened 2017-06-13
Add Optional Unique Domain Tracking


snyderr:
It would be nice to have the option to get a list of all domains visited during testing. By having this feature in the Selenium2Library we could make keywords like `Delete All Cookies` more useful. Currently `Delete All Cookies` will only remove cookies from the current domain. It would be much more useful if we could delete all the cookies from any domain visited during testing. This would allow for easier browser cleanup in between tests and reduce the need to restart the browser during testing.

Here's how you would use domain tracking in your testing:
`Library  Selenium2Library  domain_tracking=${true}` (domain_tracking will default to False)

I currently have a working implementation of this feature. Some things that I am unsure of is if storing the domains in a set is too specific. Should we just store all URL's visited in a list during testing instead? I am currently parsing the URL's into domains and adding them to a set. This means that you would only get unique domains visited during testing. This seems more useful for deleting cookies, but I could see use cases for wanting all of the URLS visited. Perhaps we could add both or leave the unique domains up to the person using the library?

I can put a PR up if this feature seems like something that people would be interested in, I just have a little touching up to do.

Current Keyword Ideas:
`Get Domains Visited`: Returns a set of domains visited during testing.

`Clear Domains Visited`: Clears all domains visited during testing.

`Delete All Cookies Across All Domains`: Navigates to all domains visited during testing and deletes the cookies. (Perhaps we could expand the current `Delete All Cookies` keyword to do this.)

aaltat:
I haven't found a need for this type of feature, but it doesn't not mean that it could not be useful. At least I can recall someone else wanting this type of feature. Let me try to rephrase the problem you are trying to solve.

The `Delete All Cookies` deletes cookies only from the current domain where the browser currently is. If I recall this is a browser security feature and does not depend on the selenium, am I right? Now you want to track all the domains where browser has visited and delete cookies from sites where the browser has visited. Did I get it right?

Now do you want to track all the domains:
1) Where the browser navigates by `Open Browser` and `Go To` keywords? 
2) Do you want to keep track the domains where the browser is redirected?
3) Do you want to keep track of the domains which load third party content to the pages?

It seems that you have a PoC in your hands, would it be possible to see the code?

snyderr:
Yes, there is a browser security limitation, it is not something to do with selenium. Yes, I would like to delete cookies from sites where the browser has visited.

[...]

PR #823, opened 2017-06-15
Implementation of Unique Domain Tracking

snyderr:
Related to [this](https://github.com/robotframework/Selenium2Library/issues/822) issue.

### DIFF
--- a/src/Selenium2Library/__init__.py
+++ b/src/Selenium2Library/__init__.py
@@ -1,4 +1,6 @@
 import warnings
+import urllib
+import urlparse
 
 from .keywords import AlertKeywords
 from .keywords import BrowserManagementKeywords
@@ -15,6 +17,7 @@ from .robotlibcore import DynamicCore
 from .utils import BrowserCache
 from .utils import LibraryListener
 from .version import VERSION
+
 
 __version__ = VERSION
 
@@ -159,7 +162,8 @@ class Selenium2Library(DynamicCore):
                  timeout=5.0,
                  implicit_wait=0.0,
                  run_on_failure='Capture Page Screenshot',
-                 screenshot_root_directory=None):
+                 screenshot_root_directory=None, 
+                 domain_tracking=False):
 
         """Selenium2Library can be imported with optional arguments.
 
@@ -196,6 +200,8 @@ class Selenium2Library(DynamicCore):
         self._speed_in_secs = 0.0
         self._timeout_in_secs = 5.0
         self._implicit_wait_in_secs = 5.0
+        self._domains_visited = set()
+        self._domain_tracking = domain_tracking
         libraries = [
             AlertKeywords(self),
             BrowserManagementKeywords(self),
@@ -218,14 +224,25 @@ class Selenium2Library(DynamicCore):
         self.ROBOT_LIBRARY_LISTENER = LibraryListener()
 
     def run_keyword(self, name, args, kwargs):
+        if self._domain_tracking:
+            self._record_current_domain()
         try:
             return DynamicCore.run_keyword(self, name, args, kwargs)
         except Exception:
             RunOnFailureKeywords(self).run_on_failure()
             raise
 
+    def _record_current_domain(self):
+        try:
+            uri = urlparse.urlsplit(self.get_location())
+            if uri.netloc:
+                self._domains_visited.add('{uri.scheme}://{uri.netloc}/'.format(uri=uri))
+        except:
+            pass   
+
     def register_browser(self, browser, alias):
         return self._browsers.register(browser, alias)
+
 
     @property
     def _browser(self):
@@ -245,3 +262,7 @@ class Selenium2Library(DynamicCore):
                       'use "Selenium2Library.browser" instead.',
                       DeprecationWarning)
         return self.browser
+
+    def _get_domains_visited(self):
+        assert(self._domain_tracking, "Must enable domain tracking to get the domains visited.")
+        return self._domains_visited
--- a/src/Selenium2Library/keywords/browsermanagement.py
+++ b/src/Selenium2Library/keywords/browsermanagement.py
@@ -12,6 +12,9 @@ from Selenium2Library.base import LibraryComponent
 from Selenium2Library.locators.windowmanager import WindowManager
 from Selenium2Library.robotlibcore import keyword
 
+from robot.api import logger
+
+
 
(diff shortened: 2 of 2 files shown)


aaltat:
This is for #822

aaltat:
The new functionality need tests, it would be great to write unit test and some acceptance tests. By a quick look code looks OK, but there are some things which needs some thinking from me before I can give better input.

aaltat:
Also the code does not work with Python 3 which is not OK.

src/Selenium2Library/keywords/browsermanagement.py
```diff
@@ -12,6 +12,9 @@
 from Selenium2Library.locators.windowmanager import WindowManager
 from Selenium2Library.robotlibcore import keyword
 
+from robot.api import logger
```
aaltat, line 15: Robot logger is not used anymore directly, if you need logging they are found from self.

src/Selenium2Library/__init__.py
```diff
@@ -218,15 +224,26 @@ def __init__(self,
...
     def run_keyword(self, name, args, kwargs):
+        if self._domain_tracking:
+            self._record_current_domain()
         try:
             return DynamicCore.run_keyword(self, name, args, kwargs)
         except Exception:
             RunOnFailureKeywords(self).run_on_failure()
             raise
 
+    def _record_current_domain(self):
```
aaltat, line 235: Put the new methods in a separate class, where so that the testing is easier

src/Selenium2Library/__init__.py
```diff
@@ -196,6 +200,8 @@ def __init__(self,
         self._speed_in_secs = 0.0
         self._timeout_in_secs = 5.0
         self._implicit_wait_in_secs = 5.0
+        self._domains_visited = set()
```
aaltat, line 203: Would just list be better?

src/Selenium2Library/__init__.py
```diff
@@ -245,3 +262,7 @@ def _current_browser(self):
                       'use "Selenium2Library.browser" instead.',
                       DeprecationWarning)
         return self.browser
+
+    def _get_domains_visited(self):
+        assert(self._domain_tracking, "Must enable domain tracking to get the domains visited.")
```
aaltat, line 267: Using plain assert raises assertion error, which is not good idea. Either ValueError or create custom error class. Most likely later one would be better.

src/Selenium2Library/keywords/browsermanagement.py
```diff
@@ -62,6 +65,32 @@ def close_browser(self):
...
+            raise AssertionError("Cannot use this keyword when domain tracking is not enabled.")
+        self.ctx._domain_tracking = False
+        for domain in self.ctx._domains_visited:
+            self.go_to(domain)
+            self.browser.delete_all_cookies()
+        self.ctx._domains_visited = set()
+        self.ctx_domain_tracking = True
+
+    @keyword
+    def get_domains_visited(self):
```
aaltat, line 80: Could you describe the use case for this keyword?

src/Selenium2Library/keywords/browsermanagement.py
```diff
@@ -62,6 +65,32 @@ def close_browser(self):
             self.browsers.close()
 
     @keyword
+    def delete_all_cookies_across_all_domains(self):
```
aaltat, line 68: Cookie keywords belongs to [CookieKeywords](https://github.com/robotframework/Selenium2Library/blob/master/src/Selenium2Library/keywords/cookie.py). Would it be good idea to incorporate the code with existing `Delete All Cookies`keyword? There is some backwards incompatible issue worries but that should not be problem.

src/Selenium2Library/keywords/browsermanagement.py
```diff
@@ -62,6 +65,32 @@ def close_browser(self):
...
+    def get_domains_visited(self):
+        """ Returns a list of unique domains visited. """
+        if not self.ctx._domain_tracking:
+            raise AssertionError("Cannot use this keyword when domain tracking is not enabled.")
+        return self.ctx._domains_visited
+
+    @keyword
+    def clear_domains_visited(self):
+        """ Clears the list of unique domains visited. """
+        if not self.ctx._domain_tracking:
```
aaltat, line 89: I think keyword should pass, even the attribute is false.

src/Selenium2Library/keywords/browsermanagement.py
```diff
@@ -62,6 +65,32 @@ def close_browser(self):
...
+        for domain in self.ctx._domains_visited:
+            self.go_to(domain)
+            self.browser.delete_all_cookies()
+        self.ctx._domains_visited = set()
+        self.ctx_domain_tracking = True
+
+    @keyword
+    def get_domains_visited(self):
+        """ Returns a list of unique domains visited. """
+        if not self.ctx._domain_tracking:
```
aaltat, line 82: I think keyword should not fail if feature is not enabled. Retuning just False or None could be better.

Or do you have use case why it should fail?

aaltat:
This still could make to the next release, if you could make the improvements suggested by the review? Or would yo be willing to make the improvements in some later point of in time?

snyderr:
@aaltat I am currently too busy to make the suggested changes but will do so when I find the time.

After using my implementation in my SUT I am unsure if this is the best way to go about implementing this. Run time is greatly effected right now as the tracking is triggered on every single keyword call. Since this cannot be multi-threaded I think it would be wise to limit the keywords that can trigger the domain tracking. I'm not sure that this feature is worth implementing until we can get the impact to the run time decreased.

What are your thoughts on this?

I have a "manual" system currently in pace on my SUT. Whenever I expect a page change I store the current domain. Then at the end of testing I sort through the list and clear all the cookies from uniquely visited domains. I would like to have this automated (automatically detect domain changes) but I'm not sure how to do this without impacting run time.

aaltat:
If the performance impact is too high, then this PR should be closed. 

The problem is that many things can lead for domain changes and without tracking almost all keywords, it will be customised solution for you only. And also in this scenario, there would be performance hit.

Other way would that you use proxy or protocol sniffing and write a separate library for domain tracking. Then you would use that library information to clear the cookies from all domains.

aaltat:
Will you continue on working with the PR or should it be closed?

snyderr:
I think this should be closed until a better implementation can be made. I don't enough free time to work on this now. I will come back to this at a later date and nail down a more efficient solution. Currently the run time impact is too large to include this feature in the next release.

PR #826, opened 2017-06-19
Makes ElementFinder as context aware

aaltat:


aaltat:
This PR does not contains the `base.py` refactoring.

PR #835, opened 2017-06-21
Provides fix to keep API for adding custom locator strategy

aaltat:
Provides idea how to support custom locator strategy API. In the API it has four arguments ((${browser}    ${criteria}    ${tag}    ${constraints})) as like the old `ElementFinder,find` method line `return strategy(browser, criteria, tag, constraints)` required.

Some unit test fails, which reveal bug in my PR, but this PR is more to discuss is to solution good one or should we do something else?


aaltat:
Closing to favor #837

Issue #834, opened 2017-06-21
Capture Page Screenshot + Should Be Equal in Documentation


chanwooh:
For the keyword "Capture Page Screenshot"'s documentation, it uses the keyword "Should Be Equal" to imply that certain files should be equal. However, the keyword cannot actually be used to compare images or screenshots (it simply uses the arguments as strings in these cases and fails). 

I think the documentation should be clarified by noting that "Should Be Equal" in this case would not work, or remove the keyword from the example altogether.

aaltat:
I don't quite understand what you are trying to explain. Could you clarify the problem?

chanwooh:
![screen shot 2017-06-21 at 13 20 26](https://user-images.githubusercontent.com/14134553/27404862-719d82dc-5684-11e7-98cb-1fadb579fc43.png)

In Example 1 (shown above), "Should Be Equal" seems to be testing if screenshot ${file1} and screenshot ${${OUTPUTDIR}${/}selenium-screenshot-1.png} are equivalent as _images_. This is misleading since the keyword actually compares the _names_ of the screenshots for equality. That's why I think there should be a note that "Should Be Equal" cannot be used to compare the contents of ${file1} and ${${OUTPUTDIR}${/}selenium-screenshot-1.png}.

aaltat:
Now I understand what you are trying to say. I also understand why you might have come to this assumption. 

Would you have ideas how to improve the keyword documentation? Also perhaps in a pull request?

chanwooh:
I think just a clear note before the examples will suffice, since there isn't an easy "built-in" way to compare screenshots. I'll submit a pull request, thanks!

aaltat:
Issue should not be closed before we decide what to with the PR.

aaltat:
Closing

PR #836, opened 2017-06-22
Changed Capture Page Screenshot documentation for clarity

chanwooh:
Changed documentation to note that "Should Be Equal" keyword cannot be used to compare screenshots - from Issue #834

### DIFF
--- a/src/Selenium2Library/keywords/screenshot.py
+++ b/src/Selenium2Library/keywords/screenshot.py
@@ -74,24 +74,31 @@ class ScreenshotKeywords(LibraryComponent):
         exists, screenshot-2.png will be tried, and so on, until a
         unique name is found.
 
+        *Note: The keyword "Should Be Equal" in the following examples do not
+        compare the contents of ${file1} and
+        ${${OUTPUTDIR}${/}selenium-screenshot-1.png}, and only compares the
+        names of the screenshots for equality. In other words, "Should Be Equal"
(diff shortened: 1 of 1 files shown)


HelioGuilherme66:
My feelings are negative about this PR. The `Should Be Equal` keyword is from the BuiltIn library. The note should be shorter, just mentioning that is comparing the filenames. If we look at the whole help text, we see that we are talking about filenames. Also, the example came to illustrate the changes in the file naming.

aaltat:
I agree that this is not improvement to the right direction. But I agree that documentation could be improved. Would the documentation be clearer if the `Should Be Equal` would be replaces with: `Should Be Equal As Strings`. Or other option could be `BuiltIn.Should Be Equal`.

pekkaklarck:
In my opinion any note about explaining what Should Be Equal does just makes the docs of *this* keyword more complicated. I would either just leave the docs as they are or remove Should Be Equal and assigning the return value into a variable from the example altogether (assuming that the docs mention that the path is returned).

aaltat:
The keyword documentation says that path is returned.

@chanwooh do you want to give a second try with the enhancement?

chanwooh:
Sure, I'll tweak the examples to remove the parts that use `Should Be Equal`. I'll also remove the note, since it wouldn't be applicable anymore.

aaltat:
Closing due inactivity

PR #837, opened 2017-06-22
Fixes the changed Custom Locator API

aaltat:
Now the Custom Locator API uses the arguments:
${browser}, ${locator}, ${tag} and ${constraints}

Also moved usage of the CustomLocator to ElementFinder. And the
CustomLocator use the context for the browser.

### DIFF
--- a/src/Selenium2Library/locators/elementfinder.py
+++ b/src/Selenium2Library/locators/elementfinder.py
@@ -3,6 +3,7 @@ from robot.utils import NormalizedDict
 from selenium.webdriver.remote.webelement import WebElement
 
 from Selenium2Library.base import ContextAware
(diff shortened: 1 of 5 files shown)


src/Selenium2Library/locators/elementfinder.py
```diff
@@ -3,6 +3,7 @@
 from selenium.webdriver.remote.webelement import WebElement
 
 from Selenium2Library.base import ContextAware
+from Selenium2Library.locators.customlocator import CustomLocator
```
pekkaklarck, line 6: Should use `from .locators import ...` here.

aaltat: This is fixed in #840

PR #844, opened 2017-07-03
Fixes bug in _current_browser method in lib __init__

### LINKED ISSUE #842: In method _current_browser (in lib __init__) fix return self.browser to return self._browser
In https://github.com/robotframework/Selenium2Library/blob/master/src/Selenium2Library/__init__.py, this is now: 
```
    def _current_browser(self):
        warnings.warn('"Selenium2Library._current_browser" is deprecated, '
                      'use "Selenium2Library.browser" instead.',
                      DeprecationWarning)
        return self.browser
```
But it should be
```
    def _current_browser(self):
        warnings.warn('"Selenium2Library._current_browser" is deprecated, '
                      'use "Selenium2Library.browser" instead.',
                      DeprecationWarning)
        return self._browser
```
Fix the return statement

aaltat:
Fixes #842

### DIFF
--- a/src/Selenium2Library/__init__.py
+++ b/src/Selenium2Library/__init__.py
@@ -244,6 +244,6 @@ class Selenium2Library(DynamicCore):
 
     def _current_browser(self):
         warnings.warn('"Selenium2Library._current_browser" is deprecated, '
-                      'use "Selenium2Library.browser" instead.',
+                      'use "Selenium2Library._browser" instead.',
                       DeprecationWarning)
         return self._browser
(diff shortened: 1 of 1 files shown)


src/Selenium2Library/__init__.py
```diff
@@ -246,4 +246,4 @@ def _current_browser(self):
         warnings.warn('"Selenium2Library._current_browser" is deprecated, '
                       'use "Selenium2Library.browser" instead.',
```
ombre42: Warning should refer to `_browser` as well.

aaltat: Good catch. Fixed

PR #845, opened 2017-07-04
Dynamic api to string conversion fix

aaltat:


pekkaklarck:
I would consider trying to import the needed utils from `robot.utils` and only implementing them if they don't exist. Then the S2L implementation could be Python 2 only, because all the needed utils are in RF 3 and that's the first Python 3 compatible version. Once S2L's Robot dependency is raised to 3, then the S2L variants could be simply removed altogether.

aaltat:
Good idea. I will do it in the next commit.

aaltat:
I need to write tests for the change. Perhaps in the unit level . 

Also suggestion made by Pekka is not yet implemented.

PR #850, opened 2017-07-13
Adds run_on_failure method to the library instance

aaltat:
Also restores old _run_on_failure private method, which is now
deprecated.


aaltat:
I am not sure is the method documentation needed. I though it would be good idea but, now I having second though on the documentation.

Issue #851, opened 2017-07-17
Add keyword that checks focus


jsevern55:
This was reported and closed on : 
> Add a keyword to check focus #351

Adding a keyword such as `Element Should be focused` would be _very_ useful.

Is there a way to check this currently in RobotFramework?

Thank you for your time

aaltat:
That was closed because of inactivity in the issue, which usually means that need for the request has ended. Could you describe your need or requirements why you want to check a focus? Also could you describe what a focus means in your case?

jsevern55:
Thanks for the reply

I test div modals for w3c compliance (accessibility testing), and one of the many requirements is to verify visual/operational focus on forms is placed on the first available user modifiable element within the form.

When an element gains visual and operational focus (usually they go hand in hand), the user should be able to begin typing immediately for text fields, use up or down arrows for drop-down menus, or press the space bar for check boxes/radio buttons.

aaltat:
Sorry for the long delay but I am holidays and I have limited access to the internet.

I am unfamiliar with the topic, but I generally understand the topic. As far I recall the selenium API, the are not any direct support for this in the selenium API. Is the focus generally visible somehow in the DOM or how does it work?

jsevern55:
No worries at all. I just got back from a vacation myself.

If you reference this [stack overflow help topic](https://stackoverflow.com/questions/7552169/test-if-an-element-is-focused-using-selenium-webdriver), you'll see that selenium webdriver has a solution for this. I'm aware that robot framework utilizes selenium2library, but there is a solution in the selenium world.

Essentially, this function will return a boolean value if the expected element focused is the actual element focused: `element.equals(driver.switchTo().activeElement())`

aaltat:
I see your point. I think this would be good addition to the library. Few things to consider, from implementation point of view:
1) Now you are suggesting assert type of keyword, should there be also wait type of keyword?
2) Returning boolean is not necessary a good idea. If focus is not in the right element, raising error is better.

Would you be willing to submit a PR? If you are remember to write tests too?

WinterSoldier55:
1. Yes, there should be a "Wait for element focus" type keyword
2. If focus is not placed on the correct element, then raise an error

I don't think I understand your very last sentence:

> Would you be willing to submit a PR? If you are remember to write tests too?

What is "PR". Also, I do not understand the second question. I apologize...

aaltat:
[...]

PR #855, opened 2017-08-04
New keyword `Should Be Focused`

HelioGuilherme66:
See issue #851.

In reality the elements are not with Focus, but selected. If we do Mouse Over, we can see in the browser elements getting focus, but it would change the state to active.

More testing is required. It was only tested on Chrome, Firefox and PhantomJS. Would be nice to have it tested on Edge, IE 11, and Safari.

### DIFF
--- a/src/Selenium2Library/keywords/element.py
+++ b/src/Selenium2Library/keywords/element.py
@@ -223,15 +223,15 @@ class ElementKeywords(LibraryComponent):
         Key attributes for arbitrary elements are `id` and `name`. See
         `introduction` for details about locating elements.
 
-        New in SeleniumLibrary 2.0.1.
+        New in SeleniumLibrary 3.0.0.
         """
         element = self.find_element(locator)
-        try:
+        if self.browser.capabilities['browserName'] != "firefox":
+            focused = self.browser.switch_to.active_element
+        else:
             focused = self.browser.execute_script('return document.activeElement;')
-        except (IOError, RuntimeError) as err:
-            raise AssertionError("Internal Error: '%s'" % (err))
-        if not element == focused:
-            raise AssertionError("ERROR: Element '%s' is not with focus." % (locator))
+        if element != focused:
+            raise AssertionError("Element '%s' is not with focus." % (locator))
 
     @keyword
     def element_should_be_visible(self, locator, message=''):
--- a/src/Selenium2Library/keywords/element.py
+++ b/src/Selenium2Library/keywords/element.py
@@ -217,6 +217,23 @@ class ElementKeywords(LibraryComponent):
             raise AssertionError("Element '%s' is disabled." % (locator))
 
     @keyword
+    def element_should_be_focused(self, locator):
+        """Verifies that element identified with `locator` is focused.
+
+        Key attributes for arbitrary elements are `id` and `name`. See
+        `introduction` for details about locating elements.
+
+        New in SeleniumLibrary 2.0.1.
+        """
+        element = self.find_element(locator)
+        try:
+            focused = self.browser.execute_script('return document.activeElement;')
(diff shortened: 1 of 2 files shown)


src/Selenium2Library/keywords/element.py
```diff
@@ -217,6 +217,23 @@ def element_should_be_enabled(self, locator):
             raise AssertionError("Element '%s' is disabled." % (locator))
 
     @keyword
+    def element_should_be_focused(self, locator):
+        """Verifies that element identified with `locator` is focused.
+
+        Key attributes for arbitrary elements are `id` and `name`. See
+        `introduction` for details about locating elements.
+
+        New in SeleniumLibrary 2.0.1.
```
aaltat: Next release will be 3.0.0, could you change the text accordingly.

src/Selenium2Library/keywords/element.py
```diff
@@ -217,6 +217,23 @@ def element_should_be_enabled(self, locator):
...
+        """
+        element = self.find_element(locator)
+        try:
+            focused = self.browser.execute_script('return document.activeElement;')
+        except (IOError, RuntimeError) as err:
+            raise AssertionError("Internal Error: '%s'" % (err))
```
aaltat: Lines 229 - 232 could be replaced with this: https://github.com/SeleniumHQ/selenium/blob/fb6c35ed4974adc2c2f8308ee9d78fc4f6f3004a/py/selenium/webdriver/remote/switch_to.py#L34

src/Selenium2Library/keywords/element.py
```diff
@@ -217,6 +217,23 @@ def element_should_be_enabled(self, locator):
...
+        if not element == focused:
```
aaltat: if element != focused:

test/acceptance/keywords/element_focus.robot
```diff
@@ -0,0 +1,35 @@
+*** Settings ***
+Documentation     Tests Focus Verification and Wait for Focus
+Suite Setup       Open Browser To Start Page
+Test Setup        Go To Page "mouse/index.html"
+# Suite Teardown    Close All Browsers
```
aaltat: Is this line needed?

src/Selenium2Library/keywords/element.py
```diff
@@ -217,6 +217,23 @@ def element_should_be_enabled(self, locator):
...
+            raise AssertionError("ERROR: Element '%s' is not with focus." % (locator))
```
aaltat: Is the word `ERROR` needed in message?

aaltat (changes requested):
Looks pretty good, there are few things to consider.

src/Selenium2Library/keywords/element.py
HelioGuilherme66: I have seen that proposal, but it is for Remote webdriver. It was not implemented, so the only alternative is to use JavaScript.

test/acceptance/keywords/element_focus.robot
HelioGuilherme66: I won't be in the final version. I use it when running test file individually.

src/Selenium2Library/keywords/element.py
aaltat: Drivers inherits the remote, see the documentation example for the `Chrome`: http://selenium-python.readthedocs.io/api.html#module-selenium.webdriver.chrome.webdriver

Also this works fine: 
```
from selenium import webdriver

driver = webdriver.Chrome()
driver.get('http://robotframework.org/')
focus = driver.switch_to.active_element
print isinstance(focus, WebElement)
print focus
driver.quit()
```
Please change the code to use the `driver.switch_to.active_element`

aaltat (approved):

PR #866, opened 2017-08-18
Get Matching XPath Count - documentation isn't up-to-date

GLMeece:
Ran the document generator; this keyword's documentation is now updated.

You made the [code change I suggested](https://github.com/robotframework/Selenium2Library/pull/714) awhile back, but the documentation hasn't been generated since. I think the user base would appreciate the update. I often "warn" people on the Slack channel to make sure they cast the return to an **int**, but as it turns out they can just add the parameter `return_str=${False}`.

Since "no code was harmed in the making of this Pull Request" it should be easy to merge it in. :dark_sunglasses:


aaltat:
The public documentation is only updated when we make a release. But there is a good point and perhaps we should consider providing documentation in same manner as core framework provides the documents.

aaltat:
Because we are not going to update the documentation at this stage, I am closing the PR.

PR #881, opened 2017-09-01
Fixes Selenium2Library -> SeleniumLibrary in MANIFEST.in

aaltat:


aaltat:
I did, also checked the test folderand Selenium2Library was not found.

pekkaklarck:
Many files in project root still have it but most are also work-in-progress.

aaltat:
Yes, there was bunch of those files and therefore I did miss this one in the `MANIFEST.in`

Issue #882, opened 2017-09-02
Externally useful attributes should be declared public and documented.


pekkaklarck:
Old Selenium2Library has several attributes that are useful externally from custom libraries and extensions:

```
_run_on_failure_keyword
_speed_in_secs
_timeout_in_secs
_implicit_wait_in_secs
_current_browser()
```

We should indicate that using these externally is OK by removing the leading underscore. The current WebDriver instance should also be accessible as a normal attribute (i.e. without calling a method like `current_browser()`) and we should name the attribute `webdriver` to make it explicit what it actually is. I would also remove unnecessary `_in_secs` postfixes. In other words, I would like to have these attributes instead:

```
run_on_failure_keyword
speed
timeout
implicit_wait
webdriver
```

Renaming these attributes is obviously a backwards incompatible change. We should thus add propertys that enable the old functionality but emit deprecation warnings. These propertys can then be removed in 3.1 or later.

---

*UPDATES:*

1. We used `browser` attribute instead of `webdriver` until 3.0 RC 1.

2. In 3.0 RC 1 we added also methods `find_element` and `find_elements` directly to the library itself. These methods form the recommended public API to find elements for external tools.

pekkaklarck:
This is mostly fixed if #884 is merged. The remaining decisions to make are:

- Should deprecated attributes be unit tested somehow? Shouldn't be too hard, but adding tests for features that are going to be removed pretty soon feels a bit like waste.

- What do to with `_browsers`?

aaltat:
1) I think that it is waste too, lets not do unit test for those. 
2) Is the cache for drivers. There already are public API to register driver, get the current driver and switch between drivers and close driver. With those API it might be better to leave it private.

pekkaklarck:
If we aren't going to add more tests and we'll keep `_browsers` as-is, isn't this issue done?

aaltat:
Yes it's. Closing

pekkaklarck:
We decided to change `browser` to `webdriver' still. We'll deprecate `browser` in 3.0 rc 1 so that early adopters aren't punished, but we'll remove it still before the final release.

aaltat:
I think it would be better to use `driver` instead of `webdriver`. `webdriver` is module in selenium, which might confuse people, which we are referring to. Also all example in selenium uses word `driver` when the refer to `selenium.webdriver.Chrome`

It would be good idea to change also `BrowserCache` to `DriverCache`

[...]

Issue #176, opened 2013-03-04
Return value of `Register keyword to run on failure` cannot always be used to restore original state


stevejefferies:
There seems to be an issue with the implementation of the "Register keyword to run on failure" keyword. The keyword returns the value of the previously registered keyword which works fine if this is set to a keyword e.g. "Capture Page Screenshot" but if the registered keyword is set to "None" the return value is the string "No keyword". If you then try to use the return value "No keyword" and pass this back into "Register keyword to run on failure" a warning is generated: 

"Keyword 'No keyword' could not be run on failure: No keyword with name 'No keyword' found."

I guess this keyword should either be returning "Nothing" as a string rather than "No keyword" to be consistent with the current way of disabling the run-on-failure functionality or it could return None instead

Example test case (simplistic example):

```
*** Settings ***
Library           Selenium2Library

*** Test Cases ***
Test Case 1
    ${previous_kw}=    Register Keyword To Run On Failure    Nothing
    ${previous_kw}=    Register Keyword To Run On Failure    Capture Page Screenshot
    Register Keyword To Run On Failure    ${previous_kw}
    Open Browser    http://www.google.com
    Page Should Contain    invalid_string
```

The implementation could be changed to the following which would return None from line 2 of the above test case and no warnings are observed on failure:

```
def register_keyword_to_run_on_failure(self, keyword):
    """Documentation here....
    """
    old_keyword = self._run_on_failure_keyword
    old_keyword_text = old_keyword if old_keyword is not None else "No keyword"

    if keyword:
        new_keyword = keyword if keyword.strip().lower() != "nothing" else None
    else:
        new_keyword = None
    new_keyword_text = new_keyword if new_keyword is not None else "No     keyword"

    self._run_on_failure_keyword = new_keyword
    self._info('%s will be run on failure.' % new_keyword_text)

    return old_keyword
```

ombre42:
I think that changing the return value is an unnecessary and a riskier change. When this was first implemented, it should have returned "nothing", but it seems too late to change that. If we consider "No keyword" and "nothing" to be synonymous, the warning will go away.

boakley:
I just stumbled on this today. This really ought to be fixed. I shouldn't have to add special case code in my test when saving and restoring the keyword. If changing the return value is unacceptable, perhaps changing it so that passing in "No Keyword" has the same effect as passing in "Nothing".

[...]

PR #884, opened 2017-09-02
Cleanup

pekkaklarck:
1. Little import cleanup
2. Fix registering keyword to run on failure (#176)
3. Make public attributes actually public (#882)

This PR contains somewhat unrelated fixes, but commits ought to mainly contain only a single fix.


aaltat:
Looks good on mobile screen, need to take closer look from bigger screen. Spotted one thing though.

PR #885, opened 2017-09-02
Removes SeleniumLibrary Firefox profile

### LINKED ISSUE #883: Remove SeleniumLibrary profile for Firefox
The SeleniumLibrary has long contained a custom Firefox profile, but the profile is based on very old version of Firefox. Also the profile was last updated in 2011. I would be better to use the default profile provided by the Selenium [FirefoxProfile](https://seleniumhq.github.io/selenium/docs/api/py/webdriver_firefox/selenium.webdriver.firefox.firefox_profile.html?highlight=firefoxprofile#selenium.webdriver.firefox.firefox_profile.FirefoxProfile). 

The change is possibly backwards incompatible and therefore added bwic label.

aaltat:
The SeleniumLibrary default Firefox profile is now removed. When
Firefox created and if Firefox profile is not defined, then the
default Selenium FirefoxProfile is used.

Fixes #883


src/SeleniumLibrary/base/librarycomponent.py
```diff
@@ -56,3 +59,15 @@ def assert_page_not_contains(self, locator, tag=None, message=None,
...
+    def log_dir(self):
+        try:
+            logfile = BuiltIn().get_variable_value('${LOG FILE}')
+        except RobotNotRunningError:
+            logfile = os.getcwd()
+        if logfile != 'NONE':
+            logdir = os.path.dirname(logfile)
+        else:
+            logdir = BuiltIn().get_variable_value('${OUTPUTDIR}')
+        return logdir
```
pekkaklarck: 1. It took me a little time to make sure `RobotNotRunningError` cannot occur when querying `${OUTPUTDIR}`. I think the code should be reorganized a little.
2. There seems to be a logic bug when you actually get that `RobotNotRunningError`. The returned directory seems to be the parent directory of the cwd in that case.
3. `os.getcwd()` returns bytes on Python 2. This causes problems if the path contains non-ASCII characters. Should either use `os.getcwdu()` (which doesn't exist on Python 3) or just return `'.'`.

Perhaps the code could be something like this:
```python
try:
    logfile = BuiltIn().get_variable_value('${LOG FILE}')
    if logfile == 'NONE':
        return BuiltIn().get_variable_value('${OUTPUTDIR}')
    return os.path.dirname(logfile)
except RobotNotRunningError:
    return os.getcwd() if PY3 else os.getcwdu()
```

src/SeleniumLibrary/utils/seleniumversion.py
```diff
@@ -0,0 +1,8 @@
+from collections import namedtuple
+
+import selenium
+
+SeleniumVersion = namedtuple('SeleniumVersion', 'major minor micro')
+version = selenium.__version__.split('.')
+selenium_version = SeleniumVersion(major=version[0], minor=version[1],
+                                   micro=version[2])
```
pekkaklarck: - Copyrights missing from the file.
- +1 for namedtuple.
- Like you commented elsewhere, SELENIUM_VERSION is better name for the constant.
- Can we be certain that version always has three components? If not, this could be used:
   ```python
   major, minor, micro = (selenium.__version__.split('.') + ['0', '0'])[:3]
   SELENIUM_VERSION = SeleniumVersion(major, minor, micro)
   ```
   If earlier releases have always contained three components, the above is an overkill.

pekkaklarck (approved):
Looks good! Please take a look at the line notes, though.

src/SeleniumLibrary/base/librarycomponent.py
aaltat: Agreed and done.

src/SeleniumLibrary/utils/seleniumversion.py
aaltat: Looked releases all to they to the 2.50.0 and there always has been three components, in the version field. But in one day, there might not be, so why risk it.

Others done.

PR #895, opened 2017-09-09
New tooling to ease creating releases

pekkaklarck:
This is related to #634. Tasks included into the PR:

- [x] Take https://github.com/robotframework/rellu into use to ease creating release notes, setting version, etc. This includes updating requirements-dev.txt accordingly.
- [x] Update BUILD.rst.
- [x] Fix including required packages into distribution in setup.py.
- [x] Remove outdated and unnecessary files (build_dist.py, setup.cfg, SELENIUM_COPYRIGHT.txt)

This PR is now ready for review. The upadted BUILD.rst can be seen in rendered format here:
https://github.com/pekkaklarck/SeleniumLibrary/blob/rellu/BUILD.rst

Known missing pieces that can be added later, possibly even after Se 3.0 final:

- Add documentation to Invoke tasks. This is actually covered by robotframework/rellu#7.
- Consider adding something more about testing, at least about CI, into BUILD.rst. If atest/README.rst is very good, the current info may even be adequate.


aaltat:
There are some `rellu` related errors in the Travis.

pekkaklarck:
Seems to occur only with Python 2.7. Does Travis try to install rellu which is only Python 3(.6) compatible? If yes, is it due to Travis using requirements-dev.txt? If yes, perhaps we need to split rellu and related tools to separate requirements-build.txt?

pekkaklarck:
It seems that splitting build requirements to `requirements-build.txt` fixed the issue. Also enhanced BUILD.rst a little with the latest commits.

tasks.py
```diff
@@ -1,17 +1,79 @@
...
+
+**REMOVE ``--pre`` from the next command with final releases.**
+If you have pip_ installed, just run
+``pip install --pre --upgrade robotframework-seleniumlibrary``
+to install the latest release or use
+``pip install robotframework-seleniumlibrary=={version}``
+to install exactly this version. Alternatively you can download the source
+distribution from PyPI_ and install it manually.
+
+Rellu {version} was released on {date}.
```
aaltat: Rellu -> SeleniumLibrary

aaltat (commented):
One small thing to fix. All other looks good.

tasks.py
pekkaklarck: Guess did I copy this from Rellu's tasks.py? =)

I can fix this later today or you can merge and fix yourself.

PR #897, opened 2017-09-14
896 enhancement open element in new tab and new window

ponkar:
This Issue will contribute for 2 Keywords -

Open In New Tab
Open In New Window

Application - There are times when we want to perform some operations in new window/tab of the same application we testing. With this what can be tested is the operations which are performed in new window can be verified if they are reflecting in master window.


ponkar:
@pekkaklarck @aaltat  please can you review this PR and let me know your views on it.

pekkaklarck:
The code simply holds down the shift key while clicking an element, right? Isn't it up to the browser to decide how it reacts to that? If yes, calling this action `Open In New Window` isn't right.

Having a way to click elements while holding shift and possibly some other modifiers would anyway be useful. Needed to design such keywords carefully to make them generic, though. One possibility is adding new configuration options to the existing `Click Element`.

Regardless what we decide to do with this, I would say this needs to wait post SL 3.0 final.

aaltat:
I agree with @pekkaklarck that having own keyword for this is not good idea. Instead the other `Click *` keywords should be enhanced to support this type of behaviour.

Would you be interested creating a such PR if we can first agree on the design?

ponkar:
I agree over this
Yes we can think in a direction to reuse the click element keyword with an
extra flag like parameter eg new_tab=true or false with false as default
value.

aaltat:
I will close this PR and let's discuss this more on the issue tracker.

Issue #816, opened 2017-06-02
Modify Capture Page Screenshot keyword not fail if browser is not open. 


aaltat:
For new users, it might be confusing when the `Open Browser` keyword fails due some invalid configuration and then also the `Capture Page Screenshot` keyword also fails. It might be good idea not to raise exception if browser is not open when running `Capture Page Screenshot` keyword

alexandrul:
What should it do in this case? Get a screenshot of the desktop (hoping for some helpful message dialogs) or create an empty file?

pekkaklarck:
I was initially thinking it should just log a message that a screenshot cannot be taken because browser is not open. Taking a screenshot of the desktop is a valid idea too. Should probably somehow separate that from normal case anyway.

pekkaklarck:
The simplest solution could be adding something like this to the beginning of the keyword:

```python
if not self.browser:
    self.info('Cannot take screenshot because browser is not not open')
    return
```

but, as already discussed, taking a screenshot of the entire desktop might be even better. Either way, I don't think this needs any documentation changes but a new acceptance test is needed.

Ar you @alexandrul interested to look at this?

aaltat:
Taking a screen shot from a desktop might be problematic, because there might not be a desktop running. Example if using headless chrome or PhantomJS. Also there are other scenarios which might prevent taking a screen shot from a desktop, although it might be running. Example running the browser using the Selenium grid causes some problems.

Perhaps it would be best to split the issue in two parts. First make the change Pekka suggested, because it's easy to do and there is a low risk doing it. Then in separate issue, investigate is it possible to find a solution that would work in most use cases and in cross platform way.

pekkaklarck:
It's a good point that it's not always possible to take a screenshot of the desktop. Let's not complicate this unnecessarily and just do the simple thing that avoids the annoying warning. I would even open a separate issue about taking a screenshot of the desktop unless someone really needs such functionality. It's anyway possible to use Robot's Screenshot library for that purpose if needed.

aaltat:
The original issue is now done. If there is a need to take a screen capture from the desktop, separate issue should be raised.

PR #899, opened 2017-09-15
No Fail For Capture Page Screenshot When No Browser Is Opened (Issue #816)

olivierverville:
The keyword `Capture Page Screenshot` will no longer fail, and simply return with a message when no browser is opened.

### DIFF
--- a/src/SeleniumLibrary/keywords/screenshot.py
+++ b/src/SeleniumLibrary/keywords/screenshot.py
@@ -113,13 +113,13 @@ class ScreenshotKeywords(LibraryComponent):
 
         """
         try:
-			# try to access the browser property
-			b = self.browser
-			
-		except RuntimeError:
-			# we got an error, just exit
-			self.info("Couldn't capture page screenshot because no browser is opened")
-			return
+            # try to access the browser property
+            b = self.browser
+            
+        except RuntimeError:
+            # we got an error, just exit
+            self.info("Couldn't capture page screenshot because no browser is opened")
+            return
         
         path, link = self._get_screenshot_paths(filename)
         self._create_directory(path)
--- a/src/SeleniumLibrary/keywords/screenshot.py
+++ b/src/SeleniumLibrary/keywords/screenshot.py
@@ -112,9 +112,14 @@ class ScreenshotKeywords(LibraryComponent):
         | File Should Exist | ${OTHER_DIR}${/}sc-000001.png |
 
         """
-        if not self.browser:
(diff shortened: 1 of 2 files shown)


aaltat:
Please also write a test and use status checker [1] to verify message

[1] https://github.com/robotframework/statuschecker/blob/master/README.rst

olivierverville:
Not sure how to setup a test for this. If I close the browser, it will break other tests since they rely on the global suite setup to open their browser. Two ideas:

**1)** Call `Close All Browsers` in my new test case's setup, call `Capture Page Screenshot` within the test case itself, and call the suite's setup keyword back in my test case's teardown

**2)** Simply add a `Capture Page Screenshot` step to the global teardown. (after the browser(s) have/has been closed)

Thoughts?

aaltat:
The first option is best. If I recall correctly, there are other test that do the similar thing.

olivierverville:
I've added an acceptance test, and according to Travis it passes, but the build fails. The log doesn't really help, I can't seem to find what crashes.

src/SeleniumLibrary/keywords/screenshot.py
```diff
@@ -112,6 +112,15 @@ def capture_page_screenshot(self,
...
         """
+        try:
+            # try to access the browser property
+            b = self.browser
+            
+        except RuntimeError:
+            # we got an error, just exit
+            self.info("Couldn't capture page screenshot because no browser is opened")
+            return
+
```
pekkaklarck: Comments and empty lines are unnecessary and somewhat against the style used elsewhere in this codebase. I'd also say it's clearer to just have

```python
try:
    self.browser
except RuntimeErrror:
  # ...
```

than

```python
try:
    b = self.browser
except RuntimeErrror:
  # ...
```
when the `b` variable isn't used for anything. Could also consider using `try/except/else` instead of returning from `except`.

It might actually also be a good idea to create new `BrowserNotOpen` exception and the `browser` property to use it instead of the generic `RuntimeError`. The code here would then look like

```python
try:
    self.browser
except BrowserNotOpen:
  # ...
```

which would be really clean. Not sure is it worth it, though.

pekkaklarck:
The failure is most likely caused by the Status checker tool finding a problem when it does post-execution analysis. Do you get that failure also locally?

src/SeleniumLibrary/keywords/screenshot.py
olivierverville: Good suggestions. I'll get back to this on Monday

olivierverville:
Here we go :)

PR #910, opened 2017-09-20
Fixes Set Selenium Speed when calling before opening browser

### LINKED ISSUE #898: "Set Selenium Speed" doesn't work when called before opening browser in release 3.0.0b1
Hello,
I have some test cases wrote by RF script and Selenium2Library(1.8.0) and worked fine.
However after upgrading Selenium2Library to 3.0.0b1, the keyword "Set Selenium Speed" doesn't work anymore, operating the browser by selenium becomes so fast that cause my original test cases failed.

Using keyword "Get Selenium Speed" after setting it, the result is the one I set it:
>${Speed} = Selenium2Library.Get Selenium Speed
>INFO 	${Speed} = 0 seconds
>Selenium2Library.Set Selenium Speed 0.3 s
>${Speed} = Selenium2Library.Get Selenium Speed
>INFO 	${Speed} = 300 milliseconds

But observing the real behaviour on the browser, seems the speed is 0.

Therefore I need to downgrade Selenium2Library back to 1.8.0, and everything back to normal, browser operating speed is actually what I set.

**Environment:**
1. OS: Windows 7 64 bits
2. Python 2.7.11
3. Original Selenium2Library: 1.8.0
4. New Selenium2Library: 3.0.0b1
5. Browser: Chrome 61.0.3163.79 (64 bits)
6. Web driver: ChromeDriver 2.32.498550

aaltat:
If Set Selenium Speed was called before the browser
is opened, the speed did not affect the browser. This PR fixes the
problem.

Fixes #898

### DIFF
--- a/src/SeleniumLibrary/keywords/browsermanagement.py
+++ b/src/SeleniumLibrary/keywords/browsermanagement.py
@@ -143,7 +143,7 @@ class BrowserManagementKeywords(LibraryComponent):
                                      ff_profile_dir, remote_url)
         try:
             browser.get(url)
-        except:
+        except Exception:
             self.ctx.register_browser(browser, alias)
             self.debug("Opened browser with session id %s but failed "
                        "to open url '%s'" % (browser.session_id, url))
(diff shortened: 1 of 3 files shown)


aaltat:
Use `Should Be True    ${end} - ${start} >= 2` instead of `Evaluate` and `Should Be True` combination

Issue #911, opened 2017-09-21
Continue to allow access to _table_element_finder class


GLMeece:
Per discussion with @pekkaklarck on the Slack group...

## Essential Issue

Having created some Python helper functions, using the **Selenium2Library**, I have been able to leverage some of the private methods to accomplish some useful things for my testing framework. An example below shows a function `get_column_number` which used to be able to return the column number of whatever (valid) column _name_ was passed in:

```python
from robot.libraries.BuiltIn import BuiltIn

def _get_sel2lib():
    """== Gets Current Selenium 2 Instance from Robot Framework ==
    
    - Uses the Robot Framework API to get an object of the current Selenium 2 instance.
    - This is an internal helper function only. 
    
    == Calling ==
    
    | *Args* | [none] | |
    | *Returns* | ``object`` | An object instance of the current Selenium 2 library.
    | *Raises* | [none] | |
    """
    return BuiltIn().get_library_instance('Selenium2Library')

def get_column_number(table_locator, col_text, loglevel='INFO'):
    """== Returns Number of Specified Column ==
    
    Returns the number of the first column found which contains the ``col_text`` string.

    - Does not require an exact match; if there is a column named ``Foobar`` and you input ``Foo`` then it will match.
    
    === Calling ===
    
    | *Args* | ``table_locator`` (str) | The table locator containing the column whose number you desire. |
    |        | ``col_text`` (str) | The string of the column you want to locate. |
    | *Returns* | ``int`` | Number of column containing string ``col_name``. |
    | *Raises* | AssertionError | If ``col_name`` cannot be found in the table specified. |

    === Example in Robot ===

    | ${column_number} =    Get Column Number    xpath=//*[@id="deviceListHeader"]    Unsafe
    """
    sel2lib = _get_sel2lib()
    locators = sel2lib._table_element_finder._parse_table_locator(table_locator, 'header')
    found_it = False
    for locator in locators:
        elements = sel2lib._element_finder.find(sel2lib._current_browser(), locator)
        counter = 0
        for element in elements:
            counter += 1
            if col_text in element.text:
                output = counter
                found_it = True
                break
    if found_it is False:
        sel2lib.log_source(loglevel)
        raise AssertionError("No column containing '{}' found in the table identified via {}!".format(col_text, table_locator))
    return output
```

[...]

PR #912, opened 2017-09-21
Add table element finder in private API

aaltat:


aaltat:
This is for #911

test/unit/api/test_public_api_objects.py
```diff
@@ -0,0 +1,19 @@
+import unittest
+
+from SeleniumLibrary import SeleniumLibrary
+from SeleniumLibrary.locators import ElementFinder
+from SeleniumLibrary.locators import TableElementFinder
```
pekkaklarck: Last two lines could (and IMHO should) be combined:

    from SeleniumLibrary.locators import ElementFinder, TableElementFinder

src/SeleniumLibrary/__init__.py
```diff
@@ -32,6 +32,7 @@
                                       TableElementKeywords,
                                       WaitingKeywords)
 from SeleniumLibrary.locators import ElementFinder
+from SeleniumLibrary.locators import TableElementFinder
```
pekkaklarck: Could (and IMHO should) be combined with the above import:

    from SeleniumLibrary.locators import ElementFinder, TableElementFinder

pekkaklarck (approved):
Code changes look good to me. Always nice when the diff has more red than green. I added two nitpick comments about imports, though.

There were lot of changes to tests, apparently due to indentation changes. Didn't bother to look at them too closely now, but tests seem to pass so they ought to be good too.

aaltat:
The changes in the unit test are caused because the ElementFinder is not anymore a instance of the class. Now it's inherited and comes through few properties and mocking was easier this way.

test/unit/api/test_public_api_objects.py
aaltat: Done

aaltat:
Fixed the imports and I should really stop doing imports that way.

Issue #914, opened 2017-09-22
Desing better interface for table keywords


aaltat:
The `TableElementFinder` is poorly designed and it needs refactoring. The  `TableElementFinder`  does not support all the same locator strategies which are supported by the rest of library. It would be better to redesign the whole class to be better unified with the rest of the library. 

Improvements for `TableElementFinder`
- [ ] Add more acceptance tests
- [ ] It should not matter which locator strategy is used to identify the table where actions are performed
- [ ] Support also webelements
- [ ] When identifying the table, support also identifying table by element above, example if DOM looks like this `<div id="table"><table>...</table></div>` then ` | Table Should Contain | id=table | item |` should work
- [ ] Design is is possible to return table content to Robot side. It should be possible for simple tables like 

```
<table>
  <tr>
    <th>Firstname</th><th>Lastname</th><th>Age</th>
  </tr>
  <tr>
    <td>Jill</td><td>Smith</td><td>50</td>
  </tr>
</table>
```
But with some complex table, example with nested cells and tables it might be really difficult. Also when table cell is empty, finding cells might be difficult.

PR #916, opened 2017-09-23
Add support for parent element for ElementFinder and TableElementFinder to support normal locator strategies

aaltat:
This helps to part of fixing the #914


aaltat:
This needs more tests, specially on those methods which are implemented without direct support from the selenium API.

aaltat:
Will not be done like this

Issue #908, opened 2017-09-18
New `strategy:value` syntax to specify locator strategy in addition to current `strategy=value`


pekkaklarck:
Currently when specifying a non-default locator for an element, the syntax is `name=value` like in `xpath=//h1` or `css=h1`. This works well otherwise, but the same syntax is also used by [Robot Framework's named argument syntax](http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#named-argument-syntax) and this can cause confusion.

Using the same syntax also causes actual problems if a keyword would accept `**kwargs`, because then Robot considers all arguments looking like `name=value` to match it. There's also a possibility that some time in the future Robot will always do that, regardless on the function signature. Robot allows escaping the problematic `=` like in `name\=value` (e.g. `css\=h1`) but that is a bit ugly if needed often.

My proposal is that in addition to the existing `=`, we allow using `:` like `name:value` (or `name: value` or even `name : value` as we allow spaces to used with `=` too). The reason I propose `:` is that it's visually quite close to `=`, it's used by Robot on the command like (e.g. `-v name:value`), and it's also used by Python in similarly as `=` in some contexts (e.g.`{'name': 'value'}` and `dict(name=value)`). We don't need to deprecate the current syntax at least now, but documentation should be written so that the new syntax is recommended.

This change causes backwards incompatibility problems if someone has used a default locator (e.g. id) which has `:` in its value like `Click Element    foo:bar`. We don't expect `:` to be too common in locator values, and there's always a possibility to use `id:foo:bar` or `id=foo:bar` to avoid the problem. If there are lot of complaints during the SL 3.0 release candidate phase, we can still consider ways to make the transition smoother.

pekkaklarck:
This issue is now otherwise done, but table related keywords still support only the `strategy=value` syntax. This is due to table keywords having their own locator logic. That's a bad idea in general and #923 aims to fix it.

aaltat:
After the #923 is merged the #926 updates the documentation.

pekkaklarck:
PR #923 fixed inconsistencies with table locators and #926 updated related docs. This issue can be considered done.

PR #918, opened 2017-09-25
Generic library documentation update

pekkaklarck:
Includes explaining the new `strategy:value` syntax (#908).

Still plan to update the documentation of the `__init__`, but going through individual keywords is too much work for 3.0.

### DIFF
--- a/src/SeleniumLibrary/__init__.py
+++ b/src/SeleniumLibrary/__init__.py
@@ -42,133 +42,190 @@ __version__ = '3.0.0b1'
 class SeleniumLibrary(DynamicCore):
     """SeleniumLibrary is a web testing library for Robot Framework.
 
-    This document is about using SeleniumLibrary. For information about
-    installation, support, and more please visit the
-    [https://github.com/robotframework/SeleniumLibrary|project page].
+    This document explains how to use keywords provided by SeleniumLibrary.
+    For information about installation, support, and more, please visit the
+    [https://github.com/robotframework/SeleniumLibrary|project pages].
+    For more information about Robot Framework, see http://robotframework.org.
 
-    SeleniumLibrary uses the Selenium 2 (WebDriver) libraries internally to control a web browser.
-    See http://seleniumhq.org/docs/03_webdriver.html for more information on Selenium 2
-    and WebDriver.
+    SeleniumLibrary uses the Selenium WebDriver modules internally to
+    control a web browser. See http://seleniumhq.org for more information
+    about Selenium in general.
 
-    SeleniumLibrary runs tests in a real browser instance. It should work in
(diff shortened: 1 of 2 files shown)


src/SeleniumLibrary/__init__.py
```diff
@@ -42,133 +42,190 @@
...
+    | `Click Link`    | example | # Match also based on link text and ``href``.   |
+    | `Click Button`  | example | # Match based on ``id``, ``name`` or ``value``. |
+
+    If the locator accidentally starts with some of the explicit locator
+    prefixes or with the implicit XPath prefixes, it is possible to use
+    the explicit ``default`` prefix to enable the default strategy.
+
+    Examples:
+
+    | `Click Element` | name:foo         | # Find element with class ``foo``.               |
```
aaltat: I thin comment should say: 
```
Find element with name attribute ``foo``
```

src/SeleniumLibrary/__init__.py
```diff
@@ -42,133 +42,190 @@
...
+    | `Click Element` | default:name:foo | # Use default strategy with value ``class:foo``. |
+    | `Click Element` | //foo            | # Find element using XPath ``//foo``.            |
+    | `Click Element` | //foo            | # Use default strategy with value ``//foo``.     |
```
aaltat: The locator is same in the above xpath example but the comment says something else, should the locator be just `foo` without the double forward slashes.

src/SeleniumLibrary/__init__.py
```diff
@@ -42,133 +42,190 @@
...
 
     = Timeouts =
 
-    There are several `Wait ...` keywords that take timeout as an
+    There are several ``Wait ...`` keywords that take timeout as an
     argument. All of these timeout arguments are optional. The timeout
-    used by all of them can be set globally using the
-    `Set Selenium Timeout` keyword. The same timeout also applies to
-    `Execute Async Javascript`.
+    used by all of them can be set globally using the `Set Selenium Timeout`
```
aaltat: Should we also say that timeout can be set in the library import?

aaltat (changes requested):

src/SeleniumLibrary/__init__.py
pekkaklarck: Yeah, I changed the example but apparently forget the comment. Will fix.

src/SeleniumLibrary/__init__.py
pekkaklarck: It should be `default://foo`. Will fix.

src/SeleniumLibrary/__init__.py
pekkaklarck: Good idea. Will add.

aaltat (approved):

Issue #892, opened 2017-09-06
Improve Documentation of Select Frame and Select Window


HelioGuilherme66:
When selecting a child frame (for example inside a `frameset`), we must go back to the main frame. There is no keyword for this, except `Select Window` without arguments. I propose to add text to `Select Window` to explain this, and to `Select Frame` to have reference to `Select Window`.
Add a note that frame or iframe are the same :)

Alternatively, modify `Select Frame` to accept `locator=None` and behave like `Select Window`.

aaltat:
There is `Unselect Frame` which should return to the main, if I recall the code correctly. If using the unselect keyword is not clear from the ` Select Frame` keyword, the documentation could be enhanced.

aaltat:
Done in #919

pekkaklarck:
Enhancing documentation in general is covered by #925. I'll mark this as its duplicate.

PR #919, opened 2017-09-26
Improves documentation for `Select Frame` and `Unselect Frame`.

HelioGuilherme66:
For #892.

~~REVIEW required - I can't get the example correct.~~

Added note on having to Unselect Frame to start searching from main frame.

Expanded examples.

### DIFF
--- a/src/SeleniumLibrary/keywords/browsermanagement.py
+++ b/src/SeleniumLibrary/keywords/browsermanagement.py
@@ -327,23 +327,11 @@ class BrowserManagementKeywords(LibraryComponent):
         
         Please note that the frame search always start from the document root or main frame.
         
-        Examples:
-        | # Two frame elements with equal names | | |
-        | Select Frame   | xpath=`/``/`frame[@name='left']`/`iframe[@name='left'] | # Selects the 'left' iframe |
-        | Click Link     | foo                                                    | # Clicks link 'foo' in 'left' iframe |
-        | Unselect Frame |                                                        | # Returns to main frame |
-        | Select Frame   | left                                                   | # Selects the 'left' frame |
-  
-        | # Multiple frames with unique names or id | | |
-        | Select Frame   | third  | # Selects the 'third' frame |
-        | Click Link     | bar    | # Clicks link 'bar' in 'third' frame |
-        | Unselect Frame |        | # Returns to main frame |
-        | Select Frame   | first  | # Selects the 'first' frame |
-        | Click Link     | bar    | # Clicks link 'bar' in 'first' frame |
-        | Unselect Frame |        | # Returns to main frame |
-        | Select Frame   | second | # Selects the 'second' frame |
-        | Click Link     | bar    | # Clicks link 'bar' in 'second' frame |        
-        
+        Example:
+        | Select Frame   | xpath=`/``/`frame[@name='top]`/`iframe[@name='left'] | # Selects the 'left' iframe |
+        | Click Link     | foo                                                  | # Clicks link 'foo' in 'left' iframe |
+        | Unselect Frame |                                                      | # Returns to main frame |
+        | Select Frame   | left                                                 | # Selects the 'top' frame |        
         """
         self.info("Selecting frame '%s'." % locator)
         element = self.find_element(locator)
@@ -411,7 +399,7 @@ class BrowserManagementKeywords(LibraryComponent):
     def unselect_frame(self):
         """Sets the top frame as the current frame.
         
-        It does the same as `Select Window` without arguments, returning to the Main frame.
+        Cancels a previous `Select Frame` call, returning to the Main frame.
         """
         self.browser.switch_to.default_content()
 
--- a/src/SeleniumLibrary/keywords/browsermanagement.py
+++ b/src/SeleniumLibrary/keywords/browsermanagement.py
@@ -328,10 +328,10 @@ class BrowserManagementKeywords(LibraryComponent):
         Please note that the frame search always start from the document root or main frame.
         
         Example:
-        | Select Frame   | xpath=`/``/`frame[@name='top]`/`iframe[@name='left'] | # Selects the 'left' iframe |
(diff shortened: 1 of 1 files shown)


src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -387,7 +396,10 @@ def list_windows(self):
 
     @keyword
     def unselect_frame(self):
-        """Sets the top frame as the current frame."""
+        """Sets the top frame as the current frame.
+        
+        It does the same as `Select Window` without arguments, returning to the Main frame.
```
aaltat: By reading the code, for me it looks like `Select Window` and unselect frame are not the same things. `Select Window` is able to switch between different windows and unselect frame just returns to the top most frame set.

aaltat (changes requested):

src/SeleniumLibrary/keywords/browsermanagement.py
HelioGuilherme66: I specified "without arguments" for this case.

HelioGuilherme66:
This is how the help looks now.
imagem

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -322,6 +322,28 @@ def select_frame(self, locator):
...
+        Please note that the frame search always start from the document root or main frame.
+        
+        Examples:
+        | # Two frame elements with equal names | | |
+        | Select Frame   | xpath=`/``/`frame[@name='left']`/`iframe[@name='left'] | # Selects the 'left' iframe |
+        | Click Link     | foo                                                    | # Clicks link 'foo' in 'left' iframe |
+        | Unselect Frame |                                                        | # Returns to main frame |
+        | Select Frame   | left                                                   | # Selects the 'left' frame |
+  
+        | # Multiple frames with unique names or id | | |
```
aaltat: I do not think that having this example brings value compared to the first example. Please remove it.

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -322,6 +322,28 @@ def select_frame(self, locator):
...
         Key attributes for frames are `id` and `name.` See `introduction` for
         details about locating elements.
+        
+        See `Unselect Frame` to cancel the frame selection and return to the Main frame.
+        
+        Please note that the frame search always start from the document root or main frame.
+        
+        Examples:
+        | # Two frame elements with equal names | | |
+        | Select Frame   | xpath=`/``/`frame[@name='left']`/`iframe[@name='left'] | # Selects the 'left' iframe |
```
aaltat: Would it be better just say: 
```
| Select Frame | xpath=//iframe[@name='left'] | # Selects the 'left' iframe |
```

src/SeleniumLibrary/keywords/browsermanagement.py
aaltat: It is wrong to indicate that they are the same thing, because they are not. Could you remove this update.

aaltat (changes requested):

src/SeleniumLibrary/keywords/browsermanagement.py
HelioGuilherme66: Yes, if we do not look at the rest of the example. Then we select by name the first found frame without xpath. The first example is to show the structure.

src/SeleniumLibrary/keywords/browsermanagement.py
HelioGuilherme66: I don't say they are the same things.
How about saying?

> It results in the same effect as `Select Window` without arguments, returning to the Main frame.

src/SeleniumLibrary/keywords/browsermanagement.py
HelioGuilherme66: This example shows (was meant to show) that having three frames on the same level, one has to unselect them to return to search point. Did not want to complicate more the example by showing the frame's structure. But maybe we should?

src/SeleniumLibrary/keywords/browsermanagement.py
aaltat: I think the current documentation is good. Why we should tell about Select Window keyword.

src/SeleniumLibrary/keywords/browsermanagement.py
aaltat: But that is already shown in the first example. Select frame, do something and unselect the frame. Having second similar example doesn't add value.

src/SeleniumLibrary/keywords/browsermanagement.py
aaltat: In this  KW it is not good idea to teach people to select similar elements from the DOM. Example must be clear and easy to understand.

pekkaklarck:
FWIW, I tend to agree with comments by @aaltat. Examples in keyword documentation ought to be simple and short. Separate documentation, possibly in form of an executable demo, can be used to explain more complex usages.

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -322,6 +322,16 @@ def select_frame(self, locator):
 
         Key attributes for frames are `id` and `name.` See `introduction` for
         details about locating elements.
+        
+        See `Unselect Frame` to cancel the frame selection and return to the Main frame.
+        
+        Please note that the frame search always start from the document root or main frame.
+        
+        Example:
+        | Select Frame   | xpath=`/``/`frame[@name='top]`/`iframe[@name='left'] | # Selects the 'left' iframe |
```
aaltat: The `xpath=//...` syntax is problematic from the documentation point of view. Also we like to promote the new separator `:` in locators. It could be good idea to write the xpath like this `xpath: //...`

aaltat (changes requested):
Looks good, expected one small thing.

aaltat (approved):

PR #921, opened 2017-09-26
Add support for parent element for ElementFinder and TableElementFinder to support normal locator strategies

aaltat:


aaltat:
Things to do: 
- [x] Write better test in test_tableelementfinder.py test
- [x]  Make acceptance test to use newly supported locators like `id`, `name` and so on.

pekkaklarck:
Test also new `strategy:value` syntax.

aaltat:
The table element keyword are now done. I will fix the library documentation in separate PR.

aaltat:
This fixes #923

pekkaklarck:
So much better than it used to be! More functionality with less code (+281 additions, −293 removals) is always nice.

PR #926, opened 2017-09-26
Fixes documentation for locating tables

aaltat:
Before this one can be merged, the https://github.com/robotframework/SeleniumLibrary/pull/921 should be merged too.


pekkaklarck:
I'm not sure is the note that table locators used to work strangely worth a separate section. I'll merge this anyway because the current docs are wrong. Can still fine tune the docs later if needed.

PR #935, opened 2017-09-28
Fixes to alerts

### LINKED ISSUE #933: Clean up keywords related to alerts
The current alert related keywords are a mess. Biggest problems:

1. Selecting should the alert be dismissed by pressing `Ok` or `Cancel` needs to be done using separate `Choose Ok/Cancel On Next Confirmation` keywords. I think there used to be a technical reason to do it this way, but looking at the current code this doesn't seem to be the case. Much better to just pass this information as a normal argument to keywords dismissing alerts.

2. There are several keywords with same/similar functionality. In practice `Confirm Action`, `Dismiss Alert` an `Get Alert Message` all close the alert by default.

3. Strange configuration options and default values. For example, `Dismiss Alert` actually accepts the alert by default and needs to be separately configured to dismiss it.

-------

If I've understood it correctly, there can be three kind of Javascript alerts:

1. An alert with a message and Ok button (`window.alert`).
2. An alert with a message and Ok/Cancel buttons (`window.confirm`)
3. An alert with a message, text input field, and Ok/Cancel buttons (`window.prompt`)

We need to think what kind of a set of keywords are needed to handle these different alerts. The more we can keep backwards compatibility (i.e. keyword names) the better, but the most important goal is making the set of keywords consistent and easy to use.

### LINKED ISSUE #934: Regressions in `Dismiss Alert` and `Confirm Action` compared to 1.8
At the moment (SL 3.0 beta 3) there are some regressions compared to version 1.8:

1. `Dismiss Alert    accept=${FALSE}` leaves the alert open but it in 1.8 it closes it regardless the value of `accept`. It's behavior to accept by default and require configuration to dismiss is strange, but we already have #933 about that and other alert related oddities.

2. `Dismiss Alert` always returns the message of the alert but in 1.8 it returns true or false depending was the alert close or not. This is strange behavior too.

3. `Confirm Action` accepts the first alert but subsequent alerts are dismissed by default. In 1.8 all alerts were accepted by default.

### LINKED ISSUE #942: Support configurable timeout with alert related keywords
In S2L 1.8 alert related keywords use hard coded one second wait when looking for alerts. Currently in SL 3.0 beta 3 the timeout has been increased to 5 seconds but is still hard coded. As #917 already proposed, this timeout should be configurable.

We've made a decision that these keywords will use the same timeouts as `Wait ...` keywords by default, but they also have `timeout=None` argument to override.

We also have #933 about cleaning up alerts in general, and that involves deprecating several keywords. The support for custom timeouts should be only added to keywords that are going to be preserved.

pekkaklarck:
This PR is contains following changes to alerts and to some other (un)related functionality:

- Regression fix to Dismiss Alert and Confirm Action. Fixes #934.
- Small fix to Set Selenium Speed.
- New and better alert related keywords. Fixes #933.
- New `is_noney` utility and related small fix to wait keywords.
- Configurable timeouts for alerts. Fixes #942.
- Configure Travis to use `--dotted` when possible to ease seeing failed tests.

This PR is now ready for final review.

### DIFF
--- a/src/SeleniumLibrary/keywords/alert.py
+++ b/src/SeleniumLibrary/keywords/alert.py
@@ -78,8 +78,10 @@ class AlertKeywords(LibraryComponent):
     def alert_should_not_be_present(self, action=ACCEPT, timeout=0):
         """Verifies that no alert is present.
 
-        If the alert actually exists, ``action`` argument determines
-        how it should be handled same way as with `Handle Alert`.
+        If the alert actually exists, the ``action`` argument determines
+        how it should be handled. By default the alert is accepted, but
+        it can be also dismissed or left open the same way as with the
+        `Handle Alert` keyword.
 
         ``timeout`` specifies how long to wait for the alert to appear.
         By default the alert is not waited at all, but a custom time can
--- a/src/SeleniumLibrary/keywords/alert.py
+++ b/src/SeleniumLibrary/keywords/alert.py
@@ -19,7 +19,7 @@ from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.support.ui import WebDriverWait
 
 from SeleniumLibrary.base import keyword, LibraryComponent
-from SeleniumLibrary.utils import is_truthy
+from SeleniumLibrary.utils import is_truthy, secs_to_timestr, timestr_to_secs
 
 
 class AlertKeywords(LibraryComponent):
@@ -38,20 +38,23 @@ class AlertKeywords(LibraryComponent):
         self.input_text_into_alert(text, self.LEAVE)
 
     @keyword
-    def input_text_into_alert(self, text, action=ACCEPT):
+    def input_text_into_alert(self, text, action=ACCEPT, timeout=None):
         """Types the given ``text`` into an input field in an alert.
 
         The alert is accepted by default, but that behavior can be controlled
         by using the ``action`` argument same way as with `Handle Alert`.
 
+        ``timeout`` specifies how long to wait for the alert to appear.
+        If it is not given, the global default `timeout` is used instead.
+
         New in SeleniumLibrary 3.0.
         """
-        alert = self._wait_alert()
+        alert = self._wait_alert(timeout)
         alert.send_keys(text)
         self._handle_alert(alert, action)
 
     @keyword
-    def alert_should_be_present(self, text='', action=ACCEPT):
+    def alert_should_be_present(self, text='', action=ACCEPT, timeout=None):
         """Verifies that an alert is present and, by default, accepts it.
 
         Fails if no alert is present. If ``text`` is a non-empty string,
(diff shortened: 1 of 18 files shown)


src/SeleniumLibrary/keywords/alert.py
```diff
@@ -14,153 +14,183 @@
...
-        keyword or another like `Get Alert Message`.
+        New in SeleniumLibrary 3.0.
         """
-        alert_text = self._handle_alert(self.ACCEPT_ALERT)
-        if text and alert_text != text:
-            raise AssertionError("Alert text should have been "
-                                 "'%s' but was '%s'"
-                                 % (text, alert_text))
+        try:
+            alert = self._wait_alert(timeout=0)
```
aaltat: Why the timeout is zero? It would make the tests more stable if we allow small polling.

src/SeleniumLibrary/keywords/alert.py
```diff
@@ -14,152 +14,183 @@
...
+        Fails if no alert is present. If ``text`` is a non-empty string,
+        then it is used to verify alert's message. The alert is accepted
+        by default, but that behavior can be controlled by using the
+        ``action`` argument same way as with `Handle Alert`.
+
+        The ``action`` argument is new in SeleniumLibrary 3.0. In earlier
+        versions the alert was always accepted.
+        """
+        message = self.handle_alert(action)
+        if text and text != message:
```
aaltat: Now the text is exact match, but I was thinking could `if text and text not in message:` be also useful. Just a random thought, without a real use case.

aaltat (commented):
I did read the code but didn't have time to read the test when commuting. Also would like to take a look from the bigger screen.

But generally looks OK.

src/SeleniumLibrary/keywords/alert.py
pekkaklarck: Yeah, if I had designed this I'd probably supported glob-style patterns (e.g. `foo *`) similarly as `Run Keyword And Expect Error` does. Don't think changing existing logic is a good idea, though. Could add `pattern=False`, but that's a topic for a separate issue. It's anyway already possible to do, for example:

    ${text} =    Alert Should Be Present
    Should Match    ${text}    foo *

src/SeleniumLibrary/keywords/alert.py
pekkaklarck: We expect that there's no alert so in the common case *any* polling would slow down execution unnecessarily. Very small timeout could be considered, but at least using the global timeout that's 5 seconds by default would be a very bad idea. I'd leave the default timeout here to be zero and just make the timeout configurable with an optional argument.

aaltat:
Again I did read the code but didn't have time to read the test when commuting. Will read rest of the PR later.

src/SeleniumLibrary/keywords/alert.py
aaltat: True. And alerts are pretty simple in any case.

Issue #757, opened 2017-02-08
Keyword to wait that windows is open


aaltat:
Many cases, like #756, the selecting of a window may fail because the windows is not yet open. Then a common solution is to do this: 
```
| Wait Until Keyword Succeeds | 5 | 1 | Select Window | window_title |
```
It would be better to create keyword that allow to wait a window is opened, example:
```
| Wait Until Window Is Open | window_title | timeout |
| Select Window | window_title |
```
Other option would be to enhance the `Select Window` to perform polling and add a timeout 
```
| Select Window | window_title | timeout |
```

pekkaklarck:
I initially thought we should add `Wait Until Window Is Open`, but having just one keyword for waiting and selecting the window actually sounds even better. Adding optional timeout to `Select Window` might thus be a good idea. I guess it should work so that if no timeout is given there's no polling either. Would similar timeout functionality be useful also with other keywords?

aaltat:
For me also wait and select window sounds more realistic use case.

Having similar functionality could be useful in other keywords too. At least I usually build my UI facing libraries in such way that they poll and assert. Also if time out is zero it only performs assert, but I rarely use that because polling is usually the best option to build stable test.

There is at least one down side, if one is writing test by using `Page Should *` keywords, then test are not that readable than `Wait Until *` keywords. I actually used, with my poll and assert keywords, a `Verify *` prefix. For me it made understanding what the keyword does easier and it made writing the test more fluent.

pekkaklarck:
I agree. I actually think that `Should` keywords should rather get `Wait` counterparts than have an optional timeout. Like we already have `Page Should Contain` and `Wait Until Page Contains`. Having a timeout would be more useful in keywords that do something like `Click Button    example    timeout=2s`.

PR #936, opened 2017-10-02
Enhanced Click Element keyword

ponkar:


ponkar:
@aaltat I cherry picked in the new branch from master and still see this issue. Can you please look into this and let me know the root cause

ponkar (commented):
@aaltat Please check these changes which I have made and am running PR against them and no idea where it failing

aaltat:
The history looks good and I found the reason what fails. Next time run all the acceptance test locally and you can see the error too. In practice, the reason for failure is here: 
https://github.com/robotframework/SeleniumLibrary/pull/936/files#diff-530ec8e37846ff25bdd55df7c643b806L429

Our acceptance test also verify what the keyword logs and that is done by using the [robotstatuschecker](https://github.com/robotframework/statuschecker). In your commit, you have removed the logging from the keyword and therefore one of the test fails. Restore the logging, then test should pass.

src/SeleniumLibrary/keywords/element.py
```diff
@@ -420,14 +420,40 @@ def get_vertical_position(self, locator):
...
+    def click_element(self, locator, modifier=None):
         """Click element identified by `locator`.
 
         Key attributes for arbitrary elements are `id` and `name`. See
         `introduction` for details about locating elements.
-        """
-        self.info("Clicking element '%s'." % locator)
-        self.find_element(locator).click()
+        
+        modifier(optional) : If Element need to be opened in new window
```
aaltat: 1) The keyword documentation needs some desining. 
1.1) Arguments are written using back ticks \`\`modifier\`\`
1.2) Then argument is described in few sentences. Each argument are in their own chapters.
1.3) Example are written using Robot Framework table syntax, example:
```
Example:
| Click Element | xpath://button |  SHIFT | # Holds shift when clicking the element | 
```

2) Also we are picky about the white space and trailing white space is not allowed. usually one can configure the  editor to remove all trailing white space.

src/SeleniumLibrary/keywords/element.py
```diff
@@ -420,14 +420,40 @@ def get_vertical_position(self, locator):
...
+        
+        eg Click Element    <element_locator>    SHIFT              ----> Opens Element in new Window
+           Click Element    <element_locator>    SHIFT+CTRL         ----> Opens Element in new tab
+           CLick Element    <element_locator>                       ----> Opens Element in same window
+           
+        Note : Attempt to open element in new tab in firefox might not get window handles of all open windows, however
```
aaltat: I do not understand what you are trying to say. Could you explain it better.

src/SeleniumLibrary/keywords/element.py
```diff
@@ -420,14 +420,40 @@ def get_vertical_position(self, locator):
...
+        opening element in window in any browser can return window handles of all opened windows.
+        """
+
+        if(modifier!=None):
```
aaltat: Write: 
`if modifier:`

src/SeleniumLibrary/keywords/element.py
```diff
@@ -420,14 +420,40 @@ def get_vertical_position(self, locator):
...
+        modifier(optional) : If Element need to be opened in new window
+        
+        eg Click Element    <element_locator>    SHIFT              ----> Opens Element in new Window
+           Click Element    <element_locator>    SHIFT+CTRL         ----> Opens Element in new tab
+           CLick Element    <element_locator>                       ----> Opens Element in same window
+           
+        Note : Attempt to open element in new tab in firefox might not get window handles of all open windows, however
+        opening element in window in any browser can return window handles of all opened windows.
+        """
+
```
aaltat: Remove all empty lines in the method.

src/SeleniumLibrary/keywords/element.py
```diff
@@ -420,14 +420,40 @@ def get_vertical_position(self, locator):
...
+        if(modifier!=None):
+            element = self.find_element(locator)
+            action = ActionChains(self.browser)
+            
+            modifier = modifier.strip()
+            keys = [key.strip().encode() for key in modifier.lower().split('+') ]
+
+            if ("ctrl" or "control") in keys and "shift" in keys and '' not in keys:
+                self.info("Opening element in new tab..")
```
aaltat: Because what happens with modifier keys is browser specific implementation and also may change between different OS too, it would be better to just log the keys what are going to be logged. Fix also line 450.

src/SeleniumLibrary/keywords/element.py
```diff
@@ -420,14 +420,40 @@ def get_vertical_position(self, locator):
...
+        """
+
+        if(modifier!=None):
+            element = self.find_element(locator)
+            action = ActionChains(self.browser)
+            
+            modifier = modifier.strip()
+            keys = [key.strip().encode() for key in modifier.lower().split('+') ]
+
+            if ("ctrl" or "control") in keys and "shift" in keys and '' not in keys:
```
aaltat: Write unit test for keys and their combinations: SHIFT, CTRL, ALT, COMMAND. It might be easier to put the logic in separate method for unit test purposes.

test/acceptance/open_and_close.robot
```diff
@@ -63,3 +63,29 @@ Closing all browsers clears cache
     Run Keyword And Expect Error
     ...    No browser with index or alias 'Browser 2' found.
     ...    Switch Browser    Browser 2
+
+Click Element Without Modifier
+    Open Browser    ${ROOT}/jquery.html    ${BROWSER}
+    Set Selenium Implicit Wait    60s
+    Click Element    link=Link with id
+    Sleep    5s
```
aaltat: Adding plains sleep is not good idea and because #757 is not implemented please do something like this: 
``` 
| Wait Until Keyword Succeeds | 10 | 1 | Location Should Be  | ${ROOT}/broken.html |
```

Fix in all places where `Sleep` keyword is used.

test/acceptance/open_and_close.robot
```diff
@@ -63,3 +63,29 @@ Closing all browsers clears cache
...
+    Select Window    ${windows[1]}
+    Location Should Be    ${ROOT}/broken.html
+
+Click Element With Incorrect Keys
+    Open Browser    ${ROOT}/jquery.html    ${BROWSER}
+    Set Selenium Implicit Wait    60s
+    ${error_message}=    Set Variable    KeyError: "'SHIFT+' click is not allowed with this keyword"
+    Run Keyword And Expect Error    ${error_message}    Click Element    link=Link with id    SHIFT+
+    ${error_message}=    Set Variable    KeyError: "'SHIFT+ALT' click is not allowed with this keyword"
+    Run Keyword And Expect Error    ${error_message}    Click Element    link=Link with id    SHIFT+ALT
```
aaltat: I do not see reason why we should block combinations which may or may not work. I think this should actually work.

test/acceptance/open_and_close.robot
```diff
@@ -63,3 +63,29 @@ Closing all browsers clears cache
...
+    @{windows}=    List Windows
+    Log To Console    ${windows}       
+    Select Window    ${windows[1]}
+    Location Should Be    ${ROOT}/broken.html
+
+Click Element With Incorrect Keys
+    Open Browser    ${ROOT}/jquery.html    ${BROWSER}
+    Set Selenium Implicit Wait    60s
+    ${error_message}=    Set Variable    KeyError: "'SHIFT+' click is not allowed with this keyword"
+    Run Keyword And Expect Error    ${error_message}    Click Element    link=Link with id    SHIFT+
```
aaltat: Why this test should fail?

test/acceptance/open_and_close.robot
```diff
@@ -63,3 +63,29 @@ Closing all browsers clears cache
     Run Keyword And Expect Error
     ...    No browser with index or alias 'Browser 2' found.
     ...    Switch Browser    Browser 2
+
+Click Element Without Modifier
+    Open Browser    ${ROOT}/jquery.html    ${BROWSER}
+    Set Selenium Implicit Wait    60s
```
aaltat: Why the implicit wait is increased to 60 seconds?

aaltat (changes requested):

ponkar:
Was away for few days and will surely work on your coments

src/SeleniumLibrary/keywords/element.py
ponkar: I have removed that note as its not so vital

src/SeleniumLibrary/keywords/element.py
ponkar: done !!

src/SeleniumLibrary/keywords/element.py
ponkar: done !!

test/acceptance/open_and_close.robot
ponkar: I had kept it with the objective if there is a delay to load elements for the site to be loaded.
However removing it also works !!

src/SeleniumLibrary/keywords/element.py
ponkar: @aaltat : Please can you guide where to write unittest and any example if any ?

test/acceptance/open_and_close.robot
ponkar: Removed it

test/acceptance/open_and_close.robot
ponkar: While writing scripts sometimes if parameters go incomplete (SHIFT+)or incorrect(SHIFT+ALT) such operations should not be allowed.

test/acceptance/open_and_close.robot
ponkar: SHIFT+ALT + Click - > actually tries to save the link with Save window
SHIFT+ -> Is the incorrect way of specifying keys

So such things dont make sense as our keyword is more focused about opening link either in same window or new tab or new window

ponkar (commented):
@aaltat Please verify this commit

aaltat:
We did discuss in the Slack and I promised to take a look. But I want to get the 3.0 rc1 done first.

DirkRichter:
fyi: missing `self.info("Clicking element '%s'." % locator)` if modifier==None

aaltat:
I am closing this PR on favor of #1139

PR #940, opened 2017-10-02
Improved xpath locator documentation

andreabisello:


pekkaklarck:
This is a good example to add, but I'm not sure does it deserve its own table. What about changing the xpath example above the table you added to contain this xpath instead? Then you could add a comments to all examples in that table explaining how different locators work. There are similar comments in the examples in the Default locator strategy section.

pekkaklarck:
Also, we nowadays prefer the new `strategy:value` format with examples. With xpath you need to add an extra space like `xpath: //*[contains(text(), 'example')]` or otherwise Libdoc will render it as an URL.

aaltat:
I agree with Pekka

andreabisello:
ok, i will update asap

aaltat:
@abioneperhobby how is the update looking?

andreabisello:
@aaltat @pekkaklarck new commit pushed

andreabisello:
@aaltat @pekkaklarck there is a conflict. i don't know what i should to do. can i solve the conflict? thanks

pekkaklarck:
You could update and rebase your fork. If you don't know git too well, it may be easier to just create a new fork and new PR. I can also add such an example myself.

src/SeleniumLibrary/__init__.py
```diff
@@ -121,6 +121,10 @@ class SeleniumLibrary(DynamicCore):
     Notice that using the ``sizzle`` strategy or its alias ``jquery``
     requires that the system under test contains the jQuery library.
 
+    
+    This will Click Element containing ABC text. Notice that this approach can be error prone
```
aaltat, line 125: I don't think this is worth of mention. We are trying to explain how the library is working and not trying to explain how xpath works.

src/SeleniumLibrary/__init__.py
```diff
@@ -129,8 +133,10 @@ class SeleniumLibrary(DynamicCore):
 
     Examples:
 
-    | `Click Element` | //div[@id="container"] |
-    | `Click Element` | (//div)[2]             |
+    | `Click Element` | //div[@id="container"] | Click Div With id : container |
+    | `Click Element` | (//div)[2]             | Click the second div in the document |
+    | `Click Element` | xpath=//*[contains(text(),'ABC')] | Click element containint text ABC |
```
aaltat, line 138: Could you make the column same size by adding spaces.  The first two examples, column two and three needs some adjustments, because last example has longer column width.

aaltat (changes requested):

pekkaklarck:
@abioneperhobby, does 4724be0b21ae9560aaa6d7862c43e538c12d0e63 cover what needed? Notice that i's in a PR which hasn't yet been merged.

andreabisello:
@pekkaklarck i think yes. with that pull request my pull request can be dropped.

pekkaklarck:
Cool, closing this one then. The enhanced docs have also been merged.

PR #944, opened 2017-10-05
Wait For Condition checks that condition contains 'return'.

### LINKED ISSUE #943: `Wait For Condition` should validate that condition contains `return`
`Wait For Condition` requires the condition to return the value to be evaluated. For example, I just used

    Wait For Condition    document.title == 'Example'

and it didn't work. After reading the docs I changed it to this working variant

    Wait For Condition    return document.title == 'Example'

The keyword should check that there is `return` in the condition. This change can cause problems if the `return` isn't mandatory, but AFAIK it it.

pekkaklarck:
Fixes #943.


aaltat:
Darn, that alert test with FF fails. But they are fixed in separate pull request. Perhaps I will review that first and then this one.

PR #945, opened 2017-10-05
Fixes cookie object and Get Cookie keyword documentation

### LINKED ISSUE #932: Add keyword to get all cookie information
We currently have `Get Cookie Value` but it only returns the value, not other information such as expiry. We have few possibilities how this could be implemented:

1. Enhance the existing `Get Cookie Value` to allow returning other info too. For backwards compatibility reasons its default behavior cannot change, but it could get a new optional argument like `info='value'` that could be used like this:

        ${value} =    Get Cookie Value    name
        Log    ${value}
        ${expiry} =    Get Cookie Value    name    info=expiry
        Log    ${expiry}

2. Add separate new keyword like `Get Cookie` or `Get Cookie Information` that returns a custom object containing all cookie information in its attributes. This info could then be used in the data easily:

        ${cookie} =    Get Cookie    name
        Log    ${cookie.value}
        Log    ${cookie.expiry}

3. Enhance the existing `Get Cookie Value` to return an object with all the cookie information.

My favorite is option 2. If we decide to do that, then we could consider deprecating the existing `Get Cookie Value`. We should also think what to do with the current `Get Cookies` which returns a string in format `cookie1=value1; cookie2=value2`. It should rather return a list of cookie objects, but that change would be backwards incompatible. I have an idea how it could be deprecated pretty gracefully, though.

aaltat:
Fixes #932


src/SeleniumLibrary/keywords/cookie.py
```diff
@@ -44,9 +44,9 @@ def get_cookies(self):
 
     @keyword
     def get_cookie_value(self, name):
-        """Returns value of cookie found with `name`.
+        """Returns value of cookie found with ``name``.
 
-        If no cookie is found with `name`, this keyword fails.
+        If no cookie is found with ``name``, this keyword fails.
```
pekkaklarck: I guess we could silently deprecate this keyword in favor of `Get Cookie Information`.

src/SeleniumLibrary/keywords/cookie.py
```diff
@@ -55,24 +55,55 @@ def get_cookie_value(self, name):
...
     def get_cookie(self, name):
-        """Returns the cookie found with `name` in extended variable format.
-
-        If no cookie is found with `name`, this keyword fails.
+        """Returns a cookie object found with ``name``.
+
+        If no cookie is found with ``name``, keyword fails. The cookie object
+        contains details about the cookie. Attributes available in the object
+        are documented in the table below.
+        | = Attribute = |             = Explanation =                            |
```
pekkaklarck: Empty row before the table would be nice.

src/SeleniumLibrary/keywords/cookie.py
```diff
@@ -55,24 +55,55 @@ def get_cookie_value(self, name):
...
+        | domain        | Specifies hosts to which the cookie will be sent       |
+        | expiry        | The maximum lifetime of the cookie as EPOCH            |
```
pekkaklarck: This is a bit confusing. Is this a timeout like 1000 seconds or is it the actual time when the cookie expires? If it's a timeout, reference to EPOCH is wrong. If it's a time, then we should expose it to users as a datetime object.

Also, sentences should end with `.`.

src/SeleniumLibrary/keywords/cookie.py
```diff
@@ -55,24 +55,55 @@ def get_cookie_value(self, name):
...
         """
         cookie = self.browser.get_cookie(name)
         if cookie:
-            cookie_information = CookieInformation(cookie)
-            return cookie_information
+            return CookieInformation(
+                cookie.get('domain', None), cookie.get('expiry', None),
+                cookie.get('httpOnly', None), cookie['name'],
+                cookie.get('path', None), cookie.get('secure', None),
+                cookie['value'])
```
pekkaklarck: No need to have e.g. `cookie.get('doman', None)` when `cookie.get('domain)` works exactly the same way. But do we even need that or could we just use `cookie['name']`? I don't see the spec saying anything that some of thee values would be optional: https://w3c.github.io/webdriver/webdriver-spec.html#cookies

If we can trust that the return value from `get_cookie` always have exactly the same content, we could actually use `CookieInformation(**cookie)`. That is a bit risky as it would break if any new information is added to the return value.

src/SeleniumLibrary/keywords/cookie.py
```diff
@@ -55,24 +55,55 @@ def get_cookie_value(self, name):
...
         """Adds a cookie to your current session.
 
-        "name" and "value" are required, "path", "domain", "secure" and
-         "expiry" are optional.  Expiry supports the same formats as
-         the DateTime library.
+        ``name`` and ``value`` are required, ``path``, ``domain``, ``secure``
+        and ``expiry`` are optional.  Expiry supports the same formats as
+        the [http://robotframework.org/robotframework/latest/libraries/DateTime.html|DateTime]
+        library and is converted to EPOCH timestamp which is supported by the
+        Selenium.
```
pekkaklarck: The note how the value is internally turned into an EPOCH timestamp isn't too relevant.

src/SeleniumLibrary/keywords/cookie.py
```diff
@@ -55,24 +55,55 @@ def get_cookie_value(self, name):
...
+        | domain        | Specifies hosts to which the cookie will be sent       |
+        | expiry        | The maximum lifetime of the cookie as EPOCH            |
+        | httpOnly      | HttpOnly cookie cannot be accessed by client-side APIs |
+        | name          | The name of a cookie                                   |
+        | path          | Indicates a URL path, example /                        |
+        | secure        | Cookie will be send only by using secure connection    |
+        | value         | Value of the cookie                                    |
+        | full_info     | All the above attributes joined as string              |
+
+        New in SeleniumLibrary 3.0
```
pekkaklarck: Note about when the keyword was added could be at the end of the doc and should have `.` at the end.

src/SeleniumLibrary/keywords/cookie.py
```diff
@@ -55,24 +55,55 @@ def get_cookie_value(self, name):
...
+        are documented in the table below.
+        | = Attribute = |             = Explanation =                            |
+        | domain        | Specifies hosts to which the cookie will be sent       |
+        | expiry        | The maximum lifetime of the cookie as EPOCH            |
+        | httpOnly      | HttpOnly cookie cannot be accessed by client-side APIs |
+        | name          | The name of a cookie                                   |
+        | path          | Indicates a URL path, example /                        |
+        | secure        | Cookie will be send only by using secure connection    |
+        | value         | Value of the cookie                                    |
+        | full_info     | All the above attributes joined as string              |
```
pekkaklarck: `__str__` seems to have all the information already. I don't think we need `full_info` at all.

src/SeleniumLibrary/keywords/cookie.py
```diff
@@ -89,20 +120,19 @@ def add_cookie(self, name, value, path=None, domain=None, secure=None,
...
 class CookieInformation(object):
-    def __init__(self, cookie):
-        self.domain = cookie['domain'] if 'domain' in cookie else None
-        self.expiry = cookie['expiry'] if 'expiry' in cookie else None
-        self.httpOnly = cookie['httpOnly'] if 'httpOnly' in cookie else None
-        self.name = cookie['name'] if 'name' in cookie else None
-        self.path = cookie['path'] if 'path' in cookie else None
-        self.secure = cookie['secure'] if 'secure' in cookie else None
-        self.value = cookie['value'] if 'value' in cookie else None
+    def __init__(self, domain, expiry, httpOnly, name, path, secure, value):
```
pekkaklarck: Isn't there any better order for these arguments than alphabetical? I'd have name and value first, and others could be in same order as here in the spec: https://w3c.github.io/webdriver/webdriver-spec.html#cookies

Should use same order also in the table listing the attributes.

src/SeleniumLibrary/keywords/cookie.py
```diff
@@ -89,20 +120,19 @@ def add_cookie(self, name, value, path=None, domain=None, secure=None,
...
+        self.secure = secure
+        self.value = value
 
     @property
     def full_info(self):
-        cookie_info = ', '.join("{}={}".format(key, value) for
-                                (key, value) in self.__dict__.items())
-        return cookie_info
+        return ', '.join("{}={}".format(key, value) for
+                         (key, value) in self.__dict__.items())
```
pekkaklarck: 1. As I already commented above, this property isn't too useful as `__str__` already shows all information.

2. Should have values in a predefined order, preferably in same order as passed to `__init__`. This ensures important information is first and also makes testing easier.

3. I'd also consider having all values on separate lines (i.e. joining with `\n` instead of `, `) to make easier to read the result.

src/SeleniumLibrary/keywords/cookie.py
```diff
@@ -55,24 +55,55 @@ def get_cookie_value(self, name):
...
-        If no cookie is found with `name`, this keyword fails.
+        """Returns a cookie object found with ``name``.
+
+        If no cookie is found with ``name``, keyword fails. The cookie object
+        contains details about the cookie. Attributes available in the object
+        are documented in the table below.
+        | = Attribute = |             = Explanation =                            |
+        | domain        | Specifies hosts to which the cookie will be sent       |
+        | expiry        | The maximum lifetime of the cookie as EPOCH            |
+        | httpOnly      | HttpOnly cookie cannot be accessed by client-side APIs |
```
pekkaklarck: This is a boolean, right? Types should be mentioned at least when they aren't strings.

pekkaklarck (changes requested):
Looks good in general but there are still few things I'd fix. Also `Add Cookie` implementation seems to have some bugs, but they aren't aren't really part of this PR.

src/SeleniumLibrary/keywords/cookie.py
aaltat: Good idea, will do.

src/SeleniumLibrary/keywords/cookie.py
aaltat: True, will do

src/SeleniumLibrary/keywords/cookie.py
aaltat: It's time when cookie expires and not a timeout. Epoch is returned directly by the Selenium and therefore we just blindly return that and returning something more human readable is good idea. When you are referring to the datetime object are you actually referring to the Python DateTime object? Or should we decide to return some string which is in more human readable format?

src/SeleniumLibrary/keywords/cookie.py
aaltat: Yes it's a boolean. Good point about the documentation.

src/SeleniumLibrary/keywords/cookie.py
aaltat: I was thinking about that. Having `name` and `value` did feel too little information. But again having all feels too much.

I think most of the people are only interested about the `name` and `value` and not about the other details. Perhaps it would be good idea to revert this back to `name` and `value`.

src/SeleniumLibrary/keywords/cookie.py
aaltat: That was the original implementation but atleast in Windows all the keys are not always returned. Then it could lead to situations where the advanced variables are not always populated when running the same test by using different browser and/or different OS. 

I agree with the `cookie.get` comment.

src/SeleniumLibrary/keywords/cookie.py
pekkaklarck: I'd return Python's [datetime](https://docs.python.org/2/library/datetime.html#datetime.datetime). It has several benefits:

1. Trivial to construct from an EPOCH time with `datetime.fromtimestamp` factory method.
2. Has useful attributes that can be accessed like `${cookie.expiry.day}`.
3. Is supported as an input format by Robot's DateTime library.

src/SeleniumLibrary/keywords/cookie.py
pekkaklarck: I'd rather remote `full_info` altogether and have all info in `__str__` in format like this:

    name=<name>
    value=<value>
    path=<path>
    ...

src/SeleniumLibrary/keywords/cookie.py
aaltat: Done

src/SeleniumLibrary/keywords/cookie.py
aaltat: Done

src/SeleniumLibrary/keywords/cookie.py
aaltat: Done

src/SeleniumLibrary/keywords/cookie.py
aaltat: Done

src/SeleniumLibrary/keywords/cookie.py
aaltat: Done

src/SeleniumLibrary/keywords/cookie.py
aaltat: Done

PR #950, opened 2017-10-07
Depreacted Get Matching Xpath Count and Xpath Should Match X Times kw

### LINKED ISSUE #949: Enhancements to getting and validating element counts
- Deprecate `Locator Should Match X Times`, `Xpath Should Match X Times` and `Get Matching Xpath Count`. 
- Instead `Locator Should Match X Times` and `Xpath Should Match X Times` users should use: `Page Should Contain Element` keyword with `limit` argument. 
- Instead of `Get Matching Xpath Count` create `Get Matching Locator Count`.

aaltat:
There was already Locator Should Match X Times which should be used
instead of the Xpath Should Match X Times keyword. Created
Get Matching Locator Count keyword and therefore deprecated
Get Matching Xpath Count keyword.

Fixes #949

### DIFF
--- a/src/SeleniumLibrary/keywords/element.py
+++ b/src/SeleniumLibrary/keywords/element.py
@@ -171,12 +171,8 @@ class ElementKeywords(LibraryComponent):
         `Locator Should Match X Times`.
 
         Examples assumes that locator matches to three elements
-        | ${count} =                 | Get Matching Locator Count | xpath: //*[@name="div_name"] |
-        | Should Be Equal As Numbers | ${count}                   | 3                            |
-        | ${count} =                 | Get Matching Locator Count | name:div_name                |
-        | Should Be Equal As Numbers | ${count}                   | 3                            |
-        | ${count} =                 | Get Matching Locator Count | div_name                     |
-        | Should Be Equal As Numbers | ${count}                   | 3                            |
+        | ${count} =      | Get Matching Locator Count | name:div_name  |
+        | Should Be True  | ${count} > 2               |                |
 
         New in SeleniumLibrary 3.0
         """
@@ -193,9 +189,7 @@ class ElementKeywords(LibraryComponent):
         `loglevel` arguments.
 
         Examples assumes that locator matches to three elements
-        | Locator Should Match X Times | xpath: //*[@name="div_name"] | 3 |
-        | locator_should_match_x_times | name:div_name                | 3 |
-        | locator_should_match_x_times | div_name                     | 3 |
+        | locator_should_match_x_times | name:div_name | 3 |
         """
         actual_locator_count = len(self.find_element(
             locator, first_only=False, required=False)
@@ -758,25 +752,16 @@ return !element.dispatchEvent(evt);
     @keyword
     def get_matching_xpath_count(self, xpath, return_str=True):
         """Deprecated. Use `Get Matching Locator Count` instead."""
-        count = len(self.find_element("xpath=" + xpath, first_only=False,
-                                      required=False))
+        count = self.get_matching_locator_count("xpath=" + xpath)
         return str(count) if is_truthy(return_str) else count
 
     @keyword
     def xpath_should_match_x_times(self, xpath, expected_xpath_count,
                                    message='', loglevel='INFO'):
         """Deprecated. Use `Locator Should Match X Times` instead."""
-        actual_xpath_count = len(self.find_element(
-            "xpath=" + xpath, first_only=False, required=False))
-        if int(actual_xpath_count) != int(expected_xpath_count):
-            if is_falsy(message):
-                message = ("Xpath %s should have matched %s times but "
-                           "matched %s times"
-                           % (xpath, expected_xpath_count, actual_xpath_count))
-            self.ctx.log_source(loglevel)
-            raise AssertionError(message)
-        self.info("Current page contains %s elements matching '%s'."
-                  % (actual_xpath_count, xpath))
+        self.locator_should_match_x_times("xpath=" + xpath,
+                                          expected_xpath_count, message,
+                                          loglevel)
 
     @keyword
     def add_location_strategy(self, strategy_name, strategy_keyword, persist=False):
--- a/src/SeleniumLibrary/keywords/element.py
+++ b/src/SeleniumLibrary/keywords/element.py
@@ -171,12 +171,12 @@ class ElementKeywords(LibraryComponent):
         `Locator Should Match X Times`.
 
         Examples assumes that locator matches to three elements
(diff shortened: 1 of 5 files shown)


src/SeleniumLibrary/keywords/element.py
```diff
@@ -164,21 +164,47 @@ def page_should_contain_element(self, locator, message='', loglevel='INFO'):
...
+        If you wish to assert the number of matching elements, use
+        `Locator Should Match X Times`.
+
+        Examples assumes that locator matches to three elements
+        | ${count} =                 | Get Matching Locator Count | xpath: //*[@name="div_name"] |
+        | Should Be Equal As Numbers | ${count}                   | 3                            |
+        | ${count} =                 | Get Matching Locator Count | name:div_name                |
+        | Should Be Equal As Numbers | ${count}                   | 3                            |
+        | ${count} =                 | Get Matching Locator Count | div_name                     |
+        | Should Be Equal As Numbers | ${count}                   | 3                            |
```
pekkaklarck: Do we need all these variations? This looks more like an example about using different locators. A reference to the locator section in the introduction ought to be enough.

Also, it's a bit funny to validate the count when the note above says that `Locator Should Match X Times` should be used for that instead. Perhaps this example could have something like

    Should Be True    ${count} > 3

and word "exact" be added to the above note like "verify the exact number of matching elements".

Finally, we should use proper English with `.` at the end of sentences. In this doc it's missing from the first and last sentence.

src/SeleniumLibrary/keywords/element.py
```diff
@@ -164,21 +164,47 @@ def page_should_contain_element(self, locator, message='', loglevel='INFO'):
         self.assert_page_contains(locator, message=message, loglevel=loglevel)
 
     @keyword
+    def get_matching_locator_count(self, locator):
```
pekkaklarck, line 167: Should have `return_str` argument similarly as `get_matching_xpath_count` that this keyword deprecates.

src/SeleniumLibrary/keywords/element.py
```diff
@@ -731,39 +757,15 @@ def page_should_not_contain_image(self, locator, message='', loglevel='INFO'):
...
-        Incorrect:
-        | count = | Get Matching Xpath Count | xpath=//div[@id='sales-pop']
-
-        If you wish to assert the number of matching elements, use
-        `Xpath Should Match X Times`.
-        """
+        """Deprecated. Use `Get Matching Locator Count` instead."""
         count = len(self.find_element("xpath=" + xpath, first_only=False,
                                       required=False))
         return str(count) if is_truthy(return_str) else count
```
pekkaklarck: Couldn't the implementation be changes to just `return self.get_matching_locator_count('xpath:' + xpath)`? Same with the other deprecated keyword below.

pekkaklarck (changes requested):

src/SeleniumLibrary/keywords/element.py
aaltat: I agree that for backwards compatible reasons it should be there. But I don't like that argument, because `get_matching_xpath_count` by default returns a string and it would be better keyword functionality if this the keyword would return a integer. If string is really needed, one can use `Convert To String` keyword.

src/SeleniumLibrary/keywords/element.py
aaltat: The purpose of the example was to demonstrate that it's possible to get count by using different locators, but I agree that it was not good idea. It will only confuse users and is a maintenance burden.

src/SeleniumLibrary/keywords/element.py
aaltat: Good idea, will do

src/SeleniumLibrary/keywords/element.py
pekkaklarck: Oh, I thought the default was int and there was some strange reason to sometimes return str! I totally agree that int is a better return type and that no configuration is needed. Just need to document the change. At least the old keyword should have a note that the new one has different return value type.

src/SeleniumLibrary/keywords/element.py
```diff
@@ -164,21 +164,41 @@ def page_should_contain_element(self, locator, message='', loglevel='INFO'):
...
-        """Verifies that the page contains the given number of elements located by the given `locator`.
+        """Verifies that the page contains the given number of elements located by the given ``locator``.
 
         See `introduction` for details about locating elements.
 
         See `Page Should Contain Element` for explanation about `message` and
         `loglevel` arguments.
+
+        Examples assumes that locator matches to three elements
+        | locator_should_match_x_times | name:div_name | 3 |
```
pekkaklarck: This keyword is so simple that I don't know is an example really needed. If it's kept, the keyword name should be fixed and having just `Example:` before ought to be enough.

src/SeleniumLibrary/keywords/element.py
```diff
@@ -164,21 +164,41 @@ def page_should_contain_element(self, locator, message='', loglevel='INFO'):
...
+
+        If you wish to assert the number of matching elements, use
+        `Locator Should Match X Times`. Keyword will always return an integer.
+
+        Examples assumes that locator matches to three elements
+        | ${count} =      | Get Matching Locator Count | name:div_name  |
+        | Should Be True  | ${count} > 2               |                |
+
+        New in SeleniumLibrary 3.0
+        """
```
pekkaklarck: - There's still `.` missing at the end of two sentences.
- The comment before the example isn't that useful anymore. Having just `Example:` would be enough.

pekkaklarck (approved):
Looks good to me now. Added two few more comments about docs but they aren't that important. I'll go through all the docs still before the final release and can fine tune them myself too.

src/SeleniumLibrary/keywords/element.py
aaltat: Done

src/SeleniumLibrary/keywords/element.py
aaltat: Done

pekkaklarck:
Should we actually call these kws `Get Element Count` and ` Page Should Contain X Elements` instead? The latter wouldn't even be needed if we added `count=None` to `Page Should Contain Element` and interpreted it so that the default means that the page should contain one or more element but, for example, `count=3` would mean exactly three elements. Could even support something like `count={1,3}` or `count=1+` to mean a range.

aaltat:
As discussed in the issue, this needs total rewrite. Therefore keeping the history is not needed and I will create a new PR to fix the issue.

PR #955, opened 2017-10-12
Window manager cleanup

pekkaklarck:
New attempt to create a PR. Hopefully no conflicts this time.

### DIFF
--- a/test/unit/locators/test_windowmanager.py
+++ b/test/unit/locators/test_windowmanager.py
@@ -1,312 +1,253 @@
 import unittest
 import uuid
 
-from mockito import mock, unstub
+from mockito import mock
 
-from SeleniumLibrary.locators.windowmanager import WindowManager
+from SeleniumLibrary.locators import WindowManager
 
 
 class WindowManagerTests(unittest.TestCase):
 
     def test_select_with_invalid_prefix(self):
-        manager = WindowManager()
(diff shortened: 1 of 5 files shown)


test/unit/locators/test_windowmanager.py
```diff
@@ -1,312 +1,240 @@
...
-        browser = mock()
+        manager = WindowManagerWithMockBrowser()
         with self.assertRaises(ValueError) as err:
-            manager.select(browser, "something=test1")
+            manager.select("something=test1")
             self.assertEqual(
                 str(err),
                 "Window locator with prefix 'something' is not supported"
             )
-        unstub()
```
aaltat, line 20: It might be good idea to call unstub() in `def tearDown`.

aaltat (commented):
I had one comment about calling the `unstub`. But I need to read the the changes from bigger screen. Specially those unit test changes where hard to read from the mobile.

test/unit/locators/test_windowmanager.py
pekkaklarck: Is it needed when we don't actually stub anything here? We just mock `ctx`.

aaltat: I did read the mockito docs again and you are right is it needed when doing stubs.

PR #957, opened 2017-10-14
Rename browser to driver

aaltat:


aaltat:
To do:
- [ ] fix browsermanagement.py, because it says `browser` in many places where it should say `driver`
- [ ] deprecate method `self.browser`

aaltat:
Oh darn, deleted accidentally. Oh well will push again

PR #965, opened 2017-10-20
 Add 'find_element(s)' to public API

pekkaklarck:
See #882.

Also little cleanup to internal find_element(s) API.


src/SeleniumLibrary/__init__.py
```diff
@@ -389,6 +389,31 @@ def browser(self):
...
+        :param locator: Locator to use when searching the element.
+            See library documentation for the supported locator syntax.
+        :param parent: Optional parent ``WebElememt`` to search child elements
+            from. By default search starts from the root using ``WebDriver``.
+        :rtype: selenium.webdriver.remote.webelement.WebElement
+        :raises SeleniumLibrary.errors.ElementNotFound: If element not found.
+        """
+        return self._element_finder.find(locator, parent=parent)
+
+    def find_elements(self, locator, parent=None):
```
aaltat, line 407: Should this have similar sphinx style parameter definitions as find_element in above?

pekkaklarck: It wouldn't hurt. More important change would be adding return type information, but it should mention that return type is a list of `WebElement` objects, not just that the type is a list.

pekkaklarck: You had some other syntax changes to type info. Can you take care of all of them at the same time?

aaltat: Yes I can. I will merge this PR and write separate PR for fixing it.

PR #969, opened 2017-10-23
Deprecate xpath counts

aaltat:
Add Supports `Page Should Contain Element` to have range.


aaltat:
Todo: 
- [ ] Improve the error messages. 
- [ ] Read asserts acceptance test and see do they contains duplicates. 
- [ ] Improve Page Should Contains Element keyword documentation with examples.

aaltat:
This is not good enough. Making new one without the `x..y` support.

PR #972, opened 2017-10-24
Deprecate xpath counts

### LINKED ISSUE #949: Enhancements to getting and validating element counts

aaltat:
Fixes #949

### DIFF
--- a/src/SeleniumLibrary/keywords/element.py
+++ b/src/SeleniumLibrary/keywords/element.py
@@ -103,7 +103,8 @@ class ElementKeywords(LibraryComponent):
         self.info("Current page contains text '%s'." % text)
 
     @keyword
-    def page_should_contain_element(self, locator, message=None, loglevel='INFO'):
+    def page_should_contain_element(self, locator, message=None,
+                                    loglevel='INFO', limit=None):
         """Verifies that element ``locator`` is found on the current page.
 
         See the `Locating elements` section for details about the locator
@@ -112,10 +113,38 @@ class ElementKeywords(LibraryComponent):
         The ``message`` argument can be used to override the default error
         message.
 
+        The ``limit`` argument can used to define how many elements the
+        page should contain. When ``limit`` is `None`, case-insensitively,
+        page can contain one or more elements. When limit is a number,
+        page must contain same number of elements.
+
         See `Page Should Contain` for explanation about the ``loglevel``
         argument.
+
+        The ``limit`` argument is new in SeleniumLibrary 3.0
+
+        Examples assumes that locator matches to two elements.
+        | Page Should Contain Element | name: div_name | limit=1    | # Keyword fails.                  |
+        | Page Should Contain Element | name: div_name | limit=2    | # Keyword passes.                 |
+        | Page Should Contain Element | name: div_name | limit=none | # None is considered one or more. |
+        | Page Should Contain Element | name: div_name |            | # Same as above.                  |
         """
-        self.assert_page_contains(locator, message=message, loglevel=loglevel)
+        if is_noney(limit):
+            self.assert_page_contains(locator, message=message,
+                                      loglevel=loglevel)
+        else:
+            limit = int(limit)
+            count = len(self.find_elements(locator))
+            if count == limit:
+                self.info('Current page contains {} element(s).'
+                          .format(count))
+            else:
+                if is_noney(message):
+                    message = ('Page should have contained "{}" element(s), '
+                               'but it did contain "{}" element(s).'
+                               .format(limit, count))
+                self.ctx.log_source(loglevel)
+                raise AssertionError(message)
 
     @keyword
     def locator_should_match_x_times(self, locator, x, message=None, loglevel='INFO'):
--- a/src/SeleniumLibrary/keywords/element.py
+++ b/src/SeleniumLibrary/keywords/element.py
@@ -148,14 +148,7 @@ class ElementKeywords(LibraryComponent):
 
     @keyword
     def locator_should_match_x_times(self, locator, x, message=None, loglevel='INFO'):
-        """Verifies that ``locator`` matches ``x`` number of elements.
-
-        See the `Locating elements` section for details about the locator
-        syntax.
-
-        See `Page Should Contain Element` for explanation about ``message``
-        and ``loglevel`` arguments.
-        """
+        """Deprecated, use `Page Should Contain Element` instead."""
         count = len(self.find_elements(locator))
         x = int(x)
         if count != x:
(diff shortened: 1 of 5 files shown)


src/SeleniumLibrary/keywords/element.py
```diff
@@ -112,21 +113,42 @@ def page_should_contain_element(self, locator, message=None, loglevel='INFO'):
         The ``message`` argument can be used to override the default error
         message.
 
+        The ``limit`` argument can used to define how many elements the
+        page should contain. When ``limit`` is `None` page can contain one or
```
pekkaklarck: Should have double backticks around `None`. Might be a good idea to add `(default)` after `None` to make it explicit that this is the default behavior.

src/SeleniumLibrary/keywords/element.py
```diff
@@ -112,21 +113,42 @@ def page_should_contain_element(self, locator, message=None, loglevel='INFO'):
...
+        more elements. When limit is a number, page must contain same number
+        of elements.
+
         See `Page Should Contain` for explanation about the ``loglevel``
         argument.
+
+        The ``limit`` argument is new in SeleniumLibrary 3.0
```
pekkaklarck: `.` missing at the end. I'd also move this note at the very end of the doc after the examples.

src/SeleniumLibrary/keywords/element.py
```diff
@@ -112,21 +113,42 @@ def page_should_contain_element(self, locator, message=None, loglevel='INFO'):
...
+
+        Examples assumes that locator matches to two elements.
+        | Page Should Contain Element | name: div_name | limit=1    | # Keyword fails.                  |
+        | Page Should Contain Element | name: div_name | limit=2    | # Keyword passes.                 |
+        | Page Should Contain Element | name: div_name | limit=none | # None is considered one or more. |
+        | Page Should Contain Element | name: div_name |            | # Same as above.                  |
```
pekkaklarck: - Why is `name:` prefix needed? The keyword would search by name (and id) also without it.
- I've used single backticks around keyword names in examples to get linking to work. That's mainly useful when examples refer to other keywords, but using them consistently might be a good idea. What do you think?

src/SeleniumLibrary/keywords/element.py
```diff
@@ -112,21 +113,42 @@ def page_should_contain_element(self, locator, message=None, loglevel='INFO'):
...
         """
-        self.assert_page_contains(locator, message=message, loglevel=loglevel)
+        if is_noney(limit):
+            self.assert_page_contains(locator, message=message,
+                                      loglevel=loglevel)
+        else:
```
pekkaklarck: This nested if/else structure is pretty complex. I'd add `return` at the end of `if` here and remove `else` altogether. Or perhaps the whole `else` content should be in a separate helper to follow the [single level of abstraction principle](http://www.markhneedham.com/blog/2009/06/12/coding-single-level-of-abstraction-principle/).

src/SeleniumLibrary/keywords/element.py
```diff
@@ -112,21 +113,42 @@ def page_should_contain_element(self, locator, message=None, loglevel='INFO'):
...
     def locator_should_match_x_times(self, locator, x, message=None, loglevel='INFO'):
-        """Verifies that ``locator`` matches ``x`` number of elements.
-
-        See the `Locating elements` section for details about the locator
-        syntax.
-
-        See `Page Should Contain Element` for explanation about ``message``
-        and ``loglevel`` arguments.
-        """
+        """Deprecated, use `Page Should Contain Element` with `limit` argument instead."""
```
pekkaklarck: Should have double backticks around `limit` (and Libdoc should be enhanced to warn about using single backticks when not linking but that would be a backwards incompatible change).

src/SeleniumLibrary/keywords/element.py
```diff
@@ -693,25 +715,31 @@ def page_should_not_contain_image(self, locator, message=None, loglevel='INFO'):
...
-        """
-        count = len(self.find_elements('xpath:' + xpath))
+        """Deprecated. Use `Get Element Count` instead."""
+        count = self.get_element_count('xpath:' + xpath)
         return str(count) if is_truthy(return_str) else count
 
     @keyword
     def xpath_should_match_x_times(self, xpath, x, message=None, loglevel='INFO'):
-        """Deprecated. Use `Locator Should Match X Times` instead."""
+        """Deprecated, use `Page Should Contain Element` with `limit` argument instead."""
```
pekkaklarck: Double backticks around `limit` also here.

src/SeleniumLibrary/keywords/element.py
```diff
@@ -693,25 +715,31 @@ def page_should_not_contain_image(self, locator, message=None, loglevel='INFO'):
...
         self.locator_should_match_x_times('xpath:'+xpath, x, message, loglevel)
 
     @keyword
+    def get_element_count(self, locator):
+        """Returns number of elements matching ``locator``.
+
+        If you wish to assert the number of matching elements, use
+        `Page Should Contain Element` with `limit` argument. Keyword will
```
pekkaklarck: And double backticks here too.

src/SeleniumLibrary/keywords/element.py
```diff
@@ -693,25 +715,31 @@ def page_should_not_contain_image(self, locator, message=None, loglevel='INFO'):
...
+        always return an integer.
+        New in SeleniumLibrary 3.0.
```
pekkaklarck: Also this `New in ...` note would, in my opinion, work better at the very end of the kw doc after examples. Also Python has them there as this example illustrates: https://docs.python.org/3/library/argparse.html#allow-abbrev

pekkaklarck (approved):

src/SeleniumLibrary/keywords/element.py
aaltat: - True it is not needed, it just habit of being explicit and not implicit.  
- For constantly it would be good idea. Lets do it.

src/SeleniumLibrary/keywords/element.py
aaltat: Having the else is waste. I was also thinking creating a helper, but then decided against it.

aaltat (commented):
Fixes based on comments.

pekkaklarck:
Looked great to me. Possible further cleanup can be done separately.

PR #976, opened 2017-10-30
Rename browser to driver

aaltat:
To do:
- [x] fix browsermanagement.py, because it says `browser` in many places where it should say `driver`
- [x] deprecate method `self.browser`


aaltat:
But it definitely looks like my rebase with master did not go smoothly. Have to take a look later.

pekkaklarck:
I'd still use  `webdriver` instead of `driver` as the attribute name, but it seems everyone else likes the latter so let's use that. I'd still use use `WebDriverCache` and not `DriverCache`, though. More importantly, the type of the attribute (i.e. property) should be specified in the docstring. That makes the attribute type unambiguous for users and also helps IDEs.

aaltat:
The `windowmanager.py` needs some refactoring, so that it does not pass the `driver` as argument but uses the `self.driver` instead

src/SeleniumLibrary/__init__.py
```diff
@@ -343,7 +343,7 @@ def __init__(self, timeout=5.0, implicit_wait=0.0,
             WaitingKeywords(self),
             WindowKeywords(self)
         ]
-        self._browsers = BrowserCache()
+        self._webdrivers = WebDriverCache()
```
pekkaklarck: This should be `_drivers` to match ` driver`.

src/SeleniumLibrary/base/context.py
```diff
@@ -28,12 +28,12 @@ def __init__(self, ctx):
...
     @property
-    def browser(self):
-        return self.ctx.browser
+    def driver(self):
+        return self.ctx.driver
 
     @property
-    def browsers(self):
-        return self.ctx._browsers
+    def webdrivers(self):
```
pekkaklarck: This should be `drivers` to match ` driver`.

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -368,8 +367,8 @@ def set_selenium_speed(self, value):
         """
         old_speed = self.get_selenium_speed()
         self.ctx.speed = timestr_to_secs(value)
-        for browser in self.browsers.browsers:
-            self._monkey_patch_speed(browser)
+        for driver in self.webdrivers.drivers:
```
pekkaklarck: Isn't ConnectionCache iterable? If it is, this could be just `for driver in self.drivers`. If it isn't, it should be changed!

aaltat: It should be iterable, but changes the code to only iterate drivers which are not closed.

src/SeleniumLibrary/base/context.py
aaltat: Done

src/SeleniumLibrary/__init__.py
aaltat: Done

src/SeleniumLibrary/__init__.py
```diff
@@ -385,11 +385,22 @@ def failure_occurred(self):
             self._running_on_failure_keyword = False
 
     @property
+    def driver(self):
+        """Current active driver
```
pekkaklarck: `.` missing at the end.

src/SeleniumLibrary/__init__.py
```diff
@@ -385,11 +385,22 @@ def failure_occurred(self):
...
+
+        :rtype: selenium.webdriver.remote.webdriver.WebDriver
+        :raises RuntimeError if driver is not created
+        """
+        if not self._drivers.current:
+            raise RuntimeError('No driver is open')
```
pekkaklarck: Strange error message. Perhaps this should still be `No browser is open.`? Also, should we use a dedicated exception?

src/SeleniumLibrary/utils/webdrivercache.py
```diff
@@ -17,33 +17,33 @@
...
+    def drivers(self):
         return self._connections
 
-    def get_open_browsers(self):
-        open_browsers = []
-        for browser in self._connections:
-            if browser not in self._closed:
-                open_browsers.append(browser)
-        return open_browsers
+    def get_open_drivers(self):
```
pekkaklarck: Open driver sounds a bit odd. Should this perhaps be `get_active_browsers()`? Or perhaps property `active_browsers`?

pekkaklarck (approved):
Looks great. Added few line notes, though.

src/SeleniumLibrary/__init__.py
aaltat: Yes, browser is better.

It's strange exception, but that is how the original code worked. Dedicated exception would be better, would `NoBrowser` be good?

pekkaklarck: Perhaps NoOpenBrowser or NoBrowserOpen?

src/SeleniumLibrary/__init__.py
aaltat: Done

src/SeleniumLibrary/utils/webdrivercache.py
aaltat: Did go with `active_drivers`

src/SeleniumLibrary/utils/webdrivercache.py
```diff
@@ -17,33 +17,34 @@
 from robot.utils import ConnectionCache
 
 
-class BrowserCache(ConnectionCache):
+class WebDriverCache(ConnectionCache):
 
     def __init__(self):
-        ConnectionCache.__init__(self, no_current_msg='No current browser')
+        ConnectionCache.__init__(self, no_current_msg='No current driver')
```
pekkaklarck: This should still say `browser`. And have `.` at the end.

If that message is ever used, it's used by `ConnectionCache` as a message for `RuntimeError`. We have our own `NoOpenBrowser` now, but there's no good API to use that here. We could set `self._no_current` to an object using our exception (and it could extend `NoConnection`) but I'm not sure is that worth the effort.

src/SeleniumLibrary/__init__.py
```diff
@@ -384,11 +385,22 @@ def failure_occurred(self):
...
 
     @property
+    def driver(self):
+        """Current active driver.
+
+        :rtype: selenium.webdriver.remote.webdriver.WebDriver
+        :raises RuntimeError if driver is not created
+        """
+        if not self._drivers.current:
+            raise NoOpenBrowser('No browser is open')
```
pekkaklarck: - Exception type not updated in the doc string above.
- This error message could have `.` at the end.

src/SeleniumLibrary/utils/webdrivercache.py
aaltat: I did change to driver -> browser. 

If the start seeing the `RuntimeError`in real life, lets then decide what we should do.

src/SeleniumLibrary/__init__.py
aaltat: Done.

src/SeleniumLibrary/__init__.py
```diff
@@ -384,11 +385,22 @@ def failure_occurred(self):
             self._running_on_failure_keyword = False
 
     @property
+    def driver(self):
+        """Current active driver.
+
+        :rtype: selenium.webdriver.remote.webdriver.WebDriver
+        :raises SeleniumLibrary.errors.ElementNotFound: If element not found.
```
pekkaklarck: Thought it would raise NoOpenBrowser....

PR #981, opened 2017-11-01
New keyword Get Browser Capabilities.

### LINKED ISSUE #980: New keyword "Get Browser Capabilities".
This keyword was initially named "Get Browser Info" developed in PR #482 .
This keyword will be the base for the keyword "Get Browser Version" as requested on #569 .

The keyword by design returns a dictionary of Capabilities. This dictionary is DotDict, meaning that capabilities can be addressed either by index or by attributes. For example:
```
| ${capabilities}= | Get Browser Capabilities                            |                           |
| Log              | ${capabilities.browserName} ${capabilities.version} | # Using attributes        |
| Log              | Capabilities = ${capabilities}                      |                           |
| Log              | ${capabilities['platformName']}                     | # Using dictionary        |
| Log              | ${capabilities.get('may_not_exist', 'default')}     | # Gets 'default'          |
| Log              | ${capabilities['may_not_exist']}                    | # Produces KeyError       |
| Log              | ${capabilities.may_not_exist}                       | # Produces AttributeError |

```

HelioGuilherme66:
Returns a dictionary with keys accessible as attributes of the current browser capabilities.

Includes examples with logging of Common Browsers capabilities.

This PR overrides PR #482 and fixes #980.

### DIFF
--- a/src/SeleniumLibrary/keywords/browsermanagement.py
+++ b/src/SeleniumLibrary/keywords/browsermanagement.py
@@ -23,8 +23,9 @@ from selenium import webdriver
 
 from SeleniumLibrary.base import keyword, LibraryComponent
 from SeleniumLibrary.locators import WindowManager
-from SeleniumLibrary.utils import (is_falsy, is_truthy, secs_to_timestr,
-                                   timestr_to_secs, SELENIUM_VERSION)
+from SeleniumLibrary.utils import (DotDict, is_falsy, is_truthy,
+                                   secs_to_timestr, timestr_to_secs,
+                                   SELENIUM_VERSION)
 
 
 ROOT_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), ".."))
@@ -236,6 +237,36 @@ class BrowserManagementKeywords(LibraryComponent):
                                % index_or_alias)
         self.debug('Switched to browser with Selenium session id %s.'
                    % self.driver.session_id)
+
+    @keyword
+    def get_browser_capabilities(self):
+        """Gets current browser capabilities as dictionary.
+
+        Returns a Dictionary with webdriver (browser) desired capabilities attributes.
+        You may also query keys providing a default value when those keys are nonexistent.
+
+        Examples of returned attributes are (in *bold* the common ones):
+        'rotatable','takesScreenshot', 'acceptSslCerts', 'cssSelectorsEnabled', 'javascriptEnabled',
+        'databaseEnabled', 'locationContextEnabled', '*platform*', '*browserName*', '*version*',
+        'nativeEvents', 'applicationCacheEnabled', 'webStorageEnabled', 'handlesAlerts'
+        
+        Note: Firefox/Geckodriver is returning different names for '*platform*' and '*version*'.
+              It returns: '*platformName*' and '*browserVersion*'.
+        
+        Examples:
+        | ${capabilities}= | Get Browser Capabilities                            |                           |
+        | Log              | ${capabilities.browserName} ${capabilities.version} | # Using attributes        |
+        | Log              | Capabilities = ${capabilities}                      |                           |
+        | Log              | ${capabilities['platform']}                         | # Using dictionary        |
+        | Log              | ${capabilities.get('does_not_exist', 'default')}    | # Gets 'default'          |
+        | Log              | ${capabilities['does_not_exist']}                   | # Produces KeyError       |
+        | Log              | ${capabilities.does_not_exist}                      | # Produces AttributeError |
+
+        Reference, see [https://seleniumhq.github.io/selenium/docs/api/py/webdriver/selenium.webdriver.common.desired_capabilities.html|Common Desired Capabilities].
+
+        New in SeleniumLibrary 3.0.
+        """
+        return DotDict(self.driver.desired_capabilities)
 
     @keyword
     def get_source(self):
--- a/src/SeleniumLibrary/utils/attrdict.py
+++ b/src/SeleniumLibrary/utils/attrdict.py
@@ -1,0 +1,16 @@
+class AttrDict(dict):
+
+    def __getattr__(self, name):
+        try:
+            return self[name]
+        except KeyError:
+            raise AttributeError(name)
+
+    def __setattr__(self, name, value):
+        self[name] = value
+
+    def __delattr__(self, name):
+        try:
+            del self[name]
+        except KeyError:
+            raise AttributeError(name)
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,7 +47,7 @@ matrix:
         - BROWSER=chrome
         - SELENIUM=3.7.0
         - ROBOTFRAMEWORK=2.8.7
-        - ROBOT_OPTIONS=
+        - ROBOT_OPTIONS="--exclude not_on_rf_2.8.7"
     - python: "3.3"
       env:
         - BROWSER=chrome
--- a/src/SeleniumLibrary/utils/__init__.py
+++ b/src/SeleniumLibrary/utils/__init__.py
@@ -22,6 +22,10 @@ from .seleniumversion import SELENIUM_VERSION
 from .types import is_falsy, is_noney, is_string, is_truthy
 from .webdrivercache import WebDriverCache
 
+try:
+    from robot.utils import DotDict
+except ImportError:  # When using robotframework < 2.9
+    from .attrdict import AttrDict as DotDict
 
 def escape_xpath_value(value):
     if '"' in value and '\'' in value:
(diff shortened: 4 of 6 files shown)


src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -238,6 +239,39 @@ def switch_browser(self, index_or_alias):
                    % self.browser.session_id)
 
     @keyword
+    def get_browser_capabilities(self):
+        """Gets current browser capabilities from webdriver.
```
aaltat: Write: Gets current browser capabilities as dictionary

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -238,6 +239,39 @@ def switch_browser(self, index_or_alias):
...
+
+        Returns a Dictionary with webdriver (browser) desired capabilities attributes.
+        You may also query keys providing a default value when those keys are nonexistent.
```
aaltat: Instead of explaining how RF dictionary works, it might be better idea to link in RF user guide: http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#dictionary-variables and to the chapter which explains DotDic usage.

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -238,6 +239,39 @@ def switch_browser(self, index_or_alias):
...
+
+        Examples of returned attributes are (in *bold* the common ones):
```
aaltat, line 248: Does the browser drivers, like Chrome driver, has documentation what they return. If there is, it would be better to link in those documents.

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -238,6 +239,39 @@ def switch_browser(self, index_or_alias):
...
+        'rotatable','takesScreenshot', 'acceptSslCerts', 'cssSelectorsEnabled', 'javascriptEnabled',
+        'databaseEnabled', 'locationContextEnabled', '*platform*', '*browserName*', '*version*',
+        'nativeEvents', 'applicationCacheEnabled', 'webStorageEnabled', 'handlesAlerts'
+        
+        Note: Firefox/Geckodriver is returning different names for '*platform*' and '*version*'.
+              It returns: '*platformName*' and '*browserVersion*'.
+        
+        Examples:
```
aaltat, line 256: The example explains how RF dictionary works, which is best to leave to RF documentation. Perhaps providing few common attributes, like browserName would be enough.

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -238,6 +239,39 @@ def switch_browser(self, index_or_alias):
...
+        | Log              | ${capabilities.get('does_not_exist', 'default')}    | # Gets 'default'          |
+        | Log              | ${capabilities['does_not_exist']}                   | # Produces KeyError       |
+        | Log              | ${capabilities.does_not_exist}                      | # Produces AttributeError |
+
+        Reference, see [https://seleniumhq.github.io/selenium/docs/api/py/webdriver/selenium.webdriver.common.desired_capabilities.html|Common Desired Capabilities].
+
+        New in SeleniumLibrary 3.0.
+        """
+
+        capabilities = self.browser.desired_capabilities
```
aaltat: Write return DotDic (self.browser.desired_capabilities)

aaltat: Also remove not required new lines.

src/SeleniumLibrary/utils/attrdict.py
```diff
@@ -0,0 +1,16 @@
+class AttrDict(dict):
```
aaltat, line 1: Should we rename this in same way as RF DotDic.

aaltat: Although code is simple write few unit test might be good idea.

test/acceptance/create_webdriver.robot
```diff
@@ -31,6 +31,39 @@ Create Webdriver With Bad Keyword Argument Dictionary
     Run Keyword And Expect Error    kwargs must be a dictionary.
```
aaltat: Having own test suite is better idea.

test/acceptance/create_webdriver.robot
```diff
@@ -31,6 +31,39 @@ Create Webdriver With Bad Keyword Argument Dictionary
...
     ...    Create Webdriver    Firefox    kwargs={'spam': 'eggs'}
 
+Get Browser Capabilities Using Dictionary
```
aaltat: Could you change the test that one test tests only a one thing. Also get rid of the stuff which is not anymore valid like the [Setup].

test/acceptance/create_webdriver.robot
```diff
@@ -31,6 +31,39 @@ Create Webdriver With Bad Keyword Argument Dictionary
...
+    [Documentation]    Gets current browser capabilities from WebDriver
```
aaltat: If the doc repeats the test name, then it's not useful and must be removed.

test/acceptance/create_webdriver.robot
```diff
@@ -31,6 +31,39 @@ Create Webdriver With Bad Keyword Argument Dictionary
...
+    [Setup]    Run Keywords    Cannot Run In RF Pre 2.9    Set Driver Variables
+    Create Webdriver    ${DRIVER_NAME}
+    ${capabilities}=    Get Browser Capabilities
+    Log    Capabilities = ${capabilities}
+    ${browser}=    Normalize Browser Name
+    Should Match    ${capabilities['browserName']}    ${browser}
```
aaltat: Change tests that it checks keys and does not care about the value. Because value changes between browser and perhaps Selenium version. 

Test in the unit test level that correct value is in correct key.

aaltat (changes requested):
The code side looks OK but there are few things which needs to be fixed before this can be merged.

src/SeleniumLibrary/utils/attrdict.py
HelioGuilherme66: This, I don't agree. I only added AttrDict because of RF 2.8.7 compatibility. It is a temporary workaround.
The original PR had unit tests, but they will not be necessary because it is temporary.

aaltat: There already is some temporary code, waiting for the minimum RF version to raise high enough and they also have unit tests. So having something as temporally and not writing test for it, is not valid reason. Also it would lift the burden from the acceptance test, that one does not have to verify that value is in the right key in the acceptance test level.

pekkaklarck: Couldn't this just wait until we drop RF 2.8.7 support? I don't think this can make it to 3.0 anyway.

src/SeleniumLibrary/keywords/browsermanagement.py
HelioGuilherme66: Done

src/SeleniumLibrary/keywords/browsermanagement.py
HelioGuilherme66: I am keeping as it is. Better explicit than implicit. We don't have much examples on how to use SeleniumLibrary with dictionary attributes.

src/SeleniumLibrary/keywords/browsermanagement.py
HelioGuilherme66: There is the Webdriver standard mentioned in doc (and Firefox is not following it completely, for platform and version).

src/SeleniumLibrary/keywords/browsermanagement.py
HelioGuilherme66: Done

test/acceptance/create_webdriver.robot
HelioGuilherme66: Done

src/SeleniumLibrary/keywords/browsermanagement.py
aaltat: It would be better to link to the WebDriver documentation instead to trying to explain all in our documentation.

HelioGuilherme66:
I consider Done the changes to this PR.
Build is green again :)

nkrishnanChrome:
For this to be impelemented in SeleniumLibrary2.7 i had to make a small change to the existing code.
if you use browser you get attribute error. It makes sense to have current browser in session to scrape the data from the console log

  **return self._current_browser().get_log(log_type)**

aaltat:
@HelioGuilherme66 would you be still interested to contribute this to the project? Because Robot Framework minimum version is raised to 3.0.4 in the SeleniumLibrary 4.0 release and therefore we could delete that dotdic implementation from this PR. Or can we close this PR?

aaltat:
Closing due inactivity.

Issue #925, opened 2017-09-26
Cleanup and enhance keyword documentation


pekkaklarck:
Issue #924 covered enhancing generic library documentation. Similar problems as described in it also in the actual keyword docs. Would be great to go through them too before SL 3.0 final.

pekkaklarck:
Part of this issue is going through open issue related to documentation and see can they be fixed at the same time.

pekkaklarck:
This issue has taken ages to get done because there are so many keywords and some of their docs require a total rewrite. There have also been several nasty surprises with functionality, requiring us to do fixes such as #933, #966, #923, #958 and #942. I hope we've uncovered most of these surprises and it doesn't take too much time to finish this.

Below is a list of files under `src/SeleniumLibrary/keywords/` with files already cleaned up marked:

- [x]  `alert.py`
- [x]  `browsermanagement.py`
- [x]  `cookie.py`
- [x]  `element.py`
- [x]  `formelement.py` (PR #984)
- [x]  `frames.py`
- [x]  `javascript.py` (PR #984)
- [x]  `runonfailure.py`  (PR #984)
- [x]  `screenshot.py` (PR #986)
- [x]  `selectelement.py` (PR #989)
- [x]  `tableelement.py` (PR #992)
- [x]  `waiting.py`  (PR #984)
- [x]  `window.py`

pekkaklarck:
Issue #188 proposed adding examples to `Wait For Condition`. Sounds like a good idea to me but I don't have any good examples in my mind. Any ideas @aaltat or others?

pekkaklarck:
All keyword docs have now been reviewed and cleaned up, but not all PRs with the changes (see [my comment above](https://github.com/robotframework/SeleniumLibrary/issues/925#issuecomment-342012257)) have yet been merged. Once they are, we need to still regenerate the docs and see there are no obvious problems. Otherwise this issue is about to be done.

pekkaklarck:
This issue is finally about to be ready. Still need to review the generated docs and fix possible problems. The docs can be found from http://robotframework.org/SeleniumLibrary/SeleniumLibrary.html

aaltat:
The documentation looks OK for me. Closing the issue.

Issue #449, opened 2015-07-17
Update documentation on Choose File to show that it supports remote uploading


just-be-dev:
The documentation for `Choose File` states that the filepath must be available on the same host where the Selenium Server is running. #445 revealed that this statement is false. 

Selenium is setup to send files remotely to a server, but the files must be located on the system from which the test is running.

aaltat:
Selenium does this magically behind the scenes: https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/remote/webelement.py#L733 and https://github.com/SeleniumHQ/selenium/blob/e74164e117e23e9588170613d7201e4d74245578/py/selenium/webdriver/remote/webelement.py#L505

Issue #202, opened 2013-05-27
Keyword "Input Password" logs given password


Tattoo:
I noticed that if you run Robot Framework with log level TRACE, the keyword Input Password does, in fact, log it's values.

I have attached an example test case with ![a screenshot](https://f.cloud.github.com/assets/280218/566645/8139db2e-c69a-11e2-90c9-7337e4c54f26.png) demonstrating the problem when you run it 

`pybot --loglevel TRACE example.txt`

I feel this should be fixed ASAP.

example.txt:

```
*** Settings ***
Library     Selenium2Library

*** Variables ***
${super secret}=    my super secret password

*** Test cases ***
Example
    Open Browser    http://www.google.com
    Input Password    gbqfq    ${super secret}
    Close All Browsers
```

j1z0:
I actually think the idea of a "super secret" password inside a testing script is a misnomer.  Thinking your password is secure in a script is overly optimistic I would say.  There is nothing secure about a password stored in RF script and the person running the pybot command, needs to have red access to the script anyways, so the log isn't exposing any information that is not in the script already.  The user could simply open up the script and view it.

That being said the documentation for input password does say it doesn't log the password, which as you have identified it clearly does.  However that documentation isn't talking about the TRACE loglevel.  The TRACE logging you are seeing is coming from WebDriver itself and not from RF, that is in effect what TRACE does. So when S2L calls web_driver it is simply calling "send_keys" which doesn't know if your sending a password or not.

As a work around if you don't want passwords to be logged don't use the TRACE loglevel.

Further with webdriver logging passwords this issue has been raised with the Webdriver team here:

http://code.google.com/p/selenium/issues/detail?id=3442

And they are basically just ignoring it....  So I think the that could possible be done is to implement the typePassword function suggested in the Webdriver bug and submit a pull request to the selenium team.  Alternatively it may be possible to temporarily disable TRACE logging when Input Password is called... I don't recall of the top of my head...

But I have to say to that I believe the basic premise of assuming an automation framework should keep secure passwords is simply flawed.

ombre42:
Its worthy to note that, as you highlighted in the screenshot, the password is also visible at level DEBUG. Before distributing the logs, you should use rebot to reprocess the log with a log threshold of INFO if you ran at a level below that.
e.g.
python -m robot.rebot --loglevel INFO output.xml

Tattoo:
[...]

Issue #188, opened 2013-03-29
Add examples to 'Wait for Condition' keyword


emanlove:
A [request](https://groups.google.com/forum/?fromgroups=#!topic/robotframework-devel/RVYqOHdbaZE) was made for better documentation and an example of the 'Wait For Condition' keyword works.

roman-shayda:
return window.document.querySelector('your_CSS_locator').style.visibility != "visible"

pekkaklarck:
This could/should be done as part of #925. Closing this issue.

PR #984, opened 2017-11-05
More documentation cleanup

pekkaklarck:
Documentation (and code) clean-up to:
- formelements.py
- javascript.py
- runonfailure.py
- waiting.py
- Docs about using default locator strategy

See #925.

### DIFF
--- a/src/SeleniumLibrary/keywords/formelement.py
+++ b/src/SeleniumLibrary/keywords/formelement.py
@@ -211,11 +211,25 @@ class FormElementKeywords(LibraryComponent):
     def input_password(self, locator, password):
         """Types the given password into text field identified by ``locator``.
 
-        Difference between this keyword and `Input Text` is that this keyword
-        does not log the given password.
-
         See the `Locating elements` section for details about the locator
         syntax.
+
+        Difference compared to `Input Text` is that this keyword does not
+        log the given password on the INFO level. Notice that if you use
+        the keyword like
+
+        | Input Password | password_field | password |
+
+        the password is shown as a normal keyword argument. A way to avoid
+        that is using variables like
+
+        | Input Password | password_field | ${PASSWORD} |
+
+        Notice also that SeleniumLibrary logs all the communication with
+        browser drivers using the DEBUG level, and the actual password can
+        be seen there. Additionally Robot Framework logs all arguments using
+        the TRACE level. Tests must thus not be executed using level below
+        INFO if password should not be logged in any format.
         """
         self.info("Typing password into text field '%s'." % locator)
         self._input_text_into_text_field(locator, password)
--- a/src/SeleniumLibrary/keywords/formelement.py
+++ b/src/SeleniumLibrary/keywords/formelement.py
@@ -25,11 +25,12 @@ class FormElementKeywords(LibraryComponent):
 
     @keyword
     def submit_form(self, locator=None):
-        """Submits a form identified by `locator`.
+        """Submits a form identified by ``locator``.
 
-        If `locator` is empty, first form in the page will be submitted.
-        Key attributes for forms are `id` and `name`. See `introduction` for
-        details about locating elements.
+        If ``locator`` is not given, first form on the page is submitted.
+
+        See the `Locating elements` section for details about the locator
+        syntax.
         """
         self.info("Submitting form '%s'." % locator)
         if is_noney(locator):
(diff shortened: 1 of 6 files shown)


src/SeleniumLibrary/keywords/formelement.py
```diff
@@ -192,72 +193,78 @@ def select_radio_button(self, group_name, value):
...
     @keyword
     def choose_file(self, locator, file_path):
-        """Inputs the `file_path` into file input field found by `locator`.
+        """Inputs the ``file_path`` into file input field ``locator``.
 
         This keyword is most often used to input files into upload forms.
-        The file specified with `file_path` must be available on the same host
-        where the Selenium Server is running.
+        The file specified with ``file_path`` must be available on machine
+        where tests are executed.
```
aaltat, line 200: This actually might not be true, at least there is an issue about it: https://github.com/robotframework/SeleniumLibrary/issues/449

Also I have recollection that Choose File works also with when using selenium gird.

src/SeleniumLibrary/keywords/formelement.py
```diff
@@ -192,72 +193,78 @@ def select_radio_button(self, group_name, value):
...
     @keyword
-    def input_password(self, locator, text):
-        """Types the given password into text field identified by `locator`.
+    def input_password(self, locator, password):
+        """Types the given password into text field identified by ``locator``.
 
         Difference between this keyword and `Input Text` is that this keyword
-        does not log the given password. See `introduction` for details about
-        locating elements.
+        does not log the given password.
```
aaltat: This is not totally true, the password is logged if loglevel is high enough to display selenium debug messages. See issue  https://github.com/robotframework/SeleniumLibrary/issues/202 for more details.

src/SeleniumLibrary/keywords/waiting.py
```diff
@@ -25,20 +25,17 @@ class WaitingKeywords(LibraryComponent):
 
     @keyword
     def wait_for_condition(self, condition, timeout=None, error=None):
```
aaltat, line 27: There has been request to improve the documentation with examples https://github.com/robotframework/SeleniumLibrary/issues/188 Should we provide few examples in this PR?

src/SeleniumLibrary/keywords/waiting.py
```diff
@@ -51,53 +48,44 @@ def wait_for_condition(self, condition, timeout=None, error=None):
 
     @keyword
     def wait_until_page_contains(self, text, timeout=None, error=None):
```
aaltat, line 50: Should we document that text is searched as it is in the DOM which may not always be same as it is the UI? This is corner case but happens sometimes.

src/SeleniumLibrary/keywords/formelement.py
pekkaklarck: The code [explicitly validates](https://github.com/robotframework/SeleniumLibrary/blob/master/src/SeleniumLibrary/keywords/formelement.py#L194) that the file exists on the machine where the library is installed. That is generally also the machine where tests are executed.

src/SeleniumLibrary/keywords/formelement.py
pekkaklarck: True. I didn't actually remember that SeleniumLibrary logs this on DEBUG level. Need to make it more explicit that password isn't logged on INFO level.

src/SeleniumLibrary/keywords/waiting.py
pekkaklarck: I'd do it separately to get this PR closed. A note about that could be added to #925.

src/SeleniumLibrary/keywords/waiting.py
pekkaklarck: What's the difference in practice? Do you mean that CSS can change the text somehow?

aaltat: I don't know how it happens in practice but there has been several issues complaining l, with proof, that text in the DOM is with small letters and text in UI is in capital letters. And because keyword uses xpath to search the text, the text is searched as it's in the DOM. All the issues that I recall, have always been with text case.

src/SeleniumLibrary/keywords/waiting.py
aaltat: Ok

src/SeleniumLibrary/keywords/formelement.py
aaltat: I have this feeling that it should work, let me try it out and let's see what happens.

src/SeleniumLibrary/keywords/waiting.py
pekkaklarck: CSS can do all kind of text transform (changing case, adding text before/after, ...) and I assume the text to search is the original, not the styled version. That should be tested, though, before we could document it. Such documentation would also belong to the library main level, not to each keyword searching text.

aaltat: OK

aaltat (approved):

PR #989, opened 2017-11-06
Fixes and cleanup to list related keywords

pekkaklarck:
See commit messages of individual commits for more information.

### DIFF
--- a/src/SeleniumLibrary/keywords/selectelement.py
+++ b/src/SeleniumLibrary/keywords/selectelement.py
@@ -290,6 +290,22 @@ class SelectElementKeywords(LibraryComponent):
             select.select_by_visible_text(label)
 
     @keyword
+    def unselect_all_from_list(self, locator):
+        """Unselects all options from multi-selection list ``locator``.
+
+        See the `Locating elements` section for details about the locator
+        syntax.
+
+        New in SeleniumLibrary 3.0.
+        """
+        self.info("Unselecting all options from list '%s'." % locator)
+        select = self._get_select_list(locator)
+        if not select.is_multiple:
+            raise RuntimeError("Un-selecting options works only with "
+                               "multi-selection lists.")
+        select.deselect_all()
+
+    @keyword
     def unselect_from_list(self, locator, *items):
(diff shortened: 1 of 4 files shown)


src/SeleniumLibrary/keywords/selectelement.py
```diff
@@ -18,206 +18,196 @@
...
         select = self._get_select_list(locator)
         return select.first_selected_option.text
 
     @keyword
     def get_selected_list_labels(self, locator):
-        """Returns the visible labels of selected elements (as a list) from the select list identified by `locator`.
+        """Returns labels of selected options from selection list ``locator``.
 
-        Fails if there is no selection.
+        Starting from SeleniumLibrary 3.0, returns an empty list if there
```
aaltat, line 66: Should this chapter be last lines of the doc string? So that in future they are easy to spot and remove?

src/SeleniumLibrary/keywords/selectelement.py
```diff
@@ -18,206 +18,196 @@
...
         select = self._get_select_list(locator)
         return select.first_selected_option.get_attribute('value')
 
     @keyword
     def get_selected_list_values(self, locator):
-        """Returns the values of selected elements (as a list) from the select list identified by `locator`.
+        """Returns values of selected options from selection list ``locator``.
 
-        Fails if there is no selection.
+        Starting from SeleniumLibrary 3.0, returns an empty list if there
```
aaltat, line 92: Should this chapter be last lines of the doc string? So that in future they are easy to spot and remove?

aaltat (approved):

src/SeleniumLibrary/keywords/selectelement.py
pekkaklarck: In my opinion it's fine either way. It's anyway easy to find all references to SeleniumLibrary 3.0 and removing this note isn't hard either. The main reason I decided not to have the note at the end was that we'd end up with two very short paragraphs. If you want to change it, go ahead.

aaltat: Agreed

PR #994, opened 2017-11-09
Adds sphix style definitions to important methods.

aaltat:
Part of #882

### DIFF
--- a/src/SeleniumLibrary/__init__.py
+++ b/src/SeleniumLibrary/__init__.py
@@ -363,6 +363,15 @@ class SeleniumLibrary(DynamicCore):
             raise
 
     def register_driver(self, driver, alias):
+        """Add's a `driver` to the library WebDriverCache.
+
+        :param driver: Instance of the Selenium `WebDriver`.
+        :param alias: Alias given for this `WebDriver` instance.
+        :type driver: selenium.webdriver.remote.webdriver.WebDriver
+        :type alias: intger or string
+        :return: The index of the `WebDriver` instance.
+        :rtype: int
+        """
         return self._drivers.register(driver, alias)
 
     def failure_occurred(self):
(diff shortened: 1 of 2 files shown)


src/SeleniumLibrary/__init__.py
```diff
@@ -363,6 +363,15 @@ def run_keyword(self, name, args, kwargs):
             raise
 
     def register_driver(self, driver, alias):
+        """Add's a `driver` to the library WebDriverCache.
+
+        :param driver: Instance of the Selenium `WebDriver`.
+        :param alias: Alias given for this `WebDriver` instance.
+        :type driver: selenium.webdriver.remote.webdriver.WebDriver
+        :type alias: intger or string
```
pekkaklarck: - I'd prefer having `:type` right after the matching `:param`. 
- Type of alias seems wrong in many ways. Should it be a class, not free text? There's a typo. And alias actually is always a string.
- Do we need to specify the type for all arguments, including strings? I noticed you hadn't added types to such arguments in other places and I think that's OK.

src/SeleniumLibrary/__init__.py
```diff
@@ -402,26 +411,39 @@ def browser(self):
...
-        Returns a list of ``WebElement`` objects. If no matching elements
-        are found, the list is empty. Otherwise semantics are exactly same
-        as with the :meth:`find_element` method.
+        :param locator: Locator to use when searching the element.
+            See library documentation for the supported locator syntax.
+        :param tag: Limit searching only to these elements.
+        :param parent: Optional parent `WebElememt` to search child elements
+            from. By default search starts from the root using `WebDriver`.
+        :type parent: selenium.webdriver.remote.webelement.WebElement
+        :return: list of found `WebElement` or `[]` if element not found.
```
pekkaklarck: This isn't too clear. Wouldn't just `List of found elements.` be enough when the type specified explicitly below. If you want to explain the list can be empty (which ought to be pretty obvious), something like `If no matching elements are found, the list is empty.` could be added as well.

src/SeleniumLibrary/__init__.py
```diff
@@ -402,26 +411,39 @@ def browser(self):
...
+        :rtype: list[selenium.webdriver.remote.webelement.WebElement]
+
+        Always returns a list of `WebElement` objects. If no matching element
+        is found, the list is empty. Otherwise semantics are exactly same
+        as with :meth:`find_element`.
```
pekkaklarck: This is duplicate information now that arguments are explained above and can be removed.

src/SeleniumLibrary/base/context.py
```diff
@@ -60,6 +61,15 @@ def find_element(self, locator, tag=None, required=True, parent=None):
     def find_elements(self, locator, tag=None, parent=None):
         """Find all elements matching `locator`.
 
+        :param locator: Locator to use when searching the element.
+            See library documentation for the supported locator syntax.
+        :param tag: Limit searching only to these elements.
+        :param parent: Optional parent `WebElememt` to search child elements
+            from. By default search starts from the root using `WebDriver`.
+        :type parent: selenium.webdriver.remote.webelement.WebElement
+        :return: list of found `WebElement` or `[]` if element not found.
```
pekkaklarck: Also this explanation can be enhanced and the paragraph below containing duplicate information removed.

PR #1003, opened 2017-11-27
Updates robotlibcore to 1.0rc3 version.

### LINKED ISSUE #1001: Can not call methods which uses @keyword decorated methods to change the keyword name from the library instance when extending library
When doing this: 
```
from SeleniumLibrary import SeleniumLibrary


lib = SeleniumLibrary()
print lib.keywords.get('get_webelements')
```
the print should not display `False`, instead it should return the method. Now the `print lib.keywords.get('Get WebElements')` return the method, which is hard to use in Python side. 

This needs fixing in the https://github.com/robotframework/PythonLibCore/issues/2 side

aaltat:
This fixes problem when calling keywords methods, which uses the
@keyword decorator to change the keyword name, from the library
instance when extending library

Fixes #1001

### DIFF
--- a/src/SeleniumLibrary/base/robotlibcore.py
+++ b/src/SeleniumLibrary/base/robotlibcore.py
@@ -37,13 +37,14 @@ except ImportError:  # Support RF < 2.9
 
 PY2 = sys.version_info < (3,)
 
-__version__ = '1.0rc2'
+__version__ = '1.0rc3'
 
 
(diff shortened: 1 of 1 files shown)


pekkaklarck:
Should we add tests for #1001 in SL side?

aaltat:
It would be good idea. I can think something.

PR #1009, opened 2017-12-04
Fixes Wait Until Element Is Not Visible keyword

### LINKED ISSUE #1008: Change between 1.8.0 and 3.0.0 when using Wait Until Element Is Not Visible. In 3.0.0 element is consider visible if the element does not exist in the DOM.
When I have this:
```
*** Settings ***
Library           Selenium2Library


*** Test Cases ***
Wait Untill Not Visible
    Open Browser    https://www.google.fi/    Chrome
    Wait Until Page Contains Element     id=lst-ib
    Wait Until Element Is Not Visible    id=foobarNotVisible
    [Teardown]    Close All Browsers
```
Then wen I run it with 1.8.0 release, then the test passes. But when I run the test with 3.0.0 release the test fails. More details in [logs_with_differet_versions.zip](https://github.com/robotframework/SeleniumLibrary/files/1526243/logs_with_differet_versions.zip). In the 1.8.0 the element is considered as not visible when the element does not exist in the DOM. But there is change in the 3.0.0 release and element is not considered as invisible if the element is not int he DOM. 
 
It would be best consider element invisible also when the element is not in the DOM.

aaltat:
Fixes #1008

### DIFF
--- a/src/SeleniumLibrary/base/context.py
+++ b/src/SeleniumLibrary/base/context.py
@@ -85,3 +85,7 @@ class ContextAware(object):
         element = self.find_element(locator, tag)
         return (element.is_enabled() and
                 element.get_attribute('readonly') is None)
+
+    def is_visible(self, locator):
+        element = self.find_element(locator, required=False)
+        return element.is_displayed() if element else None
--- a/src/SeleniumLibrary/keywords/waiting.py
+++ b/src/SeleniumLibrary/keywords/waiting.py
@@ -129,7 +129,7 @@ class WaitingKeywords(LibraryComponent):
         ``error`` can be used to override the default error message.
         """
         self._wait_until(
-            lambda: self.find_element(locator).is_displayed(),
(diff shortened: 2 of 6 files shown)


geist-scm:
I used PIP as Tatu outlined and reverted my code in the two places I had 'fixed' back to Wail...Visible kw and it worked with SL 3.0.1.dev1 fix. Yea.

For reference here is the env I tested under showing the SL 3.0.1.dev1 code base:

Linux Dist Ubuntu 16.04.3 LTS
Linux Kernel 4.4.0-97-generic
Git commit eb4eb02
Virtualenv 15.0.1
Python Python 2.7.12
RF 3.0.2
RF Selenium 3.0.1.dev1
Selenium 3.8.0
Requests 2.18.4
EasySNMP 0.2.5
Poster 0.8.1
Chome 61.0.3163.100
Chrome Webdriver 2.33.506092

Thanks for your time,
Paul

PR #1017, opened 2017-12-13
This is for investigate FF problem in Travis

aaltat:
For investigation only. Do not merge to master.


aaltat:
Not needed. Closing.

PR #1020, opened 2017-12-14
Add a new strategy to find element

minhnguyenphuonghoang:
Currently, we can find an element by 1 locator only. 
We need to write multiple keywords to find different element 
locators if we want to and we need to handle those if else & try catch.

Now, I defined a new way to find an element by using a dictionary.

element_a = {
    'name' : 'element_name',
    'locators' : ['xpath', 'css', 'id', ...]
}

By using this way, we can define many ways to find an element.
If the first element's locator is failed to find, keep searching the 2nd then 3rd, etc.

### DIFF
--- a/src/SeleniumLibrary/base/context.py
+++ b/src/SeleniumLibrary/base/context.py
@@ -63,33 +63,23 @@ class ContextAware(object):
             return self.element_finder.find(locator, tag, True, required, parent)
         else:
             """
-            Editing code starting here
-            """
-            # 2. check for element locators
-            element = None
-            locators = locator.get('locators', '')
-            if(len(locators)==0):
-                raise AttributeError('List of locators is empty or you forgot to pass locators attribute.')
-            # 3. go & find the exsting locator
-            found = False
-            for a_locator in locators:
-                try:
-                    element = self.element_finder.find(a_locator, tag, True, required, parent)
-                    found = True
-                    break                
-                except:
-                    continue
-
-            element_name = locator.get('name', '')
-            if(element_name is ''):
-                element_name = ''
-            else:
-                element_name = "'" + element_name + "' "
+            Add a new strategy to find element.
+            locator format showing below:
+            {
+                'name':'locator_name',
+                'locators':['xpath=//locator_xpath','id', 'css'...]
+            }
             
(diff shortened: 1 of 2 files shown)


aaltat:
Interesting idea. Could you raise an issue where you explain/describe your problem and how you would like to solve the problem?

pekkaklarck:
There are various problems with the PR, starting from the code not being Python 3 compatible and tests as well as docs missing. Before getting to that, we needed to decide is this an enhancement we'd like to include in general. First of all, could you show a concrete example how this new functionality could be used?

minhnguyenphuonghoang:
Sure @aaltat and @pekkaklarck 
It's not a problem.
For example, when we do a UI testing, we capture its locator then pass into the locator attribute (keyword argument).
In this way, one time, we can pass 1 type of locator (e.g: xpath, css, id, etc.)
What if I want to have a multiple ways to find an element and if one of those strategies is passed, return my locator?

By using that idea, I implement another keyword to find an element using a dictionary as describled above.
And I modify the error message to print out the element name when the test is failed for easy understanding and easy searching.

So, with either ways, I can find an element by using 1 locator or multiple locators.


BTW, my code doesn't impact to the test but it keeps getting failed.

Please have a look.

pekkaklarck:
I'd like to see a concrete example how this approach is used in Robot Framework data.

pekkaklarck:
It seems to me that the failure on CI is caused by this change. Have you executed all acceptance locally? See contribution guidelines for more information about running tests (and many other important topics such as coding conventions and documentation): https://github.com/robotframework/SeleniumLibrary/blob/master/CONTRIBUTING.rst#code-contributions

minhnguyenphuonghoang:
OK, first, Let me check my code to pass on CI.
Then I'll give you an example how it is used in Robot Framework

aaltat:
When providing the example, try to also describe the need or the problem you want to solve. Now you are describing an solution for the problem and I don't quite understand what is the problem you are trying to solve.

aaltat:
@minhnguyenphuonghoang are you planning to work with the issue or can we close the PR?

aaltat:
Closing due inactivity.

Issue #1021, opened 2017-12-14
Some keywords do not work if text argument is not string


aaltat:
If I have this page:
```
<html>
    <div>1</div>
</html>
```
When I have this test case: 
```
*** Settings ***
Library           Selenium2Library

*** Test Cases ***
Example
    Open Browser    /path/to/example.html    Chrome
    Wait Until Page Contains    ${1}
    [Teardown]    Close Browser
```
When I run the test case with Selenium2Library 1.8.0 the test passes. When I run the same test case with 3.0.1 release, then I see this error:
```

18:35:00.405    FAIL    TypeError: argument of type 'int' is not iterable   
18:35:00.405    DEBUG   Traceback (most recent call last):
  File "c:\python27\lib\site-packages\SeleniumLibrary\__init__.py", line 360, in run_keyword
    return DynamicCore.run_keyword(self, name, args, kwargs)
  File "c:\python27\lib\site-packages\SeleniumLibrary\base\robotlibcore.py", line 102, in run_keyword
    return self.keywords[name](*args, **kwargs)
  File "c:\python27\lib\site-packages\SeleniumLibrary\keywords\waiting.py", line 66, in wait_until_page_contains
    timeout, error)
  File "c:\python27\lib\site-packages\SeleniumLibrary\keywords\waiting.py", line 220, in _wait_until
    self._wait_until_worker(condition, timeout, error)
  File "c:\python27\lib\site-packages\SeleniumLibrary\keywords\waiting.py", line 227, in _wait_until_worker
    if condition():
  File "c:\python27\lib\site-packages\SeleniumLibrary\keywords\waiting.py", line 64, in <lambda>
    self._wait_until(lambda: self.is_text_present(text),
  File "c:\python27\lib\site-packages\SeleniumLibrary\base\context.py", line 81, in is_text_present
    locator = "xpath://*[contains(., %s)]" % escape_xpath_value(text)
  File "c:\python27\lib\site-packages\SeleniumLibrary\utils\__init__.py", line 27, i
```

aaltat:
I did read code and noticed that we are not constant on the matter. With some keywords number is OK and with other it is not OK to have a number.

This would be good time to decide what object types we do actually support. I my opinion, we should support  `str`. Then I might be willing support `number` and nothing else.  Nothing else covers `False`, `None` and other Python object types. Also for non supported types we should raise `TypeError`.

[...]

PR #1024, opened 2017-12-20
Fix for #1021:Wait Until Page Contains keyword does not work if text …

badari412:
The "escape_xpath_value function" in utils.py expects the argument to be of type string. But when we use the integer variables like ${1}, ${123} etc. this functions throws the error "TypeError: argument of type 'int' is not iterable". So added a line of code to convert the argument to string.

### DIFF
--- a/src/SeleniumLibrary/utils/__init__.py
+++ b/src/SeleniumLibrary/utils/__init__.py
@@ -24,6 +24,7 @@ from .webdrivercache import WebDriverCache
 
 
 def escape_xpath_value(value):
+    value = str(value)
     if '"' in value and '\'' in value:
         parts_wo_apos = value.split('\'')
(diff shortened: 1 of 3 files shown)


aaltat:
Before thinking further with this PR, we must decide what objects we should convert to string or should we do conversation at all. There is an issue about the problem in #1021, let's move the discussion to that issue.

aaltat:
I gave the matter some tough and did update the #1021 issue. Please read the issue and please decide are you willing to improve the PR according the issue.

badari412:
@aaltat, I have gone through your update and I will improve my PR accordingly. Thanks for your input! Will let you know what I plan to implement.

aaltat:
We decided that this will be done when we drop Python 2 support. Which will happen after 2020 when Python 2 support ends. Therefore I am closing the PR.

PR #1027, opened 2017-12-26
WebDriverFactory to ease creating the WebDriver for different browsers.

aaltat:


aaltat:
This was not good idea. Will be implemented differently in new PR.

Issue #849, opened 2017-07-13
Verify string regardless of case


rubygeek:
I have a two websites that I want to verify with the same code. Problem is one website the disclaim is in ALL CAPS and one is lowercase. Example

```
Use of this site constitutes acceptance of our

USE OF THIS SITE CONSTITUTES ACCEPTANCE OF OUR
```
So i propose the following api change:

```
Element Should Contain    locator, expected, message= 
```
to 

```
Element Should Contain    locator, expected, case_sensitive=True, message=
``` 

Default of course being True to avoid breaking changes and when the developer wants case insensitive, they can pass `case_sensitive=False`

I am willing to make these changes and submit a PR if it might be approved.

aaltat:
The idea sounds good. Although I have nagging feeling, in my head, that there was something that which denies doing this. But because I can not remember what that nagging feeling is, perhaps it is not important. 

From the API point of view, it would be more backwards compatible if the API would work like this: 
```
| Element Should Contain | locator | expected | message='' | case_sensitive=True |
```
Also there would be other keyword which would benefit from the same kind functionality (assuming that my nagging feeling does not prevent the implementation), like `Element Should Not Contain`, `Element Text Should Be`,  `Wait Until Element Contains`, `Wait Until Element Does Not Contain`, `Wait Until Page Contains` and `Wait Until Page Does Not Contain`. If you can find an implementation for a one keyword, would you be willing to enhance other keywords too?

rubygeek:
Yes, I'd be happy to make this consistent across related keywords.

pekkaklarck:
This would be handy but can be surprisingly big task:
- If we use Selenium's logic for assertions and it doesn't support case-insensitivity, we need to move the whole assert logic to the library side.
- Testing and documenting these changes is a big task regardless how easy implementation is. This got very clear when same support was added to BuiltIn's comparison keywords recently:
https://github.com/robotframework/robotframework/issues/2439

Instead of adding case-insensitivity support to S2L, it would be possible to use its getter keywords (e.g. `Get Title`) in combination with BuiltIn keywords that support case-insensitivity. If case-insensitivity in the lib is seen valuable enough to be implemented, it should preferably use same `ignore_case=False` argument as BuiltIn.

aaltat:
[...]

PR #1034, opened 2018-01-19
Add ignore_case argument to element_should_contain 

SidePlace:


SidePlace:
Trying to help on #849 
Forgot to mention bug ID in the commit message

aaltat:
Code looks OK, but PR is missing test. PR are not merged without test backing them up.

SidePlace:
Hi @aaltat I just pushed my tests for it. Let me know if I'm doing something improperly.

aaltat:
We have had lately problems with the Travis CI, sometimes the test execution stops. It seems that this PR is also suffering from the same problem. I did restart the two failed runs and lets see how it goes.

test/acceptance/ignoreCase.robot
```diff
@@ -0,0 +1,19 @@
+*** Settings ***
```
aaltat, line 1: There already is a existing suite for this keyword, in ./test/acceptance/keywords/content_assertions.robot You have two options:
1) Move tests from here to the content_assertions suite
2) Move the `Element Should Contain` keyword test all in same and new suite. Also change the suite name to describe the whole keyword, not just the new feature.

test/acceptance/ignoreCase.robot
```diff
@@ -0,0 +1,19 @@
...
+Library           SeleniumLibrary
+Test Teardown     Close All Browsers
+
+*** Variables ***
+${ISSUE_URL}      https://github.com/robotframework/SeleniumLibrary/issues/849
```
aaltat, line 6: Using links to external resources are not allowed. The acceptance test runner starts small http server. which must be used to serve the required html files. Please take an example from the ./test/acceptance/keywords/content_assertions.robot suite. 

The test logic is OK,

aaltat (changes requested):
There are two things to fix.

aaltat:
If you want credit for your work, please add your name and short description what you have done in the https://github.com/robotframework/SeleniumLibrary/blob/master/CHANGES.rst

aaltat:
Closing on favour of #1043.

PR #1039, opened 2018-01-28
Webdriver factory 2

### LINKED ISSUE #550: Allow Open Browser keyword to pass capabilities also for local browsers
Sometimes it could be useful to allow capabilities be passed also to local browsers. The current implementation of the `Open Browser` only passes the capabilities  to the Selenium side only when `remote_url` argument is used.

aaltat:
Fixes #550


aaltat:
Need to fix the problem with `GeckoDriver.log` which in this PR is created now in the folder where `robot` is started.

aaltat:
Also have to fix that unit test work with older versions of Selenium.

pekkaklarck (approved):

PR #1063, opened 2018-02-24
NOT FOR FOR MASTER

aaltat:
Only to debug the hang problem in Travis


aaltat:
Closing and opening to get more info and prove that it is the Open Browser keyword that always fails.

aaltat:
This is NOT intended to master, closing.

PR #1087, opened 2018-03-12
1073-get-attribute-value-should-be

brian-mathews:
Added `Element Attribute Value Should Be` to capture element attribute and compare expected value.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,5 +1,8 @@
 Release Notes
 =============
+3.2.1
+-----
+- Added `Element Attribute Value Should Be` to capture element attribute and compare expected value. [brian-mathews]
 
 3.2.0
 -----
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,12 +1,9 @@
 Release Notes
 =============
-3.2.1
------
-- Added `Element Attribute Value Should Be` to capture element attribute and compare expected value. [brian-mathews]
 
 3.2.0
 -----
-To be defined.
+- Added `Element Attribute Value Should Be` verifies element identified by locator contains expected attribute value.. [brian-mathews]
 
 3.1.1
 -----
--- a/atest/acceptance/keywords/elements.robot
+++ b/atest/acceptance/keywords/elements.robot
@@ -57,10 +57,22 @@ Get Element Attribute
 
 Get Element Attribute Value Should Be
     Element Attribute Value Should Be  link=Absolute external link  href  http://www.google.com/
-    Element Attribute Value Should Be  link=Relative to sub-directory  href  http://localhost:7000/html/sub/index.html
     Element Attribute Value Should Be  id=image_id  src  http://localhost:7000/html/image.jpg
     ${attribute_value}=  Get Element Attribute  css=#second_div  class
     Element Attribute Value Should Be  css=#second_div  class  ${attribute_value}
+    Element Attribute Value Should Be  link=Link with Unicode äöüÄÖÜß  href  http://localhost:7000/html/index.html
+    Run Keyword And Expect Error
+    ...    Test Fail Custom Message
+    ...    Element Attribute Value Should Be  id=image_id  href  http://non_existing.com  message=Test Fail Custom Message
+    Run Keyword And Expect Error
+    ...    Element 'id=image_id' attribute should have value 'http://non_existing.com' but its value was 'None'.
+    ...    Element Attribute Value Should Be  id=image_id  href  http://non_existing.com
+    Run Keyword And Expect Error
+    ...    Element with locator 'id=non_existing' not found.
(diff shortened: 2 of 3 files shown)


src/SeleniumLibrary/keywords/element.py
```diff
@@ -383,6 +383,25 @@ def get_element_attribute(self, locator, attribute=None):
             locator, attribute = locator.rsplit('@', 1)
         return self.find_element(locator).get_attribute(attribute)
 
+    @keyword
+    def element_attribute_value_should_be(self, locator, attribute, expected, message=None):
+        """Compares returned value from ``attribute`` identified by ``locator``
```
aaltat: Verifies element identified by ``locator`` contains ``expected`` ``attribute`` value.

src/SeleniumLibrary/keywords/element.py
```diff
@@ -383,6 +383,25 @@ def get_element_attribute(self, locator, attribute=None):
...
+        to expected ``value``.
+
+        See the `Locating elements` section for details about the locator
+        syntax.
+
+        Example:
+        `Element Attribute Value Should Be` | css:img | href | value
+        """
```
aaltat: In doc string, add as last line: `New in SeleniumLibrary 3.2.`

src/SeleniumLibrary/keywords/element.py
```diff
@@ -383,6 +383,25 @@ def get_element_attribute(self, locator, attribute=None):
...
+        current_expected = self.find_element(locator).get_attribute(attribute)
+        if current_expected != expected:
+            if is_noney(message):
+                message = "Element '%s' attribute should have value '%s' but "\
```
aaltat: Joining strings with `\` is not good practice. Please use example: 
``` 
message = ("Element '%s' attribute should have value '%s' but "
           "its value was '%s'." % (locator, expected, current_expected))
```

CHANGES.rst
```diff
@@ -1,5 +1,8 @@
 Release Notes
 =============
+3.2.1
```
aaltat: Next version is 3.2.0. Move the text to correct version.

atest/acceptance/keywords/elements.robot
```diff
@@ -55,6 +55,13 @@ Get Element Attribute
     ${class}=    Get Element Attribute    ${second_div}    class
     Should Be Equal    ${class}    Second Class
 
+Get Element Attribute Value Should Be
+    Element Attribute Value Should Be  link=Absolute external link  href  http://www.google.com/
+    Element Attribute Value Should Be  link=Relative to sub-directory  href  http://localhost:7000/html/sub/index.html
+    Element Attribute Value Should Be  id=image_id  src  http://localhost:7000/html/image.jpg
+    ${attribute_value}=  Get Element Attribute  css=#second_div  class
+    Element Attribute Value Should Be  css=#second_div  class  ${attribute_value}
```
aaltat: More test should be written. 
1) Write test where keyword also fails and default error message is used.
2) Write test where custom error message is used.
3) Write test which uses non-ascii characters (like ö, ä or å) in locator. 
4) Write test where the element does not exist.
5) Write test which uses non-ascii characters (like ö, ä or å) in attribute value. 
6) Add test which uses WebElment as locator.

atest/acceptance/keywords/elements.robot
```diff
@@ -55,6 +55,13 @@ Get Element Attribute
     ${class}=    Get Element Attribute    ${second_div}    class
     Should Be Equal    ${class}    Second Class
 
+Get Element Attribute Value Should Be
+    Element Attribute Value Should Be  link=Absolute external link  href  http://www.google.com/
+    Element Attribute Value Should Be  link=Relative to sub-directory  href  http://localhost:7000/html/sub/index.html
+    Element Attribute Value Should Be  id=image_id  src  http://localhost:7000/html/image.jpg
```
aaltat: Lines 59 to 61 are testing the same thing, consider removing few of the lines.

aaltat (changes requested):
Generally code looks OK. There are few comments for doc string and it would be good to write more tests too.

aaltat:
I did restart tests in `appveyor` CI, because sometimes the tests using  Firefox hangs. Generally PR looks good, but there where some small changes needed.

aaltat:
By looking from a mobile, all looks good to go. But I need to take a second look from a bigger screen, because I sometimes miss stuff from the small screen.

aaltat:
After reading did decide that it would be better do split the one test case to smaller chunks and I did in a separate commit. One the CI runs and assuming that I did not make any mistakes, this is ready to merged.

aaltat (approved):

aaltat:
Darn, I make lots of typos and I did just missed your commit. Good catch. Will fix it in a separate commit.

Issue #1090, opened 2018-03-21
Add message to Location Should Be keyword


rubygeek:
The keyword `Location Should Be` and `Location Should Contain` would be more useful to show a custom error message like many of the other keywords that take a `message` attribute.

http://robotframework.org/SeleniumLibrary/SeleniumLibrary.html#Location%20Should%20Be
http://robotframework.org/SeleniumLibrary/SeleniumLibrary.html#Location%20Should%20Contain
I would be happy to work on this or assist one of my teammates on this task.

aaltat:
This is a good enhancement. Are you able to provide a pull request?

rubygeek:
Yes we will, thanks!

aaltat:
This is done #1091 Closing the issue.

PR #1091, opened 2018-03-22
#1090 added code & test for the issue

taniabhullar:
For issue #1090

### DIFF
--- a/src/SeleniumLibrary/keywords/browsermanagement.py
+++ b/src/SeleniumLibrary/keywords/browsermanagement.py
@@ -239,21 +239,41 @@ class BrowserManagementKeywords(LibraryComponent):
         return self.driver.current_url
 
     @keyword
-    def location_should_be(self, url):
-        """Verifies that current URL is exactly ``url``."""
+    def location_should_be(self, url, message=None):
+        """Verifies that current URL is exactly ``url`.
+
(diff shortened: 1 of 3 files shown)


src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -239,21 +239,41 @@ def get_location(self):
...
+        message.
+
+        ``message`` argument new in SeleniumLibrary 3.2.0.
+        """
         actual = self.get_location()
         if actual != url:
-            raise AssertionError("Location should have been '%s' but was "
-                                 "'%s'." % (url, actual))
+            if is_noney(message):
+                message = "Location should have been '%s' but "\
```
aaltat: Splitting strings with `\` creates code which is easy to break. It would be better to do it like this: 
```
message = ("Location should have been '%s' but "
           "was '%s'." % (url, actual))
```

Please also fix line 274 and 275.

aaltat:
Because #1087 was merged to master, there is a git conflict in `CHANGES.rst`. Could you fix the conflict and the one comment too?

taniabhullar:
Yes, will look into all that. Thanks!!

aaltat:
The chances looks good from the mobile screen and all test have passed. I need to take a look from a bigger screen but I don't expect any problems from that.

PR #1123, opened 2018-05-17
fix for broken tests

rubygeek:
fix for broken tests

### DIFF
--- a/atest/resources/html/javascript/click.html
+++ b/atest/resources/html/javascript/click.html
@@ -1,5 +1,6 @@
 <html>
   <head>
+    <meta charset="UTF-8">
     <script type="text/javascript">
       function singleClick() {
         document.getElementById("output").innerHTML="single clicked";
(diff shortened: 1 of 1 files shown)


aaltat:
Could you remind me what was broken?

rubygeek:
```
Click Element Error :: FAIL Element with locator 'id:äääääää' not ... | FAIL |
Element with locator 'id:äääääää' not found.
Double Click Element Error :: FAIL Element with locator 'id:öööö' ... | FAIL |
Element with locator 'id:öööö' not found.
```

been seeing that for while now when i run the tests

rubygeek:
any input for this PR?

aaltat:
Sorry, I have been enjoying Finland summer (usually it's rain and hails this time of year) and it's 20 - 27°C warm. Really usual but pleasant surprise.

But back to the PR. Yes those test should fail and it's OK that they do fail. We use robotstatuschecker, https://github.com/robotframework/statuschecker/ to check certain things, like log messages or keyword status from the execution. And the robotstatuschecker generates the log and report html files. Therefore one should not look at the test status at the run time. Instead one should look at the summary report in the end, or the report and log files.

PR #1135, opened 2018-06-12
Add method get_session_id

IlfirinPL:
Session id is required to download files using selenoid https://github.com/aerokube/selenoid/blob/master/docs/file-download.adoc

### DIFF
--- a/atest/acceptance/open_and_close.robot
+++ b/atest/acceptance/open_and_close.robot
@@ -58,3 +58,21 @@ Closing all browsers clears cache
     Run Keyword And Expect Error
     ...    No browser with index or alias 'Browser 2' found.
     ...    Switch Browser    Browser 2
+    
+Get Session Id
+    Open Browser    ${ROOT}/forms/prefilled_email_form.html    ${BROWSER}    Browser 1
+    ...    remote_url=${REMOTE_URL}    desired_capabilities=${DESIRED_CAPABILITIES}
+    Open Browser    ${ROOT}/forms/prefilled_email_form.html    ${BROWSER}    Browser 2
+    ...    remote_url=${REMOTE_URL}    desired_capabilities=${DESIRED_CAPABILITIES}
+    Switch Browser    Browser 1
+    ${browser 1}    Get Session Id
+    Switch Browser    Browser 2
+    ${browser 2}    Get Session Id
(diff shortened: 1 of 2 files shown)


aaltat:
Without going further for the PR, two things are needed:

Please write an issue and explain what problem you are trying to solve. We also use issues for internal tracking.

Please write test for the new functionality.

IlfirinPL:
I am currently using selenoid that is replacement of selenium grid in some ways better and to utils download file feature ( browser download file) to get it locally machine to execute tests you need to use link ex. http://selenoid-host.example.com:4444/download/**f2bcd32b-d932-4cdc-a639-687ab8e4f840**/myfile.txt
part of this link is session id,

this functionality doesn't add nothing new only expose already existing variable used in Open Browser Keyword but only as part of log in debug mode

aaltat:
I can understand your needs and therefore it's valid thing to expose as a keyword. To the PR get merged, new functionality needs to be tested. Also it might be good idea to log the session id too.

Also the documentation should be improved, to tell what the keyword does, not why you need it. Example documentation could say: `Returns the currently active browser session id.`

IlfirinPL:
As general rule you don't add unit test to getters and setters, also check similar methods that already exists ex.  get_source ,get_title and they don't add logs and any units test, so to keep it consistent its should stay the same. I agree with documentation already checked.

aaltat:
Unit test are not required in this case, but there must be acceptance test.

pekkaklarck:
- I agree every kw should have a test. 
- Would `Get Session Id` be enough as a name? I found that `Browser` part unnecessary and also a bit confusing as it is, AFAIK, Selenium session id.
- Doc should have a note that this is new in SL 3.2.

IlfirinPL:
I agree with shorter name, initial thought was that it would be to generic name

IlfirinPL:
Acceptect test added

aaltat:
I added a note about the new in SL 3.2. When Travis shows green and assuming that I did not made a mistake, PR is ready to merge.

aaltat:
All is green and this is good for merge. Thanks you from the contribution. If you want credit, in release notes, add your name and short description in here: 
https://github.com/robotframework/SeleniumLibrary/blob/master/CHANGES.rst

IlfirinPL:
merge done

PR #1140, opened 2018-06-16
Select windows timeout

### LINKED ISSUE #757: Keyword to wait that windows is open

aaltat:
Fixes #757

### DIFF
--- a/atest/acceptance/windows.robot
+++ b/atest/acceptance/windows.robot
@@ -127,6 +127,16 @@ Select Popup Window By Excluded List
     Select Window    ${parent}
     Title Should Be    Click link to show a popup window
 
+Select Popup Window With Delay By Excluded List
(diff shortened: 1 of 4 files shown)


aaltat:
Writing test where keyword also fails would be good thing.

aaltat:
Also remove `import time` from `windowmanager.py`.

PR #1144, opened 2018-06-24
Warning when multiple elements

aaltat:


atest/resources/html/javascript/click.html
```diff
@@ -16,6 +22,8 @@ <h1>Clicking</h1>
     <p id="output">initial output</p>
     <button type="button" id="singleClickButton" onclick="singleClick()">Click me!</button>
     <button type="button" id="doubleClickButton" ondblclick="doubleClick()">Double click me!</button>
+    <button type="button" id="twoButtons" onclick="twoButtons1()">Button 1</button>
+    <button type="button" id="twoButtons" onclick="twoButtons2()">Button 2</button>
```
rubygeek, line 26: wouldn't this be invalid html? I know with jquery, it would always choose first matching ID .. suggest if you really want to test this use classes.

src/SeleniumLibrary/base/context.py
```diff
@@ -53,13 +55,22 @@ def find_element(self, locator, tag=None, required=True, parent=None):
...
         :return: Found `WebElement` or `None` if element not found and
             `required` is false.
         :rtype: selenium.webdriver.remote.webelement.WebElement
         :raises SeleniumLibrary.errors.ElementNotFound: If element not found
             and `required` is true.
         """
-        return self.element_finder.find(locator, tag, True, required, parent)
+        element, multiple = self.element_finder.find(locator, tag, True,
+                                                     required, parent)
+        if multiple and warn:
```
rubygeek, line 70: so if I am selecting the collection of li which have class "foobar" a selector "css=.foobar"  this will warn?

aaltat: That is the basic idea, but we should not log the warning always. The full idea is, if keyword should find single element, then it should display warning that multiple elements where found. Because the element the SeleniumLibrary will choose, it always will choose first one returned by Selenium, may not be the correct one to choose. This is meant to protect the users against changes in the DOM and for the users new to the Selenium automation. Example for `Click Button` there would be warning logged, but not for `Page Should Contain`. 

The PR is not yet ready, because now too many keywords will log the warning and I would have to choose which keywords actually will log the warning. But I am seeking for a input, exactly like this, that am I going to the right direction for implementing the feature. Also I am looking technical input, because there is many ways to implement the feature and I have changed my mind few times, where I should I implement the feature. 

When I did write #978, the feature idea did sounded like a great idea for me. Now when implementing the feature, I am having second thoughts. But those are mainly because I can not decide where to the technical implementation should go. In any case, I would love to get the feedback.

atest/resources/html/javascript/click.html
pekkaklarck: Yeah, using same id multiple times is invalid and how browsers handle it may differ. In these tests it probably would be better to avoid them and use, for example, same class with different elements.

aaltat:
After discussion, I have better implementation in mind. Therefore closing this one.

Issue #323, opened 2014-08-29
Allow arguments to be passed into JavaScript scripts


ombre42:
Selenium allows arguments to be passed along with JavaScript that can be referenced in the script using the injected List variable named `arguments`. _Execute Javascript_, _Execute Async Javascript_, and _Wait For Condition_ do not support this feature. This would be useful for passing WebElements and data structures into scripts

When we implement _Get Webelement(s)_, people will naturally want to pass the WebElements as argument(s) to a script. Currently, one can use _Assign Id To Element_ to make a WebElement easy to reference if JS if it is difficult to reference is JS.

Implementing this without breaking test cases or adding keywords would require using keyword arguments for _Execute Javascript_ and _Execute Async Javascript_. Because they have variable arguments, we cannot add a positional argument. Keyword arguments require RF 2.8 to work.

There is a workaround for this missing functionality on [StackOverflow](http://stackoverflow.com/questions/21256845/is-there-a-way-to-provide-arguments-to-execute-javascript-in-robotframework/21273554#21273554)

pekkaklarck:
Requiring RF 2.8 shouldn't be a problem because it's not that new release anymore. I don't think minimum RF requirement needs to be raised, though, but obviously that needs to be documented in kw doc.

ombre42:
After getting far into adding this feature using keyword arguments, I realized there is a serious backwards incompatibility issue. Assignments or comparisons in JavaScript are triggering Robot Framework's free keyword argument processing. Example:
`Execute Javascript    var elem = document.getElementById('foo'); elem.selected = true;`
Fails with message _'var elem ' is not a valid keyword argument_. The equals sign would have to be escaped:
`Execute Javascript    var elem \= document.getElementById('foo'); elem.selected \= true;`
I pushed [a commit](https://github.com/ombre42/robotframework-selenium2library/commit/3dac91b5e3160d5e1866d7cad7ac7cd3c4688c20) if you want to see the progress thus far. I am not going to do anything more until we have a consensus.

I see three options to implement this feature:
1) Implement with keyword arguments, breaking _many_ scripts and always needing to escape '='s
2) Remove the capability to split the code in multiple cells (change sig to `execute_javascript(code, *args)`). This will break tests that split the code (far less than option 1), unfortunately with the vague error _WebDriverException: Message: u'JavaScript error'_.
[...]

PR #1147, opened 2018-06-26
Added execute_javascript_with_arguments keyword

arnewohletz:
Added PR as discussed in
https://github.com/robotframework/SeleniumLibrary/issues/323

### DIFF
--- a/src/SeleniumLibrary/keywords/javascript.py
+++ b/src/SeleniumLibrary/keywords/javascript.py
@@ -82,26 +82,8 @@ class JavaScriptKeywords(LibraryComponent):
         """ Executes the given JavaScript code with the given arguments
         #TODO: Add docstring
         """
-
-        if len(code) == 1:
-            js = self._get_javascript_to_execute(code)
-            self.info("Executing JavaScript:\n%s" % js)
-            return self.driver.execute_script(js)
-
-        elif len(code) > 1:
-            code, args = self._parse_javascript_and_arguments(code)
-
-            if len(args) == 0:
-                self.info("Executing JavaScript: \n%s" % code)
-                return self.driver.execute_script(code)
-            else:
-                args_list = []
-                for idx, item in enumerate(args):
-                    args_list.append("arguments[" + str(idx) + "]: " +
-                                     str(item))
-                self.info("Executing JavaScript: \n%s\nwith arguments \n%s"
-                          % (code, "\n".join(args_list)))
-                return self.driver.execute_script(code, *args)
+        code, args = self._get_javascript_to_execute(code)
+        self._execute_javascript_code(code, *args)
 
     @keyword
     def execute_async_javascript_with_arguments(self, *code):
@@ -109,36 +91,49 @@ class JavaScriptKeywords(LibraryComponent):
         arguments
         #TODO: Add docstring
         """
+        code, args = self._get_javascript_to_execute(code)
+        self._execute_async_javascript_code(code, args)
 
-        if len(code) == 1:
-            js = self._get_javascript_to_execute(code)
-            self.info("Executing Asynchronous JavaScript:\n%s" % js)
-            return self.driver.execute_async_script(js)
+    def _prepare_argument_list(self, args):
+        self.args_list = []
+        for idx, item in enumerate(args):
+            self.args_list.append("arguments[" + str(idx) + "]: " +
+                                  str(item))
 
-        elif len(code) > 1:
-            code, args = self._parse_javascript_and_arguments(code)
+    def _execute_javascript_code(self, code, *args):
+        if len(args) > 0:
+            self._prepare_argument_list(args)
+            self.info("Executing JavaScript: ""\n%s\n"
+                      "with arguments \n%s" % (
+                          code, "\n".join(self.args_list)))
+        else:
+            self.info("Executing JavaScript:\n%s" % code)
+        self.driver.execute_script(code, *args)
 
-            if len(args) == 0:
-                self.info("Executing Asynchronous JavaScript: \n%s" % code)
-                return self.driver.execute_async_script(code)
-            else:
-                args_list = []
-                for idx, item in enumerate(args):
-                    args_list.append("arguments[" + str(idx) + "]: " +
-                                     str(item))
-                self.info("Executing Asynchronous JavaScript: ""\n%s\n"
-                          "with arguments \n%s" % (code, "\n".join(args_list)))
-                return self.driver.execute_async_script(code, *args)
+    def _execute_async_javascript_code(self, code, *args):
+        if len(args) > 0:
+            self._prepare_argument_list(*args)
+            self.info("Executing Async JavaScript: ""\n%s\n"
+                      "with arguments \n%s" % (
+                          code, "\n".join(self.args_list)))
+            return self.driver.execute_script(code, *args)
+
+        else:
+            self.info("Executing Async JavaScript: \n%s" % code)
+            return self.driver.execute_script(code)
 
     def _get_javascript_to_execute(self, lines):
-        code = ''.join(lines)
-        path = code.replace('/', os.sep)
-        if os.path.isabs(path) and os.path.isfile(path):
-            code = self._read_javascript_from_file(path)
-        return code
+        if all(isinstance(n, str) for n in lines):
+            code = ''.join(lines)
+            path = code.replace('/', os.sep)
+            if os.path.isabs(path) and os.path.isfile(path):
+                code = self._read_javascript_from_file(path)
+                return code, []
+
+        code, args = self._parse_javascript_and_arguments(lines)
+        return code, args
 
     def _parse_javascript_and_arguments(self, code):
-
         is_arg = False
         is_code = True
         self.codelines = ''
@@ -163,4 +158,3 @@ class JavaScriptKeywords(LibraryComponent):
                   .format(path.replace(os.sep, '/'), path), html=True)
         with open(path) as file:
             return file.read().strip()
-
--- a/src/SeleniumLibrary/keywords/javascript.py
+++ b/src/SeleniumLibrary/keywords/javascript.py
@@ -79,10 +79,25 @@ class JavaScriptKeywords(LibraryComponent):
 
     @keyword
     def execute_javascript_with_arguments(self, code, *arguments):
+        """ Executes the JavaScript code with the given arguments
+
+        """
         js = self._get_javascript_to_execute(code)
         self.info("Executing JavaScript \n%s with arguments \n%s" % (
             js, str(*arguments)))
         return self.driver.execute_script(js, *arguments)
+
+    @keyword
+    def execute_javascript_with_arguments(self, code, *arguments):
+        """ Executes asynchronous JavaScript code with the given arguments
+
+
+        """
+
+        js = self._get_javascript_to_execute(code)
+        self.info("Executing Asynchronous JavaScript \n%s with arguments \n%s"
+                  % (js, str(*arguments)))
+        return self.driver.execute_async_script(js, *arguments)
 
     def _get_javascript_to_execute(self, lines):
         code = ''.join(lines)
--- a/src/SeleniumLibrary/keywords/javascript.py
+++ b/src/SeleniumLibrary/keywords/javascript.py
@@ -78,14 +78,22 @@ class JavaScriptKeywords(LibraryComponent):
         return self.driver.execute_async_script(js)
 
     @keyword
-    def execute_javascript_with_arguments(self, code, *arguments):
+    def execute_javascript_with_arguments(self, *code):
         """ Executes the given JavaScript code with the given arguments
         #TODO: Add docstring
         """
+
+        code, args = self._parse_javascript_and_arguments(code)
         js = self._get_javascript_to_execute(code)
-        self.info("Executing JavaScript \n%s with arguments \n%s" % (
-            js, str(*arguments)))
-        return self.driver.execute_script(js, *arguments)
+
+        if len(args) == 0:
+            self.info("Executing JavaScript \n%s" % (
+                js))
+            return self.driver.execute_script(js)
+        else:
+            self.info("Executing JavaScript \n%s\nwith arguments \n%s" % (
+                code, args))
+            return self.driver.execute_script(code, args)
 
     @keyword
     def execute_async_javascript_with_arguments(self, code, *arguments):
@@ -104,8 +112,29 @@ class JavaScriptKeywords(LibraryComponent):
             code = self._read_javascript_from_file(path)
         return code
 
+    def _parse_javascript_and_arguments(self, code):
+        is_arg = False
+        is_code = True
+        self.codelines = ''
+        self.args = []
+
+        for exp in code:
+            if str(exp).strip() == "JAVASCRIPT":
+                is_code = True
+                is_arg = False
+            elif str(exp).strip() == "ARGUMENTS":
+                is_arg = True
+                is_code = False
+            elif is_code:
+                self.codelines = self.codelines.__add__(exp + '\n')
+            elif is_arg:
+                self.args.append(exp)
+
+        return self.codelines, self.args
+
     def _read_javascript_from_file(self, path):
         self.info('Reading JavaScript from file <a href="file://%s">%s</a>.'
                   .format(path.replace(os.sep, '/'), path), html=True)
         with open(path) as file:
             return file.read().strip()
+
--- a/src/SeleniumLibrary/keywords/javascript.py
+++ b/src/SeleniumLibrary/keywords/javascript.py
@@ -77,6 +77,13 @@ class JavaScriptKeywords(LibraryComponent):
         self.info("Executing Asynchronous JavaScript:\n%s" % js)
         return self.driver.execute_async_script(js)
 
+    @keyword
+    def execute_javascript_with_arguments(self, code, *arguments):
+        js = self._get_javascript_to_execute(code)
+        self.info("Executing JavaScript \n%s with arguments \n%s" % (
+            js, str(*arguments)))
+        return self.driver.execute_script(js, *arguments)
+
     def _get_javascript_to_execute(self, lines):
         code = ''.join(lines)
         path = code.replace('/', os.sep)
--- a/src/SeleniumLibrary/keywords/javascript.py
+++ b/src/SeleniumLibrary/keywords/javascript.py
@@ -79,8 +79,8 @@ class JavaScriptKeywords(LibraryComponent):
 
     @keyword
     def execute_javascript_with_arguments(self, code, *arguments):
-        """ Executes the JavaScript code with the given arguments
-
+        """ Executes the given JavaScript code with the given arguments
+        #TODO: Add docstring
         """
         js = self._get_javascript_to_execute(code)
         self.info("Executing JavaScript \n%s with arguments \n%s" % (
@@ -88,12 +88,10 @@ class JavaScriptKeywords(LibraryComponent):
         return self.driver.execute_script(js, *arguments)
 
     @keyword
-    def execute_javascript_with_arguments(self, code, *arguments):
+    def execute_async_javascript_with_arguments(self, code, *arguments):
         """ Executes asynchronous JavaScript code with the given arguments
-
-
+        #TODO: Add docstring
         """
-
         js = self._get_javascript_to_execute(code)
         self.info("Executing Asynchronous JavaScript \n%s with arguments \n%s"
                   % (js, str(*arguments)))
--- a/src/SeleniumLibrary/keywords/javascript.py
+++ b/src/SeleniumLibrary/keywords/javascript.py
@@ -83,27 +83,52 @@ class JavaScriptKeywords(LibraryComponent):
         #TODO: Add docstring
         """
 
-        code, args = self._parse_javascript_and_arguments(code)
-        js = self._get_javascript_to_execute(code)
-
-        if len(args) == 0:
-            self.info("Executing JavaScript \n%s" % (
-                js))
+        if len(code) == 1:
+            js = self._get_javascript_to_execute(code)
+            self.info("Executing JavaScript:\n%s" % js)
             return self.driver.execute_script(js)
-        else:
-            self.info("Executing JavaScript \n%s\nwith arguments \n%s" % (
-                code, args))
-            return self.driver.execute_script(code, args)
+
+        elif len(code) > 1:
+            code, args = self._parse_javascript_and_arguments(code)
+
+            if len(args) == 0:
+                self.info("Executing JavaScript: \n%s" % code)
+                return self.driver.execute_script(code)
+            else:
+                args_list = []
+                for idx, item in enumerate(args):
+                    args_list.append("arguments[" + str(idx) + "]: " +
+                                     str(item))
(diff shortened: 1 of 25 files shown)


aaltat:
Each new functionality must be backed up with tests. I think this one needs few. But before that, I should read more about the JavaScript.

aaltat:
I did read the Selenium API and I understand how that works. I agree that having separate keyword is easier in this phase but we should think what shall we do in the long run (not required in this PR).

I also noticed that async JavaScript method in Selenium API supports arguments: http://selenium-python.readthedocs.io/api.html#selenium.webdriver.remote.webdriver.WebDriver.execute_async_script For consistency reason we should create a similar keyword for http://robotframework.org/SeleniumLibrary/SeleniumLibrary.html#Execute%20Async%20Javascript
Also the PR must be backed up with tests. Would you @horsewithnoname1985 be willing to update the PR.

emanlove:
Not sure if you have worked out a design but I was thinking that we could use "docstring" markup to delineate the JavaScript allowing for multi-line JavaScript and then have the arguments following the "docstring". This should allow for any JavaScript code and multiple lines. It might also be helpful to provide a warning if an argument is not called out in the JavaScript. I do like, at least in the interim, have keywords named `Execute JavaScript With Arguments` and `Execute Async JavaScript With Arguments` as a way to introduce these implementations.

aaltat:
You mean something like Pekka did propose in the #323 that keyword signature is` *code_and_args` 
```
** Keywords ***
Highlight Element
    [Arguments]    ${locator}
    ${element}=    Get WebElement    ${locator}
    Execute Javascript
    ...   element = arguments[0];
    ...   original_style = element.getAttribute('style');
    ...   element.setAttribute('style', original_style + "; background: yellow;");
    ...   setTimeout(function(){
    ...        element.setAttribute('style', original_style);
    ...   }, 300);
    ...   ARGUMENTS
    ...   ${element}
```

That `ARGUMENTS` is a separator between the code and arguments?

emanlove:
I need to review Pekka proposal but I was thinking something like

```
Execute JavaScript With Arguments   """this = my_javascript_code;"""  ${arg0}  ${arg1}
```

@horsewithnoname1985, what do you think?

aaltat:
Here is the comment that I was referring: https://github.com/robotframework/SeleniumLibrary/issues/323#issuecomment-100034998

But it would be tempting to keep support for variable number of lines for the code and arguments.

emanlove:
Taking that example what I was thinking the markup, if you will, be would be

```
*** Keywords ***
Highlight Element
    [Arguments]    ${locator}
    ${element}=    Get WebElement    ${locator}
    Execute Javascript   """
    ...   element = arguments[0];
    ...   original_style = element.getAttribute('style');
    ...   element.setAttribute('style', original_style + "; background: yellow;");
    ...   setTimeout(function(){
    ...        element.setAttribute('style', original_style);
    ...   }, 300);"""
    ...   ${element}  ${anotherarg}  ${1}
```

where the docstring style is used to delineate a single argument. This of course was wishful thinking on my part. Here is my unit test to see if what I was thinking may be part of the robot framework language or if this idea was possible

```
***Test Cases***
DocString As Argument
    ${docstring}=  Set Variable  """This is a single line"""
    Length Should Be   ${docstring}  27
    Should Not Be Empty  ${docstring}
    Should Contain  ${docstring}   """

DocString With Double Space
    ${docstring}=  Set Variable  """First Name  Last Name"""
    Length Should Be   ${docstring}  27

Multi-Line DocString Using RF Syntax
    ${docstring}=  Set Variable  """This is the first line.
                   ...              This is the second line.
		   ...  This is the third line.
		   ...  This is the fourth and final line."""
    
    Length Should Not Be   ${docstring}  4


Multi-Line DocString Without RF Syntax
    ${docstring}=  Set Variable  """This is the first line.
                                     This is the second line.
		                     This is the third line.
		                     This is the fourth and final line."""
    
    Length Should Not Be   ${docstring}  4
```

noting of course that `Length Should Not Be` is not a keyword itself.

This solution that I propose starts to take into consideration the concept of an object type - docstring - just as there are number variables or boolean variables.

aaltat:
Now I see your idea. The basic idea is to have keyword which has one `*vararg` and keyword argument can present the JavaScript code and JavaScript arguments. Code and arguments are separated with a marker. In your and Pekka idea, the markers are different but the basic concept is pretty much same. 

The more I think about the idea, the more I like it. Would have to agree only with the marker. It should be clear and something which is not already reserved by Robot Framework or by JavaScript.

emanlove:
Yes, the same with a different marker. In my mind I do see a design where the triple docstring quotes would delineate into a single argument. So one could have whatever text in a docstring, like

```
Evalute DocString With Double Space
    ${docstring}=  Evaluate  return """First Name  Last Name"""
    Log To Console  ${docstring}
```

This, support of docstrings in robot framework, of course may not be widely useful and this may be a ["extractive idea"](http://ivory.idyll.org/blog/2018-how-open-is-too-open.html).

aaltat:
Hmmm. I would prefer a single marker, something similar that is used for `Run Keyword If` has for `ELSE` and `ELSE IF`. The biggest problem I see for the docstring is that I doesn't have support in the editor/IDE space. Although docstring is supported in the Python side, in Robot Framework data doesn't have support for it. It would be colored in same way as any other arguments and it doesn't stand out. Therefore sporting errors comes more difficult and it doesn't stand out as an separator. Any more good ideas for a separator, anyone?

arnewohletz:
Sorry for the little mess I made with the recent commits here.
I added another keyword ``execute_async_javascript_with_arguments`` and also added a test [(see here)](https://github.com/robotframework/SeleniumLibrary/pull/1147/commits/2fe9d9caaef5aa58919732048b28eca2a8ea2de2) for the ``execute_javascript_with_arguments``. I borrowed the design from the [scroll_into_view](https://github.com/robotframework/SeleniumLibrary/blob/master/atest/acceptance/keywords/scroll_into_view.robot) test, which, I hope, is OK. I am not sure on how to test the new asynchronous java scripts keyword, so I haven't added any test for this so far.

I like the idea of supporting multiline JavaScript code. Enabling this by putting the code inside a docstring seems very straightforward, so I support that. I’d still prefer a keyword with two arguments though, where the first argument is for the (multi-line) JavaScript code and the other's a varargs for all passed in arguments, instead of assigning the first of multiple arguments as the code:
`Execute JavaScript with Arguments  ${code}  ${arg1}  ${arg2] ${argX}`

As for the delimiter: Maybe a multi-line JavaScript code can be defined in the ``*** Variables***`` table using the ``SEPARATOR= \n`` attribute:
```
*** Variables ***
${my_multiline_js_code}=    SEPARATOR=\n
...    my_first.line_of_code; 
...    my_second.line_of_code;  
...    my_x.line.of_code;
```
But I haven't tested that

aaltat:
The down side of two arguments is that then we lose capability to define JavaScript inline. Example this is not anymore possible.
```
| Execute JavaScript | Some code |
| ... | More code here |
| ... | ARGUMENTS |
| ... | ${variable1} |
| ... | ${variable2} |
```
Instead the code should be created as a list with a separate keyword and that is a down side. Example something like this:
```
| ${code} | Create List |Some code |
| ... | More code here |
| Execute JavaScript | ${code} |
| ... | ${variable1} |
| ... | ${variable2} |
```
In long term, this causes a big backwards compatible issue and means that it's almost impossible to to join the two keywords together. Therefore I am now against the two argument implementation.

As a second issue, we should first decide one separator and not to support multiple different separators. Like the zen of python says: There should be one -- and preferably only one -- obvious way to do it. https://www.python.org/dev/peps/pep-0020/#id3

I am somewhat against the docstring, because the reasons I did explain above. Instead I would like to see something which is similar than what `Run Keyword If` uses for separating the evaluation statements. Something like `ARGUMENTS` feels natural for me, but I don't feel strongly what the actual text is.

aaltat:
The problem with the commits is easy to fix. Are you able to fix it by yourself or do you need help with it? If you need help with with it, would you be able to do ` git remote -v` from command line and post the output.

arnewohletz:
Seems the error has resolved already. 
I see your point about using two arguments in the keyword. 

How about using the backslash, like in the FOR LOOP syntax, here:

 ```robotframework
Execute JavaScript
...    :JAVASCRIPT
...    \ element = arguments[0];
...    \ original_style = element.getAttribute('style');
...    \ element.setAttribute('style', original_style + "; background: yellow;");
...    \ setTimeout(function(){
...    \    element.setAttribute('style', original_style);
...    \ }, 300);
```

If used with arguments, those could be integrated either via a 'ARGUMENTS' attribute like this

 ```robotframework
Highlight Element
[Arguments]    ${locator}    ${unused_js_arg}
${element}=    Get WebElement    ${locator}
    Execute Javascript with arguments
    ...    :JAVASCRIPT
    ...    \ element = arguments[0];
    ...    \ original_style = element.getAttribute('style');
    ...    \ element.setAttribute('style', original_style + "; background: yellow;");
    ...    \ setTimeout(function(){
    ...    \    element.setAttribute('style', original_style);
    ...    \ }, 300);
    ...    ARGUMENTS    ${element}    ${unused_arg}
```
or preferably like this

 ```robotframework
Highlight Element
[Arguments]    ${locator}    ${unused_js_arg}
${element}=    Get WebElement    ${locator}
    Execute Javascript with arguments
    ...    :JAVASCRIPT
    ...    \ element = ${element};
    ...    \ original_style = element.getAttribute('style');
    ...    \ element.setAttribute('style', original_style + "; background: yellow;");
    ...    \ setTimeout(function(){
    ...    \    element.setAttribute('style', original_style);
    ...    \ }, 300);
    ...    ARGUMENTS    ${element}    ${unused_js_arg}
```

or even automatically imported into the JavaScript

 ```robotframework
Highlight Element
[Arguments]    ${locator}    ${unused_js_arg}
${element}=    Get WebElement    ${locator}
    Execute Javascript with arguments
    ...    :JAVASCRIPT
    ...    \ element = ${element};
    ...    \ original_style = element.getAttribute('style');
    ...    \ element.setAttribute('style', original_style + "; background: yellow;");
    ...    \ setTimeout(function(){
    ...    \    element.setAttribute('style', original_style);
    ...    \ }, 300);
```

aaltat:
I am on mobile and need to take a look with device that has a bigger screen.

aaltat:
The first marker, `JAVASCRIPT` feels redundant when the JavaScript does not contain any arguments. Example like this:  
```
Execute JavaScript
...    JAVASCRIPT
...    element = arguments[0];
...    original_style = element.getAttribute('style');
...    element.setAttribute('style', original_style + "; background: yellow;");
...    setTimeout(function(){
...       element.setAttribute('style', original_style);
...    }, 300);
````
But it would allow to change the order of the JavaScript code and arguments, example like this: 
```
Execute Javascript with arguments
...    ARGUMENTS
...    ${element}
...    ${unused_arg}
...    JAVASCRIPT
...    element = arguments[0];
...    original_style = element.getAttribute('style');
...    element.setAttribute('style', original_style + "; background: yellow;");
...    setTimeout(function(){
...       element.setAttribute('style', original_style);
...    }, 300);
```
Similar feature request was asked/proposed in the [issue](https://github.com/robotframework/SeleniumLibrary/issues/323#issuecomment-99900171).

But I do not see benefit of supporting the FOR like syntax and using `\ `-character at start of the line, It just extra and not related characters which requires extra parsing and writing from users.

I am thinking, would this be good proposal. 
1) Support current style:  
```
Execute JavaScript
...    element = arguments[0];
...    original_style = element.getAttribute('style');
...    element.setAttribute('style', original_style + "; background: yellow;");
...    setTimeout(function(){
...       element.setAttribute('style', original_style);
...    }, 300);
````
2) Support arguments in the end:
```
Execute Javascript
...    element = arguments[0];
...    original_style = element.getAttribute('style');
...    element.setAttribute('style', original_style + "; background: yellow;");
...    setTimeout(function(){
...       element.setAttribute('style', original_style);
...    }, 300);
...    ARGUMENTS
...    ${element}    ${unused_arg}
```
3) Support changing the order of arguments and code:
```
Execute Javascript
...    ARGUMENTS
...    ${element}    ${unused_arg}
...    JAVASCRIPT
...    element = arguments[0];
...    original_style = element.getAttribute('style');
...    element.setAttribute('style', original_style + "; background: yellow;");
...    setTimeout(function(){
...       element.setAttribute('style', original_style);
...    }, 300);
```
4) Support defining the `JAVASCRIPT` at start (optional)
```
Execute Javascript
...    JAVASCRIPT
...    element = arguments[0];
...    original_style = element.getAttribute('style');
...    element.setAttribute('style', original_style + "; background: yellow;");
...    setTimeout(function(){
...       element.setAttribute('style', original_style);
...    }, 300);
...    ARGUMENTS
...    ${element}    ${unused_arg}
```
The point 4 is optional and largely depends which way is the implementation better, with the support or without it. 

What others think, @emanlove and @horsewithnoname1985?

**Comment by aaltat** [comment]
Did forget to say that in this way, we do not need a new keyword and the new feature can be implemented with the existing JavaScript keywords. The change is backwards incompatible but it is unlikely that someone would have `ARGUMENTS` and `JAVASCRIPT` as single arguments in their code.

**Comment by emanlove** [comment]
I agree with the idea that docstrings is a bad idea and I like yours and Pekka's idea for ARGUMENTS as a delimiter (No. 2).

**Comment by arnewohletz** [comment]
I also approve the ARGUMENTS marker and I also would very much appreciate the JAVASCRIPT marker to switch order, as it is more natural that arguments are defined before using them. @aaltat , I agree, that these feature are probably better implemented into the existing keywords instead of creating new ones.

**Comment by aaltat** [comment]
It looks like we have an agreement. Lets go with the `ARGUMENTS` and `JAVASCRIPT` markers. With `JAVASCRIPT` being optional and possibility to swap `ARGUMENTS` and `JAVASCRIPT` markers order. Would @horsewithnoname1985 be willing to implement this?

**Comment by arnewohletz** [comment]
Sorry for my late reply. I will attempt to implement of the discussed features as soon as possible

**Comment by aaltat** [comment]
No worries and take your time.

**Comment by arnewohletz** [comment]
Mh, I'm experiencing difficulties working with a web element that was passed into the JavaScript code. I am using a test where I scroll to an element, that is not initially visible on the page and JavaScript refuses to apply the method onto my web element. The following error is logged:
```JavascriptException: Message: TypeError: arguments[0].scrollIntoView is not a function```
Apparently, my passed object is not a valid DOM object, which is needed here ([see here](https://wiki.selfhtml.org/wiki/JavaScript/DOM/Element/scrollIntoView)). I tested my previous scrolling keyword, which is working fine. I looked for differences in the objects passed into the script and found that my working script passes a
```selenium.webdriver.remote.webelement.WebElement``` object, whereas the not working one passes a ```selenium.webdriver.firefox.webelement.FirefoxWebElement``` object. I don't know, if that is the reason, since the latter only extends the former, but this is the only difference I see. When I assign the to be scrolled web element inside the JavaScript code (i.e. via Document.getElementById()), it also works fine.

Here is my previous working keyword implemention:
```python
@keyword
def scroll_to_element(self, locator):
	seleniumlib = BuiltIn().get_library_instance('SeleniumLibrary')
	javascriptkeywordlib = JavaScriptKeywords(ctx=seleniumlib)

	element = SeleniumLibrary.find_element(seleniumlib, locator)
	script = "return arguments[0].scrollIntoView();"

	JavaScriptKeywords.execute_javascript_with_arguments(
		javascriptkeywordlib,
		script, element)
```

I updated the pull request with the failing implementation. Currently, I still use the ```execute_javascript_with_arguments```keyword for my implementation.

src/SeleniumLibrary/keywords/javascript.py
```diff
@@ -77,15 +77,64 @@ def execute_async_javascript(self, *code):
...
+        js = self._get_javascript_to_execute(code)
+
+        if len(args) == 0:
+            self.info("Executing JavaScript \n%s" % (
+                js))
+            return self.driver.execute_script(js)
+        else:
+            self.info("Executing JavaScript \n%s\nwith arguments \n%s" % (
+                code, args))
+            return self.driver.execute_script(code, args)
```
aaltat: By reading the code, which makes it a guess, I think the problem is in this line. You should unpack the `args` and not pass it as a list. So instead `(code, args)` write `(code, *args)`.

aaltat (commented):
I did put a comment on the line where I think the bug is by reading the code.

arnewohletz:
Thanks @aaltat, that fixed the issue. I updated my pull request, but still kept the new functionality in its own keywords ```Execute JavaScript with arguments``` and ```Execute Async JavaScript with arguments``` until we have an agreement, that the keywords work as intended. I also didn't update the docstring yet. Is that something, you'd rather like to do yourself?

aaltat:
I think the main focus should be to add the arguments support for existing keywords. Because the parsing logic for `ARGUMENTS` and `JAVASCRIPT`  is somewhat complicated, it would best to test it in the unit test. Then acceptance test can be on correct level and there needs to less of them.

If you need help, where you would need it?

arnewohletz:
I'll try to create a proper unittest. Thanks for your help

aaltat:
I was thinking this a bit and this would be quite perfect opportunity to try out approval test https://github.com/approvals/ApprovalTests.Python Are you familiar with the approval test concept? Or unit test and TDD in general?

arnewohletz:
No, I'm not familiar with the approval test concept, but I know some about TDD. Let me have a look at it.

aaltat:
This could be something that we could work together. I am busy for few days, but I this is something that we could work together.

What is your timezone and timetable after few days? Would it be possible talk over slack or Skype?

aaltat:
I did get exited and here is an example how to do it: https://github.com/robotframework/SeleniumLibrary/pull/1155 But there is few things more investigated.

arnewohletz:
I added some approval tests for the keyword. What do you think?

aaltat:
Sorry, I have been on holidays and therefore did not have time to investigate the errors. I looked Travis and there are two failures with Jython and pypy. Most likely the problem with pypy can be solved and I can take a look at the problem. But the problem with Jython is more complicated, because approval test dependency installation fails. I can research also this problem.

About your code, move the complex logic from keywords to own methods. The keywords should call two methods, one which parses `ARGUMENTS` and `JAVASCRIPT`. And the second one calling the Selenium API to execute the JavaScript code.

aaltat:
I looked the error with pypi and it looks like an bug in the ApprovalTests implementation. I did raise an issue about it: https://github.com/approvals/ApprovalTests.Python/issues/29 and lets see how it goes. I have not looked the problem with Jython and what could be done.

aaltat:
I did provide a fix for the approval test, to fix support for pypy. 

I have been thinking about the Jython problem and the best idea that I have is to: 
1) Tune the Jython Travis configuration so that approval test is not installed with Jython.
2) Make `try/except` in the approval tests imports in unit tests.
3) When running in Jython, implement skip for tests using approval tests.

aaltat:
After thinking this little bit more. I think the best solution it to make separate PR, where all the infrastructure details are implemented for using the approval tests. I can create an issue and sort out the the implementation details. Once that is done, we can continue with this PR.

aaltat:
The support for ApprovalTests is added. The ApprovalTests now works with pypy and for Jython the tests must be skipped. There is an example in the [test_javascript.py](https://github.com/robotframework/SeleniumLibrary/blob/master/utest/test/keywords/test_javascript.py) file. Although the reporter might require some tuning, to make it more configurable for different OS and needs, the basic idea works.

Would you be willing to continue the work for the PR?

BeniKeyserman:
@horsewithnoname1985 Is there any progress on this PR? 
Do you need any help?

arnewohletz:
@aaltat , sorry for my very late reply. I also had a holiday break. Yes, I will continue on this PR within these days. So I assume the unit tests using approvaltests for the new keyword are fine? Then I will work on the keyword implementation logic as you suggested on July 27th.
Thanks, @BeniKeyserman , I'll come back to you, if I need any help with this.

aaltat:
No worries, holiday should be spend in the sun rather than coding.

I did also play with the ApprovalTest, just because I wanted to learn more about them and #1176 shows what I did have in mind. Please note that the `_separate_code_and_args` method is not ready for production. It needs some refactoring so that code easier to read and understand. Also the error messages should be more meaningful. But there are few differences in our approaches of testing. 

I think you misunderstood how `verify_all` is used. Instead of passing a single result, one should pass [list of results](https://github.com/robotframework/SeleniumLibrary/pull/1176/files#diff-f7401d5faf698f02d89f91e4ba6c39abR48). When passing list of result, the [results](https://github.com/robotframework/SeleniumLibrary/pull/1176/files#diff-f65a6de679d5cf6480793d68cccc6694) are easier to read with the reporter. The main idea of ApprovalTests is about visualizing the results in string format.

Your unit test do not have error testing. I did write some examples, how error handling can be [tested by using ApprovalTests](https://github.com/robotframework/SeleniumLibrary/pull/1176/files#diff-f7401d5faf698f02d89f91e4ba6c39abR51).

Also there are missing test, from me and you, like reading the JavaScript from file. 

But because master has moved forward, there are few few conflicts which needs to be sorted out. Also your PR contains few commits which are already in master and should not be part of this PR. Are you able to sort out the problems or do you need help?

arnewohletz:
Hi @aaltat , I see your point and will have a look on these tests next. I saw your ` _separate_code_and_args`method. We should bring this method, and the changes I just committed together.
One unit test currently fails, because I made changes in the `_get_javascript_to_execute` method. The unit test sends a single string, whereas RF tests would send a one string item tuple into the method for a one line javascript code. Since the users doesn't call that method directly, could the unit test be changed from:
```python
def test_get_javascript(self):
    code, args = self.javascript._get_javascript_to_execute('code here')
    ...
```
to 
```python
    def test_get_javascript(self):
        code, args = self.javascript._get_javascript_to_execute(tuple(['code here']))
    ...
```
?
Did something occur to the Travis CI? I did not see any test being executed after the commit.
Thanks for your support on the approvaltests. You're right, I misunderstood something. I will add your proposals to the PR next and see how it goes.

aaltat:
Changing the unit test is fine, although I would change it like this: `self.javascript._get_javascript_to_execute(('code here`, ))`

The Travis problem is related to your PR:  how you solving the conflicts. But that is actually a small problem compared to the problem that the commit history in this PR is mixed, meaning that commits which are already merged in master are somewhere in the between of your commits.  In practice, this commit tries to rewrite the history of SeleniumLibrary and that is not acceptable, because it will lead to greater problems. Instead commit history should be: all commits which are in the master are in the bottom of the `git log` command. Then your commits are top of those commits. In this way git diff will actually show the files and lines which you have changed and doesn't show the files which someone else have changed. This will also ease the review process.

There is at least two ways to fix the problem: cherry-pick and rebase. By quickly looking, I would assume that cherry-pick might be the easiest way, but I am not sure. Are you able to solve the problem or do you want me to do it for you?

aaltat:
Closing in favor of #1176

arnewohletz:
Hi @Tattoo , sorry for not dealing with this issue in time. I see you moved this issue into a less confusing PR

Tattoo:
@horsewithnoname1985 I'm not @aaltat :)

aaltat:
No worries. I did try to recover the commit history from this PR, but it was actually quite a big task. Therefore I did decide to create a new one.

PR #1148, opened 2018-06-29
 add take_screenshot_on_element keyword

baikaishuipp:
add take_screenshot_on_element keyword

### DIFF
--- a/src/SeleniumLibrary/keywords/screenshot.py
+++ b/src/SeleniumLibrary/keywords/screenshot.py
@@ -26,7 +26,6 @@ def scop_to(filename, left, top, right, bottom):
     im = Image.open(filename) # uses PIL library to open image in memory
     im = im.crop((left, top, right, bottom)) # defines crop points
     im.save(filename) # saves new cropped image
-    print 'crop', filename
 
 class ScreenshotKeywords(LibraryComponent):
 
--- a/src/SeleniumLibrary/keywords/screenshot.py
+++ b/src/SeleniumLibrary/keywords/screenshot.py
@@ -21,6 +21,12 @@ from robot.utils import get_link_path
 from SeleniumLibrary.base import LibraryComponent, keyword
 from SeleniumLibrary.utils import is_noney
 
+def scop_to(filename, left, top, right, bottom):
+    from PIL import Image
+    im = Image.open(filename) # uses PIL library to open image in memory
+    im = im.crop((left, top, right, bottom)) # defines crop points
+    im.save(filename) # saves new cropped image
+    print 'crop', filename
 
 class ScreenshotKeywords(LibraryComponent):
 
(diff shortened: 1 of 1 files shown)


aaltat:
The keyword takes an screenshot from a element, did I understand correctly? Please correct me if I am understanding the keyword wrong. Rest of the comments assume that my assumption is correct and I need to think again if my assumption was incorrect.

This idea has been proposed also before and then we did reject the idea, but I think we should reconsider the idea. Back then the idea was rejected, because similar functionality exist in the https://github.com/datakurre/robotframework-selenium2screenshots The reason for reconsidering the idea is that similar functionality actually exist in the Selenium API: http://selenium-python.readthedocs.io/api.html#selenium.webdriver.remote.webelement.WebElement.screenshot and the idea of the library is to expose Selenium API has a useful keywords. Because of that reasoning I am favour of adding the keyword to SeleniumLibrary. What do others think?

@baikaishuipp could you please try is the functionality offered by the Selenium API sufficient for your needs? If it's good enough for your needs, would you be willing to change the PR to use the Selenium API? Also if you are filling to change the PR, we should then discuss about the used documentation format in keywords and we should discuss how one can test the keyword in best possible way.

baikaishuipp:
@aaltat Yeah, U understand correctly. I will  try the Selenium API and check it works or not.

baikaishuipp:
@aaltat I tried, the Selenium API does not work. 
OS: Windows 10

Selenium Version: 3.13.0

Browser: Chrome

Browser Version: 67

There is an error when use it:`WebDriverException: Message: unknown command: session/{session_id}element/{element_id}/screenshot`. 

But It works for Firefox...

aaltat:
Did you try with the latest version of Chromedriver=2.40? If you did I did spot an open issue for [Chromedirver](https://bugs.chromium.org/p/chromedriver/issues/detail?id=1938&q=element%20screenshot&colspec=ID%20Status%20Pri%20Owner%20Summary). But because it is in the specification, I hope that it gets implemented. You could go and express your need for the issue to be resolved in Chromedriver.

baikaishuipp:
chromedriver=2.40 does not works, and this issue not include in 2.40

There is the release log below: 
Latest Release: ChromeDriver 2.40
Supports Chrome v66-68

Changes include:
Fixed Chromedriver hang on open when user-data-dir is specified and exists
ChromeDriver supports IPv6 on requests
Fixed Chromedriver couldn't find the Android file using valid file path
/session/:sessionId/send_command and /session/:sessionId/send_command_and_get_result changed to proper extension commands

aaltat:
Thank you for the input. I need to think for a while what to do with this feature. Having keyword by using the Selenium API would be useful and easy to do. But if the Selenium API works only with limited number of browsers, example only with Firefox, then having keyword that works only Firefox is not useful. Then your original implementation would be better, but it requires adding PIL as a dependency and it also clashes with the implementation of the https://github.com/datakurre/robotframework-selenium2screenshots I need to think about this for a while.

baikaishuipp:
@aaltat pls let me know if you got decision

aaltat:
Well I was planning to think over the weekend. I am in mixed feelings.

aaltat:
After some consideration, I am not going to approve the PR, because similar functionality exist in the https://github.com/datakurre/robotframework-selenium2screenshots Also having PIL as a dependency is a negative for me, because I am not familiar with the PIL.

emanlove:
My (late) two cents .. if the Selenium API now has a method for this I could see creating a SeleniumLibrary keyword for it. This runs along the general guideline that the SeleniumLibrary is wiring Selenium APIs into Robot Framework. For PIL we can use PILLOW which is essentially an update version of PIL and one that is pip compatible. Of this would add a dependency to the library but if it would be required for the Selenium API anyways.. it may be a non issue.

PR #1175, opened 2018-08-15
Fixed documentation for custom locator example, so argument lines up with usage

DanielPBak:
Self-explanatory. The example here is incorrect, I have fixed it.

### DIFF
--- a/src/SeleniumLibrary/__init__.py
+++ b/src/SeleniumLibrary/__init__.py
@@ -201,7 +201,7 @@ class SeleniumLibrary(DynamicCore):
 
     This keyword is a reimplementation of the basic functionality of the
     ``id`` locator where ``${browser}`` is a reference to a WebDriver
-    instance and ``${strategy}`` is name of the locator strategy. To use
+    instance and ``${locator}`` is name of the locator strategy. To use
     this locator it must first be registered by using the
     `Add Location Strategy` keyword:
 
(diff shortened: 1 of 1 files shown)


aaltat (changes requested):
There is a bug in the documentation. I would actually change the `criteria` and `strategy` to `locator`. Please note that `strategy` is referenced at least once in the chapter and that actually should be changed also. Would you be willing to make this change?

aaltat:
Thanks for pointing out the error. I did have an improvement idea, would you be willing to do what I did suggest in the comment?

DanielPBak:
Sure!

DanielPBak:
I have implemented the changes you requested :)

aaltat (commented):
Also in here: https://github.com/robotframework/SeleniumLibrary/blob/master/src/SeleniumLibrary/__init__.py#L204 word strategy should be changed to locator.

aaltat:
Sorry for the long delay. It looks ok in the example level, but you missed a word in few lines below. Could you fix that one too?

aaltat:
Thanks for the contribution. If you want, you can add your name to: https://github.com/robotframework/SeleniumLibrary/blob/master/CHANGES.rst

DanielPBak:
Sure! https://github.com/robotframework/SeleniumLibrary/pull/1178

PR #1176, opened 2018-08-16
JavaScript parser improvement

aaltat:


aaltat:
Fixes #323

aaltat:
Need to take a look on few things:
1) Are error messages good.
2) Can `_separate_code_and_args` simplified. Now it has quite many moving parts. 
3) Proof read the new keyword docs. The error message says `marker` but doc says `statement`

atest/acceptance/keywords/javascript_alert.js
```diff
@@ -0,0 +1 @@
+alert(arguments[0]);
```
pekkaklarck: Newline missing at the end of file.

src/SeleniumLibrary/keywords/javascript.py
```diff
@@ -42,19 +43,31 @@ def execute_javascript(self, *code):
...
-        | `Should Be Equal`    | ${sum}               | ${2}          |
+        | `Execute JavaScript` | alert(arguments[0]); | ARGUMENTS | 123 |
+        | `Execute JavaScript` | ARGUMENTS | 123 | JAVASCRIPT | alert(arguments[0]); |
         """
-        js = self._get_javascript_to_execute(code)
-        self.info("Executing JavaScript:\n%s" % js)
-        return self.driver.execute_script(js)
+        js_code, js_args = self._get_javascript_to_execute(code)
+        self.info('Executing JavaScript:\n%s\nBy using argument(s):\n"%s"'
+                  % (js_code, ', '.join(js_args)))
```
pekkaklarck: Logging could be enhanced a little:
- Better not to mention anything about arguments if there are none. Current `""` looks odd.
- Instead of `"argument(s):"` it's possible to use `"argument%s:" % robot.utils.plural_or_not(js_args)` to get the trailing `s` when needed.
- Quotes around arguments like `"first, second"` look a bit strange. I'd consider omitting them altogether or alternatively quoting each argument separately. In the latter case `robot.utils.seq2str` or `seq2str2` might be worth a look. Just logging `js_args` as-is would be fine otherwise as well, but the `u` prefixes with Unicode strings inside a list in Python 2 are annoying.

src/SeleniumLibrary/keywords/javascript.py
```diff
@@ -73,19 +91,52 @@ def execute_async_javascript(self, *code):
...
-
-    def _get_javascript_to_execute(self, lines):
-        code = ''.join(lines)
-        path = code.replace('/', os.sep)
-        if os.path.isabs(path) and os.path.isfile(path):
-            code = self._read_javascript_from_file(path)
-        return code
+        js_code, js_args = self._get_javascript_to_execute(code)
+        self.info('Executing Asynchronous JavaScript:\n%s\nBy using argument(s):\n"%s"'
+                  % (js_code, ', '.join(js_args)))
```
pekkaklarck: Same comment about logging as above. Logging should probably be done using a helper.

src/SeleniumLibrary/keywords/javascript.py
```diff
@@ -73,19 +91,52 @@ def execute_async_javascript(self, *code):
...
+                get_code = False
+                get_args, found_args = True, True
+                continue
+            if not get_code and not get_args:
+                get_code, found_code = True, True
+            if get_code:
+                js_code.append(line)
+            if get_args:
+                js_args.append(line)
+        return js_code, js_args
```
pekkaklarck: This looks a bit complicated. Don't have time to look at more closely could it be improved. Will look again later.

pekkaklarck (commented):
Looks good. Some nitpickey line notes added. I'll review the code related to finding JS code and arguments when I got time.

src/SeleniumLibrary/keywords/javascript.py
aaltat: I did not know that such utils did exist. I definitely will take a look.

src/SeleniumLibrary/keywords/javascript.py
aaltat: Yes, it's complicated. At least raising an error could be moved to a separate method. Need to think what else could be done.

src/SeleniumLibrary/keywords/javascript.py
aaltat: Helper is good idea. Will definitely take a look.

atest/acceptance/keywords/javascript_alert.js
aaltat: Done

PR #1213, opened 2018-10-04
Fix deprecation warning about inspect.getargspec

### LINKED ISSUE #1212: DeprecationWarning in base/robotlibcore.py:115
I get a lot of DeprecationWarnings when using this library (it's great!) when running it on Python 3.

```
DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() or inspect.getfullargspec()
```

Looking at the documentation, it's a quick-fix as the replacement is nearly drop-in:

https://docs.python.org/3/library/inspect.html#inspect.getfullargspec
https://docs.python.org/2/library/inspect.html#inspect.getargspec

I can submit a PR for this.

adongy:
Note that Python 3 added keywords only arguments and defaults but they are ignored here.

Closes #1212

### DIFF
--- a/src/SeleniumLibrary/base/robotlibcore.py
+++ b/src/SeleniumLibrary/base/robotlibcore.py
@@ -112,13 +112,18 @@ class DynamicCore(HybridCore):
         return args
 
     def _get_arg_spec(self, kw):
-        spec = inspect.getargspec(kw)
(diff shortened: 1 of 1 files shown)


adongy:
Closing this PR, check the related issue for the PR on the base project.

PR #1216, opened 2018-10-04
Fix missing backquote in Location Should Be documentation

### LINKED ISSUE #1215: Typo in Location Should Be documentation
In SeleniumLibrary 3.2.0, a backquote is missing in the `Location Should Be` documentation, around the `url` parameter, resulting in an invalid reStructuredText syntax.
This is visible in the HTML documentation itself and in the reports:
http://robotframework.org/SeleniumLibrary/SeleniumLibrary.html#Location%20Should%20Be
```
Verifies that current URL is exactly `url.
```

cbonte:
Fixes #1215


aaltat:
Thanks for the fix.

Issue #1239, opened 2018-11-22
Add keyword to mimic add blocker functionality


aaltat:
When testing websites, they should be tested with add blocker turned on. For ease the testing add keyword which will perform elements blocking based on the locator.

aaltat:
Keyword should block all elements.

aaltat:
Done in #1239

PR #1241, opened 2018-11-23
Cover Element should cover all matching elements

SergiuTudos:
For specific element we'll need to use specific locator


atest/resources/html/links.html
```diff
@@ -10,6 +10,7 @@
       <a href="http://www.google.com">Absolute external link</a><br/>
       <a id="some_id" href="broken.html">Link with id</a><br/>
       <a href="index.html"><img src="image.jpg" id="image_id"/></a><br/>
+      <a href="index.html"><img src="image.jpg" id="image_id"/></a><br/>
```
aaltat: Having same id is not allowed by some browsers. Please change the id.

SergiuTudos: Updated and ready to merge

aaltat:
I have few question
1) If element is not found keyword does not fail, but should it fail? I have leaning towards that it should fail, but I do not feel strongly on the matter. But regardless which way the keyword will work, situation must have a test. 
2) Should keyword return count of how many elements it did cover? Feel that it could useful feature to check that correct elements are covered.

SergiuTudos:
I added the case when element is not found
For returning elements count I think we're going to do over-engineering to the keyword. Who needs this can use get elements count.

SergiuTudos:
can someone rerun tests?

HelioGuilherme66:
@SergiuTudos 
Done.
Green again!

SergiuTudos:
Thank you !!!
Waiting for sensei to merge code

aaltat:
Fixes #1239

PR #1242, opened 2018-11-25
Press keys support

### LINKED ISSUE #498: New `Press Keys` keyword with clean support for special keys
The current `Press Key` keyword has a severe limitation that pressing special keys like tab requires using escaped ASCII codes like `\\09` (double escaping needed because `\` itself is an escape character in Robot's data) and not all special keys (e.g. arrows) are supported. 

Issue #275 proposed adding support for WebDriver's [special keys](http://selenium.googlecode.com/git/docs/api/py/webdriver/selenium.webdriver.common.keys.html) like `ARROR_UP` to `Press Key` and it was implemented by PR #474 in version 1.7.3. Unfortunately this change broke undocumented (and tested) feature in`Press Key` to allow pressing multiple keys (#489). Such a backwards incompatible change was not considered a good idea, especially in a minor release, and changes were reverted in 1.7.4.

The general consensus in the comments of the aforementioned issues and pull requests  tracker and on the Slack channel seems to be the following:
- Adding support for named special keys is an important enhancement.
- It is a good idea to support pressing multiple keys (and `Press Key` that actually supports it is badly named).
- `Press Key` behavior should not be changed in backwards incompatible manner.

As a result the consensus seems to be that we want to add new `Press Keys` keyword that supports named keys. There is no consensus about lower level design decisions, though, but we can discuss about them in the comments of this issue.

### LINKED ISSUE #1250: New `Press Keys` keyword with clean support for special keys
The current Press Key keyword has a severe limitation that pressing special keys like tab requires using escaped ASCII codes like \\09 (double escaping needed because \ itself is an escape character in Robot's data) and not all special keys (e.g. arrows) are supported.

Issue #275 proposed adding support for WebDriver's special keys like ARROR_UP to Press Key and it was implemented by PR #474 in version 1.7.3. Unfortunately this change broke undocumented (and tested) feature in `Press Key` to allow pressing multiple keys (#489). Such a backwards incompatible change was not considered a good idea, especially in a minor release, and changes were reverted in 1.7.4. 

The general consensus in the comments of the aforementioned issues and pull requests tracker and on the Slack channel seems to be the following:

* Normal keys should be supported directly. Examples: f (one key), FOO (three keys).

* Escaped ASCII codes (e.g. \\09) supported by Press Key is not supported. They are hard to read, hard to use (double escaping), add complexity to code, and simply have no benefits over named keys.

* WebDriver's named [special keys](https://seleniumhq.github.io/selenium/docs/api/py/webdriver/selenium.webdriver.common.keys.html) (e.g. `CONTROL`, `END`) should be supported. They should, however, be supported only when used alone or separated with `+` that is used as a modifier separator. This avoids named keys accidentally matching something like BEND or AALTO and also simplifies implementation because there is no need to scan the whole string for named keys.

* Common aliases should be added to named keys when appropriate. For example, `CTRL` should be an alias for `CONTROL` and `ESC` should be alias for `ESCAPE`.

* It should be possible to pass multiple keys also as separate arguments. This makes it possible to use `| E | N | D |` (| used as argument separator) if you want to press those three keys and not single key 

* If locator is not defined, sends the keys to the currently active browser.

* The separator between modifiers and keys should be `+` with no spaces around. Other separators are not supported.  
```
| `Press Keys` | text_field | AAAAA  |    | # Sends string "AAAAA" to element identified by text_field            |
| `Press Keys` | None       | BBBBB  |    | # Sends string "BBBBB" to currently active browser                    |
| `Press Keys` | text_field | CTRL+c |    | # Pressing CTRL key down, sends string "c" and then releases CTRL key |
| `Press Keys` | text_field | E+N+D  |    | # Sends string "END" to element identified by text_field              |
| `Press Keys` | text_field | XXX    | YY | # Sends strings "XXX" and "YY" to element identified by text_field    |
| `Press Keys` | button     | RETURN |    | # Press enter key to element identified by button                     |
```
This issue replaces issue #498

aaltat:
Fixes #1250


atest/acceptance/keywords/press_keys.robot
```diff
@@ -0,0 +1,45 @@
+*** Settings ***
+Test Setup        Go To Page "forms/input_special_keys.html"
+# Test Teardown     Test CleanUp
+Resource          ../resource.robot
+Force Tags       tidii
```
aaltat: Remove tag

atest/acceptance/keywords/press_keys.robot
```diff
@@ -0,0 +1,45 @@
+*** Settings ***
+Test Setup        Go To Page "forms/input_special_keys.html"
+# Test Teardown     Test CleanUp
```
aaltat: Remove comment.

atest/acceptance/keywords/press_keys.robot
```diff
@@ -0,0 +1,45 @@
...
+    Press Keys    text_field    SHIFT+cc    SHIFT+dd
+    Click Button    OK
+    Wait Until Page Contains    CCDD     timeout=3
+
+Press Keys Special Keys SHIFT Many Times
+    Press Keys    None    SHIFT+cc    SHIFT+dd
+    Click Button    OK
+    Wait Until Page Contains    CCDD     timeout=3
+
+*** Keywords ***
```
aaltat: Not used, remove.

src/SeleniumLibrary/keywords/element.py
```diff
@@ -741,6 +740,24 @@ def press_key(self, locator, key):
         element = self.find_element(locator)
         element.send_keys(key)
 
+    @keyword
+    def press_keys(self, locator=None, *keys):
+        """Write doc"""
```
aaltat: Write proper documentation.

src/SeleniumLibrary/keywords/element.py
```diff
@@ -741,6 +740,24 @@ def press_key(self, locator, key):
...
+        if is_truthy(locator):
+            element = self.find_element(locator, required=False)
+        for parsed_keys in self._parse_keys(*keys):
```
aaltat: Add support for also not using `element`.

aaltat: Done

src/SeleniumLibrary/keywords/element.py
aaltat: Done

atest/acceptance/keywords/press_keys.robot
aaltat: Done

atest/acceptance/keywords/press_keys.robot
aaltat: Done

atest/acceptance/keywords/press_keys.robot
aaltat: Done

src/SeleniumLibrary/keywords/element.py
```diff
@@ -723,24 +725,91 @@ def simulate(self, locator, event):
...
             key = self._map_ascii_key_code_to_key(int(key[1:]))
         element = self.find_element(locator)
         element.send_keys(key)
 
+    @keyword
+    def press_keys(self, locator=None, *keys):
+        """Simulates user pressing key(s) to an element or on the active browser.
+
+
+        If ``locator`` is None or false, then the ``keys`` are sent to the currently
```
aaltat: If locator evaluates as false, see `Boolean arguments`  for more details, then keys are send to currently active browser. Otherwise element is searched and keys are send to the element identified by the locator. In later case keyword fails if element is not found. See the locating... This version could be better

aaltat (commented):
Thanks for the input. I need read and think about the documentation more. There are things that needs documentation, like example the points that you gave in the issue.

aaltat:
Write test for sending single +  character and when multiple  + characters. Also document what happens to C++ string.

Issue #1199, opened 2018-09-14
Ability to set default log level for `Page Should Contain` keywords


joaonc:
All the [Page Should Contain](http://robotframework.org/SeleniumLibrary/SeleniumLibrary.html#Page%20Should%20Contain) keywords (and `Frame Should Contain`, etc.) have a default loglevel of `INFO`.

This causes the logs at `INFO` level to be *huge*, especially if you use the `Wait Until` variants, which log the whole html each time it's checked.

That means that:
1. I live with 5+ MB logs per test case fail (imagine when several tests fail)
or
2. Have to explicitly set the `loglevel` argument each time these keywords are used.

Another option would be to extend SeleniumLibrary and change that default, but I would really like to get away from that.

The request is to add the ability to set the loglevel for this family of keywords.

joaonc:
FYI, we keep logs in a small AWS EC2 instance, so space is not unlimited..  we also copy some to S3 and email them around sometimes as well.

We _really_ prefer the logs to be at `INFO` level, but having the whole DOM being spit out in these keywords makes that impractical, which means the rest of the important information at `INFO` level goes away.

Hope this makes a good case for adding this feature, tx!

aaltat:
So your problem is that if `Page Should Contain` and it's variants fails, it logs the HTML source. This is a problem for you, because it makes the log size big. Am I correct?

If I am correct, there is a similar issue open for table keywords #991 Now days I tend to agree that logging source is not useful. Therefore it would be useful to remove the log source functionality also from these keywords.

Would you be willing to provide PR to change the functionality?

joaonc:
I'll try and give it a shot.

What's your suggestion?
1. Set the default loglevel in all keywords to `NONE` or `DEBUG`
1. Remove logging altogether
1. Somehow add the ability to set the default loglevel once and have that apply to all keywords that dump source? (with a default of `NONE` if no new default is set).

pekkaklarck:
I agree logging the source is not very useful these days. An easy solution would be removing such logging altogether or changing it to use trace level. I wouldn't use debug because that is often a good level to use when running tests and we shouldn't flood the log in that case. On trace level log files easily grow so much anyway that adding some more possibly useful content there isn't an issue.

[...]

PR #1244, opened 2018-11-27
Set log source default level to TRACE.

joaonc:
Addresses issue #1199 

Having the whole DOM in these keywords at the INFO level is too much clutter and causes the size of the logs to increase dramatically.

### DIFF
--- a/src/SeleniumLibrary/keywords/browsermanagement.py
+++ b/src/SeleniumLibrary/keywords/browsermanagement.py
@@ -296,8 +296,8 @@ class BrowserManagementKeywords(LibraryComponent):
         """Logs and returns the HTML source of the current page or frame.
 
         The ``loglevel`` argument defines the used log level. Valid log
-        levels are ``WARN``, ``INFO`` (default), ``DEBUG``, and ``NONE``
-        (no logging).
+        levels are ``WARN``, ``INFO`` (default), ``DEBUG``, ``TRACE``
+        and ``NONE`` (no logging).
         """
         source = self.get_source()
         self.log(source, loglevel)
--- a/src/SeleniumLibrary/keywords/browsermanagement.py
+++ b/src/SeleniumLibrary/keywords/browsermanagement.py
@@ -292,7 +292,7 @@ class BrowserManagementKeywords(LibraryComponent):
         return url
 
     @keyword
-    def log_source(self, loglevel='INFO'):
+    def log_source(self, loglevel='TRACE'):
         """Logs and returns the HTML source of the current page or frame.
 
         The ``loglevel`` argument defines the used log level. Valid log
--- a/src/SeleniumLibrary/keywords/browsermanagement.py
+++ b/src/SeleniumLibrary/keywords/browsermanagement.py
@@ -292,7 +292,7 @@ class BrowserManagementKeywords(LibraryComponent):
         return url
 
     @keyword
-    def log_source(self, loglevel='TRACE'):
+    def log_source(self, loglevel='INFO'):
         """Logs and returns the HTML source of the current page or frame.
 
         The ``loglevel`` argument defines the used log level. Valid log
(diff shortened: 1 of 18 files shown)


joaonc:
@aaltat / @pekkaklarck The tests in this PR are failing ([here](https://travis-ci.org/robotframework/SeleniumLibrary/builds/460026885)), but the failures seem unrelated to the PR.. maybe the tests need to be updated?

aaltat:
SeleniumLibrary uses also https://github.com/robotframework/statuschecker/ to verify log messages and keyword status from the output.xml. Also some test may initially fail, but are marked as passed with the status checker and vice versa. Based on the Travis it looks like a some of log check fails. Run the tests with the acceptance test runner: ` python atest/run.py <browser>` and fix the tests.

aaltat:
Noticed few bugs in the https://github.com/robotframework/SeleniumLibrary/blob/master/atest/README.rst but now it should be fixed.

joaonc:
Trying to verify that the DOM is _not_ added to the log at `INFO` level, but bumping into [this issue](https://github.com/robotframework/statuschecker/issues/4) with `robotstatuschecker`.

Can you think of another way I can check that the DOM is not **logged** in that step?

The other way I can think is to simply remove the check, but I'd rather check that is not there than not check at all.

aaltat:
There is suggestion, in the https://github.com/robotframework/statuschecker/issues/4 how to overcome the problem. Would you need more help?

joaonc:
Need some help.  The new test cases are passing, except [this one](https://github.com/robotframework/SeleniumLibrary/blob/5b96c804bca76dc8cf235ecac896f8c1b8e3f75e/atest/acceptance/keywords/content_assertions.robot#L70) (`Page Should Contain With Custom Log Level TRACE`) b/c loglevel TRACE is not working.

Even if I just do `Log    Whatever    TRACE` it won't log anything.

Any idea what's going on?
Followed the code and seemed ok.. TRACE should work.
Note that the previous test case (same but at DEBUG level) works as expected.

pekkaklarck:
Without looking anything, my guess is that tests are run using the debug level and thus trace messages are ignored. You can use `Set Log Level` keyword to set it to trace in setup and again back to debug in teardown.

joaonc:
Yes, that was it tx!
Tests are now passing except one job [here](https://travis-ci.org/robotframework/SeleniumLibrary/builds/461540244?utm_source=github_status&utm_medium=notification).  It's weird b/c I run w/ Python 3.6 on my local and works fine.

I wonder if it's b/c it's running on [Jython](https://travis-ci.org/robotframework/SeleniumLibrary/jobs/461540248#L473) in the test case [Create Webdriver Creates Functioning WebDriver](https://travis-ci.org/robotframework/SeleniumLibrary/jobs/461540248#L663), which should pass and doesn't seem related to my changes.

aaltat:
Jython fails randomly in Travis, it is annoying but I have not been able to reproduce the problem locally. I did re-trigger the Jython run and lets see how it goes.

joaonc:
All is green now, thanks!  Hope this gets merged for the next SeleniumLibrary release.

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -292,12 +292,12 @@ def log_location(self):
         return url
 
     @keyword
-    def log_source(self, loglevel='INFO'):
+    def log_source(self, loglevel='TRACE'):
```
aaltat: Log source `loglevel` should not be changed.

aaltat (changes requested):
It looks like you changed all the keywords log level. At least the `Log Source` log level should not have been changed. Also I did not prepare this big change and I will think it over.

joaonc:
@aaltat made the fix you requested and now the Jython test is failing again..  can you kick it off again?


> did not prepare this big change and I will think it over

Note that it's the same change everywhere and works as discussed in issue #1199 
Also, added test cases to make sure the new logging level is as expected.

Sure hope this makes it..  the amount of unnecessary logging at `INFO` level is causing me to have to log at `ERROR` (only place where the DOM is not logged), which in turn causes me to miss out on important info and have to re-run tests when they fail  😞

src/SeleniumLibrary/keywords/browsermanagement.py
joaonc: Change added.

aaltat (approved):

aaltat:
I think this is good to go. I need to create new issue, to that it matches to the scope of the PR.

PR #1263, opened 2018-12-12
Get Cookies keyword able to return dict now

### LINKED ISSUE #979: `Get Cookies` should support returning cookies as objects and have better logging
Currently `Get Cookies` returns a cookies as a string in format `cookie1=value1; cookie2=value2`. This isn't very useful if you need to actually handle all cookies somehow, and the return format isn't even documented anywhere. We should change the keyword to return cookies as a list (or ordered dictionary) of CookieInformation objects that were introduced as part of #932.

Changing keyword return value is obviously backwards incompatible, so we cannot do it without a deprecation period. I propose that in v3.1 we add an argument like `return_objects=False` that can be used to explicitly change the return value. This allows users to opt-in if they want to get objects. In v3.2 we could then change the argument to `return_objects=True`, making it possible to opt-out if needed. And in v3.3 or later the whole argument can be removed. This issue only covers adding `return_objects=False` (or similar), changing and removing it later require separate issues later.

Another problem in `Get Cookies` is that it doesn't log anything. The beginning of the return value is logged automatically if the return value is assigned to a variable, but it would be better to log at least names and values explicitly.

---

**UPDATE:** @Mohorelien commented below that there actually is a use case for the current format in HTTP request headers. I still think having a way to return cookie information as objects is useful, but we should preserver the current functionality too. I'd also keep the current functionality as the default functionality to avoid all backwards incompatible changes. That way this change could even be done in SL 3.0 if we have time.

rasjani:
This is usefull as Requests Library's Create Session keywords expects
cookies parameter to be a dict, not a string that Get Cookies returns by
default.

Fixes #979

### DIFF
--- a/atest/acceptance/keywords/cookies.robot
+++ b/atest/acceptance/keywords/cookies.robot
@@ -11,6 +11,11 @@ Get Cookies
     Should Match Regexp    ${cookies}
     ...    ^(test=seleniumlibrary; another=value)|(another=value; test=seleniumlibrary)$
 
+Get Cookies As Dict
+    ${cookies}=    Get Cookies        as_dict=True
+    ${expected_cookies}=    Create Dictionary   test=seleniumlibrary    another=value
+    Dictionaries Should Be Equal  ${expected_cookies}   ${cookies}
+
 Get Cookie Value Set By Selenium
     ${value} =    Get Cookie Value    another
     Should Be Equal    ${value}       value
@@ -65,6 +70,13 @@ Get Cookies When There Are None
     Delete All Cookies
     ${cookies} =    Get Cookies
     Should Be Equal    ${cookies}    ${EMPTY}
+
+Get Cookies As Dict When There Are None
+    [Tags]    Known Issue Safari
+    Delete All Cookies
+    ${cookies} =    Get Cookies   as_dict=True
+    ${expected_cookies}=    Create Dictionary
+    Dictionaries Should Be Equal  ${expected_cookies}   ${cookies}
 
 Test Get Cookie Object Expiry
     ${cookie} =    Get Cookie      another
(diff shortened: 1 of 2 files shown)


src/SeleniumLibrary/keywords/cookie.py
```diff
@@ -39,18 +39,29 @@ def delete_cookie(self, name):
...
+        if not as_dict:
+            pairs = []
+            for cookie in self.driver.get_cookies():
+                pairs.append(cookie['name'] + "=" + cookie['value'])
+            return '; '.join(pairs)
+        else:
+            pairs = {}
+            for cookie in self.driver.get_cookies():
+                pairs[cookie['name']] = cookie['value']
+            return pairs
```
aaltat, line 67: Instead of returning dictionary, it might be better to return RF dotdict https://github.com/robotframework/robotframework/blob/2.9.2/src/robot/utils/dotdict.py

atest/acceptance/keywords/cookies.robot
```diff
@@ -1,4 +1,5 @@
 *** Setting ***
+Library           BuiltIn
```
aaltat: No need to import BuildIn

src/SeleniumLibrary/keywords/cookie.py
```diff
@@ -39,18 +39,29 @@ def delete_cookie(self, name):
...
+        ``as_dict`` to True is helpful as then the result can be passed
+        to requests library's Create Session keyword's optional cookies
+        parameter.
+
         """
-        pairs = []
-        for cookie in self.driver.get_cookies():
-            pairs.append(cookie['name'] + "=" + cookie['value'])
-        return '; '.join(pairs)
+        if not as_dict:
```
aaltat: Please use is_falsy https://github.com/robotframework/SeleniumLibrary/blob/master/src/SeleniumLibrary/utils/types.py

aaltat:
I updated the documentation, when Travis shows green (assuming that I did not make mistake), this is ready to be merged.

PR #1296, opened 2019-01-19
Allows adding external class to extend SeleniumLibrary

### LINKED ISSUE #1292: Allow extending SeleniumLibrary by creating plugin API.
It's already possible to extend SeleniumLibrary from other library or listener by adding new classes. For more details, see https://github.com/robotframework/SeleniumLibrary/blob/master/docs/extending/examples/modify_seleniumlibrary/NewKeywords.py Although this is useful, it can not be done from Robot Framework test data. Therefore, example IDE can not know that new keywords are added and existing keywords are modified. Because of this limitation, it would be useful to allow the same thing also happen in library import and perhaps also with keyword. Example something like this:
```
| Library | SeleniumLibrary | extension=MyClass,OtherClass |
```
In the above example `,` is separator for the classes.

Also arguments should be supported:
```
| Library | SeleniumLibrary | extension=MyClass;arg1;arg2 |
```
In the above example `;` is separator for the arguments. Arguments will be passed as strings and conversion to different types will not be done by the SeleniumLibrary. Also support named arguments and variable number of arguments is useful. 

 Example if plugin looks like this:
```
class PluginWithAllArgs(LibraryComponent):

         def __init__(self, ctx, arg, *varargs, **kwargs):
```
Then import could look like this:
``` 
| Library | SeleniumLibrary | extension=PluginWithAllArgs.py;argument1;varg1;varg2;kw1=kwarg1;kw2=kwarg2
```

It should be allowed to import class from module search path or with file name.

It must be only allowed to import class, adding methods from modules are not supported. In short the class should look similar to classes which are implementing SeleniumLibrary keywords.

aaltat:
Fixes #1292


aaltat:
The code is ready, from the `__init__` point of view. But documentation is missing, which should tell, what rules there are for the class which extends SeleniumLibrary. Also there should be tests for different kind of error situations.

aaltat:
Few items to do:

- [x] Write test which overwrites the `ElementFinder`.
- [x] Document how plugin supports parameters.
- [x] Document failures in importing plugin.
- [x] Document more about plugin API.
- [x] Document the LibraryComponent.
- [x] Document creating new keywords.
- [x] Document overwriting existing keywords and how to use SL keyword from your new KW.
- [x] Document if two plugin defines same keyword.
- [x] Document what `ctx` actually is. 
- [x] Try libdoc with plugin. Keyword doc should be produced with plugin kw included. 
- [x] Find easy way to plugins provide keyword documentation in similar way as with libraries. 
- [x] Write test, which should fail, if plugin doesn't contain a class.

aaltat:
Although `libdoc` works, it can not be used. `libdoc` exposes too many keywords, because `libdoc` thinks that library is using static interface. Need to find better way to do it.

aaltat:
Decide should we actually fail, if plugin does not inherit [LibraryComponent](https://github.com/robotframework/SeleniumLibrary/blob/master/src/SeleniumLibrary/base/librarycomponent.py#L28)

aaltat:
To do: 

- [x] Think more about how to generate plugin documentation
- [x]  Decide should we fail if plugin does not inherit [LibraryComponent](https://github.com/robotframework/SeleniumLibrary/blob/master/src/SeleniumLibrary/base/librarycomponent.py#L28)

There is [gist](https://gist.github.com/aaltat/64dffeb571fd2081f6385b32a51b8d07)  which provides few solutions how the `libdoc` could be used to generate documentation. But I am not happy with those solution, need to think more about the problem tough.

aaltat:
Another good idea from Pekka was that generate keyword documentation from the SeleniumLibrary with plugins and recommend to add [keywords tags](http://robotframework.org/robotframework/latest/RobotFrameworkUserGuide.html#keyword-tags) to the plugin keywords. 

I think generating whole library documentation is good idea. Just need to explain it and add tags to all plugins keywords in tests.

pekkaklarck:
SL could actually automatically add `plugin` tag for all keywords coming from plugins. Shouldn't be too complicated. Either it could modify `robot_tags` attribute of all plugin methods or it could override `get_keyword_tags` from the dynamic API and the tag for plugin kws. The latter is probably easier.

aaltat:
Adding `plugin` tag to all keywords coming from plugins was easy to do. Need to document that tag is added and write some instructions how to generate keywords documentation.

PR #1310, opened 2019-02-14
Allow extra arguments for CookieInformation

### LINKED ISSUE #1307: Get Cookies keyword will fail if the Selenium get_cookie method return value contains more keys than: name, value, path, domain, secure, httpOnly and expiry
Selenium [get_cookie](https://seleniumhq.github.io/selenium/docs/api/py/webdriver_remote/selenium.webdriver.remote.webdriver.html?highlight=get_cookie#selenium.webdriver.remote.webdriver.WebDriver.get_cookie) returns a dictionary. If the dictionary contains more keys then the `name`, `value`, `path`, `domain`, `secure`, `httpOnly` and `expiry`, then the Get Cookies keyword will fail:
```
20190212 16:54:43.113 :  FAIL : TypeError: __init__() got an unexpected keyword argument 'hCode'
20190212 16:54:43.113 : DEBUG :
  File "C:\Python27\lib\site-packages\SeleniumLibrary\__init__.py", line 369, in run_keyword
    return DynamicCore.run_keyword(self, name, args, kwargs)
  File "C:\Python27\lib\site-packages\SeleniumLibrary\base\robotlibcore.py", line 102, in run_keyword
    return self.keywords[name](*args, **kwargs)
  File "C:\Python27\lib\site-packages\SeleniumLibrary\keywords\cookie.py", line 99, in get_cookie
    return CookieInformation(**cookie)
```
This is because the Get Cookies keywords has hard coded keys what Selenium response [can contain in the CookieInformation](https://github.com/robotframework/SeleniumLibrary/blob/master/src/SeleniumLibrary/keywords/cookie.py#L152) object. The limitation has been in place since the SeleniumLibrary 3.0 release, but there was issue reported in Slack, that for that particular user, Selenium `get_cookie` did return the following:
```
{u'domain': u'1.2.3.4', u'name': u'components_state', u'value': u'alsdjaslkdjasdlkj', u'path': u'/', u'hCode': -1693023704, u'class': u'org.openqa.selenium.Cookie', u'httpOnly': False, u'secure': False}
```
SeleniumLibrary should not fail, if the Selenium `get_cookie`  returns more keys that [is specified ](https://w3c.github.io/webdriver/#dfn-table-for-cookie-conversion). Instead we should allow extra keys to be passed to the `CookieInformation` object and allow access to the extra keys in similar way as the specified keys.

aaltat:
Fixes #1307


aaltat:
Not good enough, order of keys will be random and causes inconsistency.

pekkaklarck:
I'd just use

     self.extra = extra

instead of adding separate attributes for all this non-standard info. Not sure should it even be listed in `__str__` either. Main problem at the moment is that this extra stuff breaks getting cookies. Nobody has told validating or even seeing this info is important.

aaltat:
Good point and it sounds like a better idea. At least it's easier to and needs some extra documentation.

aaltat:
Didn't mean to close the PR.

src/SeleniumLibrary/keywords/cookie.py
```diff
@@ -152,16 +161,21 @@ def _expiry(self, expiry):
...
+                 httpOnly=False, expiry=None, **extra):
         self.name = name
         self.value = value
         self.path = path
         self.domain = domain
         self.secure = secure
         self.httpOnly = httpOnly
         self.expiry = datetime.fromtimestamp(expiry) if expiry else None
+        if extra:
+            self.extra = extra
```
pekkaklarck: Conditionally setting or not setting attribute is a bit of an anti-pattern. Is there some benefit here?

aaltat: That is true. Although I did explain it in the docs. Will change.

pekkaklarck (approved):

PR #1314, opened 2019-02-28
Get and Set Inner Window Size

ciccioman3:
This keywords allows users to define/receve the window sizing by the desired inner size.
This is mission-critical when you need a specific size (for layout breakpoints or more)

### DIFF
--- a/src/SeleniumLibrary/keywords/window.py
+++ b/src/SeleniumLibrary/keywords/window.py
@@ -161,19 +161,29 @@ class WindowKeywords(LibraryComponent):
         self.driver.maximize_window()
 
     @keyword
-    def get_window_size(self):
+    def get_window_size(self, inner=False):
         """Returns current window width and height as integers.
 
         See also `Set Window Size`.
 
+        With optional `inner` parameter you receive
+        the inner size, excluding the browser bars, borders and so on.
+
         Example:
         | ${width} | ${height}= | `Get Window Size` |
+        | ${width} | ${height}= | `Get Window Size` | True |
         """
-        size = self.driver.get_window_size()
-        return size['width'], size['height']
+        if inner == True:
+            inner_width = int(self.driver.execute_script("window.innerWidth"))
+            inner_height = int(self.driver.execute_script("window.innerHeight"))
+
+            return inner_width, inner_height
+        else:
+            size = self.driver.get_window_size()
+            return size['width'], size['height']
 
     @keyword
-    def set_window_size(self, width, height):
+    def set_window_size(self, width, height, inner=False):
         """Sets current windows size to given ``width`` and ``height``.
 
         Values can be given using strings containing numbers or by using
@@ -183,55 +193,24 @@ class WindowKeywords(LibraryComponent):
         smaller will cause the actual size to be bigger than the requested
         size.
 
-        Example:
-        | `Set Window Size` | 800 | 600 |
-        """
-        return self.driver.set_window_size(int(width), int(height))
-
-    @keyword
-    def get_inner_window_size(self):
-        """Returns current inner window width and height as integers.
-
-        See also `Set Inner Window Size`.
-
-        Example:
-        | ${width} | ${height}= | `Get Inner Window Size` |
-
-        The main difference with `Get Window Size` is that you receive
-        the inner size, excluding the browser bars, borders and so on.
-        """
-        inner_width = int(self.driver.execute_script("window.innerWidth"))
-        inner_height = int(self.driver.execute_script("window.innerHeight"))
-
-        return inner_width, inner_height
-
-    @keyword
-    def set_inner_window_size(self, width, height):
-        """Sets current windows size adapted to contain page
-        with given ``width`` and ``height``.
-
-        The main difference with `Set Window Size` is that you can give
+        With optional `inner` parameter you can give
         the inner size required, excluding the browser bars, borders and so on.
         The window size is adapted to provide the correct page size for each browser.
 
-        Values can be given using strings containing numbers or by using
-        actual numbers. See also `Get Window Size` and `Set Window Size` .
-
-        Browsers have a limit how small they can be set. Trying to set them
-        smaller will cause the actual size to be bigger than the requested
-        size.
 
         Example:
-        | `Set Inner Window Size` | 800 | 600 |
+        | `Set Window Size` | 800 | 600 |
+        | `Set Window Size` | 800 | 600 | True |
         """
-        self.driver.set_window_size(int(width), int(height))
-        inner_width = int(self.driver.execute_script("window.innerWidth"))
-        inner_height = int(self.driver.execute_script("window.innerHeight"))
-        width_offset = width - inner_width
-        height_offset = height - inner_height
-        window_width = width + width_offset
-        window_height = height + height_offset
-        return self.driver.set_window_size(int(window_width), int(window_height))
+        if inner == True:
+            self.driver.set_window_size(int(width), int(height))
+            inner_width = int(self.driver.execute_script("window.innerWidth"))
+            inner_height = int(self.driver.execute_script("window.innerHeight"))
+            width_offset = width - inner_width
+            height_offset = height - inner_height
+            return self.driver.set_window_size(int(width + width_offset), int(height + height_offset))
+        else:
+            return self.driver.set_window_size(int(width), int(height))
 
     @keyword
     def get_window_position(self):
--- a/src/SeleniumLibrary/keywords/window.py
+++ b/src/SeleniumLibrary/keywords/window.py
@@ -169,10 +169,10 @@ class WindowKeywords(LibraryComponent):
         If ``inner`` parameter is set to True, keyword returns
         HTML DOM window.innerWidth and window.innerHeight properties.
         See `Boolean arguments` for more details how to set boolean
-        arguments.
+        arguments. The ``inner`` is new in SeleniumLibrary 4.0.
 
         Example:
-        | ${width} | ${height}= | `Get Window Size` |
+        | ${width} | ${height}= | `Get Window Size` |      |
         | ${width} | ${height}= | `Get Window Size` | True |
         """
         if is_truthy(inner):
@@ -195,15 +195,15 @@ class WindowKeywords(LibraryComponent):
 
         If ``inner`` parameter is set to True, keyword sets the necessary
         window width and height to have the desired HTML DOM window.innerWidth
-        and window.innerHeight
+        and window.innerHeight The ``inner`` is new in SeleniumLibrary 4.0.
         See `Boolean arguments` for more details how to set boolean
         arguments.
 
-        This keyword does NOT support Frames. If a frame is selected,
+        This ``inner`` argument does not support Frames. If a frame is selected,
         switch to default before running this.
 
         Example:
-        | `Set Window Size` | 800 | 600 |
+        | `Set Window Size` | 800 | 600 |      |
         | `Set Window Size` | 800 | 600 | True |
         """
         width, height = int(width), int(height)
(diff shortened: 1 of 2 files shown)


aaltat:
I did read what the `innerHeight` and `innerWidth` properties are and I understand what you trying to do with the keywords. What I do not understand is the need for such keyword, could you describe your need in issue?

ciccioman3:
@aaltat For many tests I need to know the size of my web page. For example if you need to test pages with different layout according to page size breakpoints you need to know the real page size. With "Set window size" i can set 800px, but the browser window will be 800, NOT the page. This happen because browsers have borders and bars included in the window size. For this reason my tests are going to fail because my web page is going to be rendered in a smaller space. If I need to test a page behaviour at 800px I need 800px of page + the browser extra space. This happen for both height and width but usually browsers have only few pixels of left and right borders.

innerwindow

aaltat:
So you want to test when `innerHeight` and `innerWidth` have certain dimension, then you webapp should do something (this case the layout) differently? In that case, I see your need. 

I have never need or heard of such needs. But in any case, let me think few days, is this good feature to have in the library.

aaltat:
Sorry that it took this long to respond.

I think you have a valid need and this is good feature to have. But because SeleniumLibrary keyword number is actually quite big, having new keywords for slightly different size is not good idea. Instead the existing get/set window could be enhanced to take in an optional `inner` argument and modify the keyword functionality to match this PR proposal. Would you be willing to modify your PR?

ciccioman3:
Yes, it sounds good! I'll modify the PR as soon as I can.

aaltat:
It seems that acceptance tests fails, could you take a look why it happens?

aaltat:
Looking pretty good, there are few things that need enhancement but I will get back when I can do review from PC and not from mobile.

src/SeleniumLibrary/keywords/window.py
```diff
@@ -161,19 +161,29 @@ def maximize_browser_window(self):
...
+        With optional `inner` parameter you receive
+        the inner size, excluding the browser bars, borders and so on.
+
         Example:
         | ${width} | ${height}= | `Get Window Size` |
+        | ${width} | ${height}= | `Get Window Size` | True |
         """
-        size = self.driver.get_window_size()
-        return size['width'], size['height']
+        if inner == True:
```
aaltat: 1) I did forgot to say, but SeleniumLibrary has own boolean handling. In this case, [is_falsy](https://github.com/robotframework/SeleniumLibrary/blob/master/src/SeleniumLibrary/utils/types.py#L32) should be used. 
2) It would be more Pythonic way to write: 
`if is_falsy(inner):`

src/SeleniumLibrary/keywords/window.py
```diff
@@ -161,19 +161,29 @@ def maximize_browser_window(self):
...
+            inner_width = int(self.driver.execute_script("return window.innerWidth;"))
+            inner_height = int(self.driver.execute_script("return window.innerHeight;"))
+
```
aaltat: Empty line is not needed here.

src/SeleniumLibrary/keywords/window.py
```diff
@@ -161,19 +161,29 @@ def maximize_browser_window(self):
...
+            return inner_width, inner_height
+        else:
```
aaltat: Else is not needed, instead this would be cleaner: 
```
    return inner_width, inner_height
size = self.driver.get_window_size()
return size['width'], size['height'
````

File: src/SeleniumLibrary/keywords/window.py [hunk 4]
```diff
@@ -183,10 +193,26 @@ def set_window_size(self, width, height):
         smaller will cause the actual size to be bigger than the requested
         size.
 
+        With optional `inner` parameter you can give
+        the inner size required, excluding the browser bars, borders and so on.
+        The window size is adapted to provide the correct page size for each browser.
+
+
```
aaltat, line 204: One empty line is enough, please remove the one extra empty line.

src/SeleniumLibrary/keywords/window.py
```diff
@@ -183,10 +193,26 @@ def set_window_size(self, width, height):
...
         Example:
         | `Set Window Size` | 800 | 600 |
+        | `Set Window Size` | 800 | 600 | True |
         """
-        return self.driver.set_window_size(int(width), int(height))
+        if inner == True:
```
aaltat: 1) [is_falsy](https://github.com/robotframework/SeleniumLibrary/blob/master/src/SeleniumLibrary/utils/types.py#L32) should be used. 
2) It would be better to start to convert the `width`, `height` as int in the beginning. 
```
width, height = int(width), int(height)
if is_falsy(inner):
```
And change `self.driver.set_window_size(int(width), int(height))` to `self.driver.set_window_size(width, height)` in lines 206 and 215.

src/SeleniumLibrary/keywords/window.py
```diff
@@ -183,10 +193,26 @@ def set_window_size(self, width, height):
...
+            self.driver.set_window_size(int(width), int(height))
+            inner_width = self.driver.execute_script("return window.innerWidth;")
+            inner_height = self.driver.execute_script("return window.innerHeight;")
+            width_offset = int(width) - int(inner_width)
```
aaltat: It could be better to do: 
```
inner_height, inner_width = int(inner_height, int(inner_width)
```
after this line. Then rest of the `int` conversions should not be needed.

src/SeleniumLibrary/keywords/window.py
```diff
@@ -161,19 +161,29 @@ def maximize_browser_window(self):
         self.driver.maximize_window()
 
     @keyword
-    def get_window_size(self):
+    def get_window_size(self, inner=False):
         """Returns current window width and height as integers.
 
         See also `Set Window Size`.
 
+        With optional `inner` parameter you receive
```
aaltat: If \`\`inner\`\` the parameter is set to true keyword returns 
CSS window.innerWidth and window.innerHeight properties. 
See `Boolean arguments` for more details how to set boolean
arguments.

src/SeleniumLibrary/keywords/window.py
```diff
@@ -161,19 +161,29 @@ def maximize_browser_window(self):
...
+            inner_height = int(self.driver.execute_script("return window.innerHeight;"))
+
+            return inner_width, inner_height
+        else:
+            size = self.driver.get_window_size()
+            return size['width'], size['height']
 
     @keyword
-    def set_window_size(self, width, height):
+    def set_window_size(self, width, height, inner=False):
```
aaltat, line 186: We need to design how the keyword should work with frames. If frame is selected, `window.innerWidth` and `window.innerHeight` returns the frame size. Logically thinking, it would be best to return to a main/outermost windows/frame (In practice `self.driver.switch_to.default_content()` would be needed.) If we decide to do that, then it is not possible to return to a frame which is was selected, because Selenium or SeleniumLibrary does not track nested frame is selected. But because this is new functionality, when `inner=True` then it is OK to do: `self.driver.switch_to.default_content()`

But what we should do with the `get_window_size` method. I feel that, if change this keyword to do: `self.driver.switch_to.default_content()` then same should be done with `get_window_size` method. 

Also these changes must be documented and tested.

aaltat (changes requested):
Here are my comments, now the question about frames is the most important to solve. Also it is the issue which I have least experience. My gut feeling is that deselecting the frame(s) is the bets choice to do, but I am not sure about the thing. What do you think?

src/SeleniumLibrary/keywords/window.py
ciccioman3: > * It would be more Pythonic way to write:
>   `if is_falsy(inner):`

Why 'is_falsy' in this case, and not 'is_truthy'?

src/SeleniumLibrary/keywords/window.py
ciccioman3: I understand that. I'm a bit unsure about doing the switch, it could be annoying for the user. Leaving the frame sizing could be "reasonable" because of the literally word meaning of the parameter "Inner". But I will think about that a bit more, and I will do some tests with frames.

aaltat: After thinking this bit more, I am starting to think that doing `self.driver.switch_to.default_content()` may not be good idea. It will strip's the possibility from the user to control what frame will be resized. If we just trust the user to select correct frame and resized the selected frame it will bring more control for the users.

In any way we choose, there must be a testswith frames.

humbienri: Are you sure you want to hijack this Keyword and not create another one? Set Window ViewPort or something like that?  I agree that automatically switching the user to the default content would be "rude" but maybe instead force to switch to default content themselves if that is what they mean by wanting to set the window size vs. wanting to set the "web app view size"?  Just some thoughts.

aaltat: The keyword default functionality is not changed. Only if `inner` argument is true, the functionality is different.

aaltat:
It seems that Travis has lost the build for this PR, although it should not. I am going to close PR and reopen it to retrigger the build in Travis.

aaltat:
I did see what the console displayed:
```
FAIL: Acceptance.Windows.Get and Set Inner Window Size with Frames
143 != 400
```
It looks like there is failure in one of the acceptance tests.

ciccioman3:
@aaltat Yes, I see. Well, there is a problem with this whole concept and frames, I guess. This is because frames has their own size behaviour (for example with a certain proportion), and when giving a new window size that does not respect that behaviour is not possible to have a specific frame size using the keyword. We should think if it worths to find a way to make it work (maybe with just one of width/height) or do not support frames at all. For the moment I don't have a clear idea about that, I will think about a bit more. By the way I'm open to any suggestion.

aaltat:
Hmmm, by quickly looking at the problem, it seems that browser does set the frame size automatically. And there might be a way to set it, at least this indicates so: https://www.reddit.com/r/HTML/comments/3g4s5r/how_to_set_iframe_heightwidth_to_fit_browser/ But all that seems complex and not something I would like to support. What you think?

I think it would be easiest to document that sizing frames is not supported. It would be cool if keyword could fail or at least warn users if frame is selected that sizing frames is not supported.

ciccioman3:
Changing the frame size is something out-of-scope, and also something that we shouldn't do because it changes the webpage layout structure. The keyword purpose is changing the window size. This size is calculated in the "inner" case, but without changing in any way the page CSS. Then I think we should not support frames.

aaltat:
Then we agree resizing frames is not supported and for that:
1) Keyword will not change the frame
2) Needs to be documented that resizing of the frame is not supported with `inner` argument as true.
3) There should be tests to document that if frame is selected what happens.
4) Warning would be great, but optional, if `inner` is true and frame has been selected.

aaltat (approved):

aaltat:
Added some documentation stuff and some minor code changes. If Travis displays green, this is ready to be merged. 

Note: Travis is having some difficulties and it might take some time to get the build result.

aaltat:
I did break something and I will fix it later.

src/SeleniumLibrary/keywords/window.py
```diff
@@ -215,12 +215,12 @@ def set_window_size(self, width, height, inner=False):
             height_offset = height - inner_height
             window_width = width + width_offset
             window_height = height + height_offset
-            res = self.driver.set_window_size(window_width, window_height)
+            size = self.driver.set_window_size(window_width, window_height)
```
ciccioman3: If I'm not wrong... this does not contain the size, but just the execution status

src/SeleniumLibrary/keywords/window.py
```diff
@@ -215,12 +215,12 @@ def set_window_size(self, width, height, inner=False):
...
             result_width = int(self.driver.execute_script("return window.innerWidth;"))
             result_height = int(self.driver.execute_script("return window.innerHeight;"))
             if result_width != width or result_height != height:
                 raise AssertionError("Keyword failed setting correct window size.")
-            return res
+            return size['width'], size['height']
```
ciccioman3: This shouldn't return width and height

src/SeleniumLibrary/keywords/window.py
```diff
@@ -215,12 +215,12 @@ def set_window_size(self, width, height, inner=False):
...
         return self.driver.set_window_size(width, height)
```
ciccioman3: As you can see, in the previous keyword return  it was the function execution result status

PR #1320, opened 2019-03-08
Fix 1301 - mention webdrivermanager in the seleniumlibrary docs.

rasjani:
Just skip the update to keyword docs if its not appropriate.

### DIFF
--- a/src/SeleniumLibrary/__init__.py
+++ b/src/SeleniumLibrary/__init__.py
@@ -56,7 +56,9 @@ class SeleniumLibrary(DynamicCore):
 
     SeleniumLibrary uses the Selenium WebDriver modules internally to
     control a web browser. See http://seleniumhq.org for more information
-    about Selenium in general.
+    about Selenium in general and SeleniumLibrary README.rst
+    https://github.com/robotframework/SeleniumLibrary#browser-drivers
+    for more details about webdriver binary instalation.
 
     == Table of contents ==
 
--- a/README.rst
+++ b/README.rst
@@ -93,8 +93,24 @@ New browser driver versions are released to support features in
 new browsers, fix bug, or otherwise, and you need to keep an eye on them
 to know when to update drivers you use.
 
+Alternatively, you can use a tool called WebdriverManager__ which can
+find the latest version or when required, any version of appropriate
+webdrivers for you and then download and link/copy it into right
+location. Tool can run on all major operating systems and supports
+downloading of Chrome, Firefox, Opera & Edge webdrivers.
+
+Here's an example:
+
+.. code:: bash
+
+      pip install webdrivermanager
+      webdrivermanager firefox chrome --linkpath /usr/local/bin
+
+
+
 __ https://seleniumhq.github.io/selenium/docs/api/py/index.html#drivers
 __ https://en.wikipedia.org/wiki/PATH_(variable)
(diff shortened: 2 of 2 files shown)


aaltat:
Looking pretty good from a mobile screen. Need to read from bigger screen too.

src/SeleniumLibrary/__init__.py
```diff
@@ -70,6 +71,24 @@ class SeleniumLibrary(DynamicCore):
     - `Shortcuts`
     - `Keywords`
 
+    = Webdrivers =
+
+    In order for SeleniumLibrary to succesfully interact with browser,
+    a webdriver module needs to be present in the host that is executing the
```
aaltat: Instead of word `module` would a better word be `binary`? Example: `a webdriver binary needs to...` Or `executable` perhaps?

src/SeleniumLibrary/__init__.py
```diff
@@ -70,6 +71,24 @@ class SeleniumLibrary(DynamicCore):
     - `Shortcuts`
     - `Keywords`
 
+    = Webdrivers =
```
aaltat: Would be better to mention need of webdrivers and just link to the README?

src/SeleniumLibrary/__init__.py
```diff
@@ -70,6 +71,24 @@ class SeleniumLibrary(DynamicCore):
...
+
+    In order for SeleniumLibrary to succesfully interact with browser,
+    a webdriver module needs to be present in the host that is executing the
+    tests. Typically this is the host running the robot itself but in case of
+    remote library been used, webdriver must be present there.
+
+    Selenium python library looks for appropriate webdriver from directories
```
aaltat: In RF context:  Selenium python library ->  Selenium tool

src/SeleniumLibrary/__init__.py
```diff
@@ -70,6 +71,24 @@ class SeleniumLibrary(DynamicCore):
     - `Shortcuts`
     - `Keywords`
 
+    = Webdrivers =
+
+    In order for SeleniumLibrary to succesfully interact with browser,
+    a webdriver module needs to be present in the host that is executing the
+    tests. Typically this is the host running the robot itself but in case of
```
aaltat: Would it be more accurate say: 
```
host that is running the browser. Typically this is the host running the robot itself but in case of 
Selenium Grid, the webdriver binary must be present in the Selenium Grid node.
```

aaltat (commented):
I am happy with the README updates. 

But I am not sure do I want to talk in keyword doc about the WebDrivers and least that much. Would it better to keep the first chapter and link to the README?

src/SeleniumLibrary/__init__.py
rasjani: i used the word "module" as it was used in introductory pagagraph: 

```SeleniumLibrary uses the Selenium WebDriver modules internally to control a web browser.``` :D

src/SeleniumLibrary/__init__.py
rasjani: Not really ? You can run robot on host X and  have the SeleniumLibrary loaded as remote library from host Y.  That scenario, webdriver binary needs to be in the Y and has nothing to do with selenium grid.

aaltat:
Thank you for the contribution.

PR #1321, opened 2019-03-08
Support for event firing webDriver

### LINKED ISSUE #1303: Add support event_firing_webdriver
Selenium has an listener API which is called [event_firing_webdriver](https://seleniumhq.github.io/selenium/docs/api/py/webdriver_support/selenium.webdriver.support.event_firing_webdriver.html#module-selenium.webdriver.support.event_firing_webdriver). It allows many interesting features and could be useful example in https://github.com/Omenia/robotframework-seleniumtestability project. 

To implement this, three things are required from the SeleniumLibrary project
- Add argument, in library import, which takes in the `event_listener` class.
- Wrapping driver with [EventFiringWebDriver](https://seleniumhq.github.io/selenium/docs/api/py/webdriver_support/selenium.webdriver.support.event_firing_webdriver.html#selenium.webdriver.support.event_firing_webdriver.EventFiringWebDriver) and `event_listener` class.
- Wrapping WebElements with [EventFiringWebElement](https://seleniumhq.github.io/selenium/docs/api/py/webdriver_support/selenium.webdriver.support.event_firing_webdriver.html#selenium.webdriver.support.event_firing_webdriver.EventFiringWebElement) and `event_listener` class.

When #1292 is done, the functionality to import classes from string should be reused to import the  `event_listener` class. SeleniumLibrary will not perform checks on the `event_listener` class, although it must implement the [Selenium abstract_event_listener](https://seleniumhq.github.io/selenium/docs/api/py/webdriver_support/selenium.webdriver.support.abstract_event_listener.html#module-selenium.webdriver.support.abstract_event_listener). 

In practice import should look like this: 
```
| Library | SeleniumLibrary | event_listener=MySeleniumEventListener |
```
Arguments could be supported, if underlying Selenium allows it, in similar way as in plugins. 

If `event_listener` is not defined or defaults to `False` wrapping drivers and WebElements should not be done.

aaltat:
Fixes #1303

### DIFF
--- a/atest/acceptance/2-event_firing_webdriver/MyListener.py
+++ b/atest/acceptance/2-event_firing_webdriver/MyListener.py
@@ -1,0 +1,23 @@
+from robot.api import logger
+from selenium.webdriver.support.events import AbstractEventListener
+
+
+class MyListener(AbstractEventListener):
+
+    def before_navigate_to(self, url, driver):
+        logger.info("Before navigate to %s" % url)
+
+    def after_navigate_to(self, url, driver):
+        logger.info("After navigate to %s" % url)
+
+    def before_click(self, element, driver):
+        logger.info("Before click")
+
+    def after_click(self, element, driver):
+        logger.info("After click")
+
+    def before_change_value_of(self, element, driver):
+        logger.info("Before send_keys")
+
+    def after_change_value_of(self, element, driver):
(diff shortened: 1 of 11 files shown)


aaltat:
Add also debug message when WebElement us wrapped.

PR #1328, opened 2019-03-12
Test if Xenial resolves pypy3.5

muizzk:


aaltat:
Error looks the same as with trusty and pypy3.5

PR #1342, opened 2019-03-19
Use same webdriver if alias/browser matches

### LINKED ISSUE #1319: It should be possible to configure Open Browser keyword to not always open a new browser.
Many times, when opening browser, there is need to open only one browser. But because depending how test are executed in Robot Framework, it is hard to find a single correct place for the `Open Browser` keyword. Most of the time, it is safer to run the `Open Browser` keyword from multiple levels in the suite structure, which will lead to many open browsers and loss in test execution time. There are also numerous other workarounds available. 

Instead of the workarounds, it would be good to implement support for the `Open Browser` keyword  to open browser only once. If the the `Open Browser` keyword is configured to open browser only one time, it would act in same way as the `Go To` keyword. 

The `alias` argument could act as a marker, should the existing browser be used or new one opened. 

Currently if test does this: 
```
| Open Browser | https://github.com/robotframework/SeleniumLibrary | Chrome | alias |
| Open Browser | https://github.com/robotframework/SeleniumLibrary | Chrome | alias |
```
It will open two browsers with same alias. 

But after the change, this:
```
| Open Browser | https://github.com/robotframework/SeleniumLibrary | Chrome | alias |
| Open Browser | https://github.com/robotframework/SeleniumLibrary | Chrome | alias |
```
will open one browser. With the first `Open Browser` keyword run, it will open the browser normally and navigate to the `https://github.com/robotframework/SeleniumLibrary`. In the second `Open Browser` keyword run, it navigate (again) to the `https://github.com/robotframework/SeleniumLibrary`. 

Also this:
```
| Open Browser | https://github.com/robotframework/SeleniumLibrary | Chrome | alias |
| Open Browser | https://github.com/robotframework/SeleniumLibrary | Firefox | alias |
```
will open only one browser.

If different alias is used: 
```
| Open Browser | https://github.com/robotframework/SeleniumLibrary | Chrome | alias1 |
| Open Browser | https://github.com/robotframework/SeleniumLibrary | Chrome | alias2 |
```
it will open two browsers. 

Also if alias is not defined
```
| Open Browser | https://github.com/robotframework/SeleniumLibrary | Chrome |
| Open Browser | https://github.com/robotframework/SeleniumLibrary | Chrome |
```
it will open two browsers. 

There is small change that someone is misusing the `alias` argument and therefore this may cause problems for users. But because the problem should be easy to fix and the probability for that is quite low, implementing this is OK. 

This idea is also presented in #701

rasjani:
Fixes #1319

### DIFF
--- a/src/SeleniumLibrary/keywords/browsermanagement.py
+++ b/src/SeleniumLibrary/keywords/browsermanagement.py
@@ -92,6 +92,8 @@ class BrowserManagementKeywords(LibraryComponent):
         approach for switching is using an index returned by this keyword.
         These indices start from 1, are incremented when new browsers are
         opened, and reset back to 1 when `Close All Browsers` is called.
+        If you use same ``alias`` with multiple instances of a same browser,
+        new instance won't be generated but existing one will be re-used.
         See `Switch Browser` for more information and examples.
 
         Optional ``remote_url`` is the URL for a
@@ -123,23 +125,37 @@ class BrowserManagementKeywords(LibraryComponent):
         Applying ``desired_capabilities`` argument also for local browser is
         new in SeleniumLibrary 3.1.
         """
+        def _resolve_browser_alias_to_browser_name(name):
+            return WebDriverCreator.browser_names.get(name.lower(), name)
+
         if is_truthy(remote_url):
             self.info("Opening browser '%s' to base url '%s' through "
                       "remote server at '%s'." % (browser, url, remote_url))
         else:
             self.info("Opening browser '%s' to base url '%s'." % (browser, url))
-        driver = self._make_driver(browser, desired_capabilities,
-                                   ff_profile_dir, remote_url)
-        driver = self._wrap_event_firing_webdriver(driver)
+
+        try:
+            driver, driver_index = self.ctx._get_driver(alias)
+            browser_name = _resolve_browser_alias_to_browser_name(browser)
+            if driver.name not in browser_name:
+                self.warn("Multiple browsers with alias %s being used." %
+                          alias)
+                raise RuntimeError
+
+        except (RuntimeError, ValueError, TypeError):
+            driver = self._make_driver(browser, desired_capabilities,
+                                       ff_profile_dir, remote_url)
+            driver = self._wrap_event_firing_webdriver(driver)
+            driver_index = self.ctx.register_driver(driver, alias)
+
         try:
             driver.get(url)
         except Exception:
-            self.ctx.register_driver(driver, alias)
             self.debug("Opened browser with session id %s but failed "
                        "to open url '%s'." % (driver.session_id, url))
             raise
         self.debug('Opened browser with session id %s.' % driver.session_id)
-        return self.ctx.register_driver(driver, alias)
+        return driver_index
 
     @keyword
     def create_webdriver(self, driver_name, alias=None, kwargs={},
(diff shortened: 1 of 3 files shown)


src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -123,23 +125,36 @@ def open_browser(self, url, browser='firefox', alias=None,
...
                       "remote server at '%s'." % (browser, url, remote_url))
         else:
             self.info("Opening browser '%s' to base url '%s'." % (browser, url))
-        driver = self._make_driver(browser, desired_capabilities,
-                                   ff_profile_dir, remote_url)
-        driver = self._wrap_event_firing_webdriver(driver)
+
+        try:
+            driver, driver_index = self.ctx._get_driver(alias)
+            browser_name = _resolve_browser_alias_to_browser_name(browser)
```
rasjani, line 139: this block here is to prevent scenario where user has created one webdriver instance with alias and then tries to create another webdriver with another browser again with the same alias .. 

Worth to mention that in that case, previous alias will be overwritten and alias lookup table in connection cache wont be able to find the old webdriver anymore with the alias even thought its still accessible by index.

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -123,23 +125,36 @@ def open_browser(self, url, browser='firefox', alias=None,
...
+            if driver.name != browser_name:
+                self.warn("Multiple browsers with alias %s being used." %
+                          alias)
+                raise RuntimeError
```
rasjani, line 143: I was thinking that maybe a separate exception here could be useful to make it clear that `_get_driver()` can throw Runtime, Value & TypeExceptions if the look fails.

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -123,23 +125,36 @@ def open_browser(self, url, browser='firefox', alias=None,
...
+        except (RuntimeError, ValueError, TypeError):
+            driver = self._make_driver(browser, desired_capabilities,
+                                       ff_profile_dir, remote_url)
+            driver = self._wrap_event_firing_webdriver(driver)
+            driver_index = self.ctx.register_driver(driver, alias)
+
         try:
             driver.get(url)
         except Exception:
-            self.ctx.register_driver(driver, alias)
```
rasjani, line 137: register driver is done right after its created so i removed this and below one because with our without issues, the driver *is* registered, albeit the driver's index wont be returned...

aaltat:
I restarted Travis, it looked like a problem with starting the tests. I need to look at changes from a bigger screen.

rasjani:
no changes to the code and  more sub builds failed. Im 100% sure that there was a build that only had failure in 2.7 but i cant spot that now =(

aaltat:
Jython job is really pain in the ***, I am waiting eagerly 1.1 2020 so that I can delete it from the Travis config (plus few other places). Also there seems to be connection problems with the Selenium/ChromeDriver/Chrome in tests with Python 2. Unfortunately logs are not preserved and console output is only way to debug.

aaltat:
Forgot to say that Travis has been acting this week quite a lot. So I will restart the Python 2 run few more times to see does it help.

aaltat:
Well there clearly is a problem, but I don't know where that problem is. I did yesterday make a PR and it worked with Travis fine. Also your PR works with Python 3, Jython and Pypy, but not with Python 2. Which is strange, because Jython and Python 2 should be quite similar. Usually if there is a Python 2 compatibility issue, both Jython and Python 2 fails. Have you run locally with Python 2 and seen that it works?

rasjani:
I get the same results on master & fix-1319 branch witth python2 and python3 and all of them do not look like would be caused by my addition.

Issue #1346, opened 2019-03-20
Remove deprecated attributes of public API


mihaiparvu:
The depraction messages were introduced 2 years ago in https://github.com/robotframework/SeleniumLibrary/pull/884/commits/14bbd1462231c19ed83921b9f5815dae2e1bc674. I think it's safe to remove them.

aaltat:
Fixed in #1347

PR #1347, opened 2019-03-20
Remove deprecated attributes

mihaiparvu:
PR for #1346.


aaltat:
In same go, we could remove the also methods and attributes usgin `warnings.warn`. Mostly things explained in here:  https://github.com/robotframework/SeleniumLibrary/blob/master/src/SeleniumLibrary/__init__.py#L652

mihaiparvu:
Due to recent removals, the class `Deprecated` from [utils/deprecated.py](https://github.com/robotframework/SeleniumLibrary/blob/master/src/SeleniumLibrary/utils/deprecated.py) is left unused. Should I delete it or keep it for future deprecations?

aaltat:
Ah, I did not remember that. If it is not used, then it can be removed.

aaltat:
Thank you for the pull request.

PR #1353, opened 2019-03-23
Fix Press Keys test to work in mac

### LINKED ISSUE #1352: Fix Press Keys acceptance test to work with Mac
In Mac, `CTRL` does not function in same way as in Windows or Linux. In mac `COMMAND` key should be used when testing Press Keys keyword.

aaltat:
Fixes #1352


aaltat:
@rasjani this should fix the `Press Keys` acceptance test in Mac.

PR #1356, opened 2019-03-23
Fixing html5 inputs for different locales

### LINKED ISSUE #1344: HTML 5 input acceptance test may fail if locale doesn't match
#1327 introduced tests for the HTML 5 different input types. But because date and time is involved, locale plays an effect and for some users the test may fail. 

Convert date and perhaps time to always same format to fix the issue.

aaltat:
Fixes #1344


aaltat:
@rasjani another fix for different locales.

PR #1360, opened 2019-03-27
Improve implcit wait testing

### LINKED ISSUE #1362: Improve Implicit Wait Tests
Current implicit wait test, test only that implicit wait is saved to the library instance. Improve testing that implicit wait has also affect.

aaltat:
Fixes #1362


aaltat:
It would be good to move `Browser Open With Implicit Wait Should Not Override Default` test from acceptance level to unit tests level. Also test few invalid values in unit tests.

PR #1368, opened 2019-04-15
New extending documentation

### LINKED ISSUE #1292: Allow extending SeleniumLibrary by creating plugin API.

aaltat:
Fixes #1292 documentation.

### DIFF
--- a/docs/extending/event_firing_webdriver/MyListener.py
+++ b/docs/extending/event_firing_webdriver/MyListener.py
@@ -1,0 +1,23 @@
+from robot.api import logger
+from selenium.webdriver.support.events import AbstractEventListener
+
+
+class MyListener(AbstractEventListener):
+
+    def before_navigate_to(self, url, driver):
+        logger.info("Before navigate to %s" % url)
+
+    def after_navigate_to(self, url, driver):
+        logger.info("After navigate to %s" % url)
+
+    def before_click(self, element, driver):
+        logger.info("Before click")
+
+    def after_click(self, element, driver):
+        logger.info("After click")
+
+    def before_change_value_of(self, element, driver):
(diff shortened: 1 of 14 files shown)


aaltat:
At least example should be modified to be able run, now they will fail. Also it would be good to update readme to say how examples can be run. Also linking to SeleniumLibrary installations instructions would be good idea.

PR #1374, opened 2019-04-21
Same alias should not open new browser

### LINKED ISSUE #1319: It should be possible to configure Open Browser keyword to not always open a new browser.

aaltat:
Fixes #1319


utest/test/keywords/test_webdrivercache.py
```diff
@@ -83,3 +83,35 @@ def test_close_only_called_once(self):
...
+        cache.register(mock())
+
+        index = cache.get_alias_or_index('bar')
+        self.assertEqual(index, None)
+
+        index = cache.get_alias_or_index(12)
+        self.assertEqual(index, None)
+
+        index = cache.get_alias_or_index(-1)
+        self.assertEqual(index, None)
```
aaltat: New line missing

atest/acceptance/multiple_browsers_with_alias.robot
```diff
@@ -0,0 +1,37 @@
+*** Settings ***
+Suite Teardown    Close All Browsers
+Resource          resource.robot
+
+
```
aaltat: Remove extra empty line

aaltat:
TODO: Fix `Open Browser` keyword documentation. Also fix two my comments.

aaltat:
And it looks like I broke something with the Python 2 compatibility. Need to look it more closely.

aaltat:
It was not Python 2 but it was RF 3.0.4. When using string None as alias, it returned index which was already closed. But it was a good thing, because it lead me think few other cases and it did reveal few bugs in the code. Now I hope that it is fixed and working. 

FYI @rasjani I think your PR did ran in to same problem, this did fix the problem: https://github.com/robotframework/SeleniumLibrary/pull/1374/files#diff-566132cdf5a534d7e99598e5acc77378R243

PR #1393, opened 2019-05-31
Selenium options string to object

### LINKED ISSUE #1331: Open Browser keyword could take in Selenium browser specific options
Selenium has been enhanced to support browser specific options for Firefox, Chrome, IE, Edge and Opera for some time ago. Since SeleniumLibrary has raised the minimum required [Selenium version to 3.8.1](https://github.com/robotframework/SeleniumLibrary/issues/1305), the Selenium options API has been stable. Example for [Chrome WebDriver](https://selenium-python.readthedocs.io/api.html#module-selenium.webdriver.chrome.webdriver) there is [Chrome WebDriver Options](https://selenium-python.readthedocs.io/api.html#module-selenium.webdriver.chrome.options). Similar system exist for all the browsers mentioned in above.

Taking in the browser specific options in the [Open Browser](http://robotframework.org/SeleniumLibrary/SeleniumLibrary.html#Open%20Browser) keyword should not be hard. Example the keyword signature could look like this: 
```
| Open Browser | url | browser | alias | remote_url | desired_capabilities | ff_profile_dir | options |
```
The Selenium options is an Python class instance and not a dictionary. Each browser specofic options object has different set of methods and attributes. 

It was decided to enhance the `Open Browser` keyword to take in Selenium options by using `options` argument. The `options` argument should support two different formats: string and Python object which is Selenium options instance. Example of a string format: 
```
| Open Browser | url | browser | options=method('arg'); attr = 'value'; method('foo', True) |
```

This enhancement proposal is one step forward to unify the `Open Browser` and `Create WebDriver` keywords. See #701

aaltat:
Fixes #1331


pekkaklarck:
Python 3.4 has reached EOL so there's no real need to support it anymore. Python 3.5 doesn't preserve dictionary order either, though. Need to wait until Python 3.6 for that.

aaltat:
Good point and that makes me think should SL drop support for Python 3.4? Or will the announcement come too late and perhaps it should be postponed to next major release?

pekkaklarck:
Don't think it matters much either way. It's unlikely there are many users that would be affected, but I doubt there's any extra work supporting Python 3.4 either. If the next release is close, perhaps support 3.4 with it still but then drop the support. I guess that would mean only stopping testing with it and removing classifier for `setup.py`, no reason to actively prohibit using 3.4 even then.

aaltat:
It seems that ChromeDriver has an issue with Selenium ActionChains and not all works like in the previous version. Hard coded the ChromeDriver version for now in #1394

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -117,6 +117,70 @@ def open_browser(self, url, browser='firefox', alias=None,
...
+        attributes are separated by a comma and arguments are separated
+        with a colon. Example like this:
+        `argument:arg1,other_argument:arg2:arg3,attribute:value`.
+
+        The method and attributes are case and space sensitive
+        and must match to the Selenium options methods and attributes names.
+        It is possible to have space before and after the comma for readability.
+        Example `argument:arg1,argument:arg2` and
+        `argument:arg1 , argument:arg2` are converted to same
+        object internally. Spaces around arguments are not removed.
```
pekkaklarck: Does "Spaces around arguments are not removed." mean that in the space in the previous example `argument:arg1 ,` would be significant?

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -117,6 +117,70 @@ def open_browser(self, url, browser='firefox', alias=None,
...
+        Arguments are always stings and are not converted to other types.
```
pekkaklarck: Another `sting`. Probably should search for possible other occurrences.

src/SeleniumLibrary/keywords/browsermanagement.py
pekkaklarck: Also, `argument:arg1` is somewhat confusing because it's using word argument twice.

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -117,6 +117,70 @@ def open_browser(self, url, browser='firefox', alias=None,
...
+        and for Firefox these
+        [https://seleniumhq.github.io/selenium/docs/api/py/webdriver_firefox/selenium.webdriver.firefox.options.html?highlight=firefox#selenium.webdriver.firefox.options.Options|methods and attributes]
+        are available. Please note that not all browsers supported by the
+        SeleniumLibrary have Selenium options available, please consult
+        the Selenium documentation which browser do support the Selenium
+        options. Selenium options are also supported, when ``remote_url``
+        argument is used.
+
+        The SeleniumLibrary ``options`` argument accepts Selenium
+        options in three different formats 1) As a sting, 2) List containing
```
pekkaklarck: No need to capitalize `As` or `List` here.

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -117,6 +117,70 @@ def open_browser(self, url, browser='firefox', alias=None,
...
+        dictionaries 3) Python object which is an instance of the
+        Selenium options object.
+
+        The sting format is allows to define Selenium options methods
```
pekkaklarck: - `sting` -> `string`
- `is allows` -> `allows`

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -117,6 +117,70 @@ def open_browser(self, url, browser='firefox', alias=None,
...
+        or attributes and it's arguments in string format. Methods and
```
pekkaklarck: - `it's` -> `its` or (probably better) `theirs`
- Sentence starting and ending with "string format" is a bit odd.

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -117,6 +117,70 @@ def open_browser(self, url, browser='firefox', alias=None,
...
+        It is possible to have space before and after the comma for readability.
+        Example `argument:arg1,argument:arg2` and
+        `argument:arg1 , argument:arg2` are converted to same
+        object internally. Spaces around arguments are not removed.
+        Arguments are always stings and are not converted to other types.
+        If there is need to define a literal colon for an argument,
+        it can be escaped by using literal backslash. Please
+        note that backslash is an escape character in Robot Framework and
+        therefore backslash must escaped. Example:
+        `argument:arg1_with\\\\:colon`.
```
pekkaklarck: Is there really a need for escaping colon? I doubt attribute names can actually contain colons, so it ought to be fine to consider values like `foo:bar:zap` to mean that `foo` is the attribute and `bar:zap` is the value.

There's an actual problem escaping commas, though. Or does the code split first from colon and only then from comma? That would be somewhat weird but I haven't yet looked the code.

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -117,6 +117,70 @@ def open_browser(self, url, browser='firefox', alias=None,
...
+
+        List containing dictionaries format is similar to the earlier
+        string format, but defines the methods and attributes and their
+        arguments using Robot Framework lists and dictionaries.
+        Method or attribute arguments, must be defined in a list. Method
+        or argument must be a dictionary key and argument list must be
+        the dictionary value. Dictionary must contain only one key
+        value pair. The dictionary must be placed in a list. Example:
+        `[{"argument": ["arg1"]}, {"other_argument": ["arg1", "arg2"]}]`
```
pekkaklarck: This format looks pretty strange. Why not to have just one dictionary that has all possible attribute/value pairs? For me it seems `[{'x': 1}, {'y': 2}]` and `{'x': 1, 'y': 2}` have same information but the latter format is a lot simpler especially in Robot's data.

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -117,6 +117,70 @@ def open_browser(self, url, browser='firefox', alias=None,
...
+        attribute name, is case and space sensitive and must match to
+        the Selenium options methods and attributes names. Although
+        the format is more complex to define, it allows to use Robot Framework
+        Python object, like `${True}` or `${None}`.
+
+        As last format ``options`` argument also support receiving
+        the Selenium options as Python class instance. In this case, the
+        instance is used as is and the SeleniumLibrary will not convert
+        the instance to other formats.
+        Example, if the following code return valua is saved to `${options}`
```
pekkaklarck: - `Example` -> `For example`
- `valua` -> `value`

src/SeleniumLibrary/keywords/webdrivertools.py
```diff
@@ -48,23 +49,25 @@ class WebDriverCreator(object):
...
             logger.info('Browser driver log file created to: %s' % service_log_path)
             self._create_directory(service_log_path)
         if (creation_method == self.create_firefox
                 or creation_method == self.create_headless_firefox):
-            return creation_method(desired_capabilities, remote_url,
-                                   profile_dir, service_log_path=service_log_path)
-        return creation_method(desired_capabilities, remote_url, service_log_path=service_log_path)
+            return creation_method(desired_capabilities, remote_url, profile_dir,
+                                   options=options, service_log_path=service_log_path)
+        return creation_method(desired_capabilities, remote_url, options=options, service_log_path=service_log_path)
```
pekkaklarck: Split line?

src/SeleniumLibrary/keywords/webdrivertools.py
```diff
@@ -104,11 +107,11 @@ def create_chrome(self, desired_capabilities, remote_url, options=None, service_
             return self._remote(desired_capabilities, remote_url, options=options)
         return webdriver.Chrome(options=options, service_log_path=service_log_path, **desired_capabilities)
 
-    def create_headless_chrome(self, desired_capabilities, remote_url, service_log_path=None):
-        options = webdriver.ChromeOptions()
+    def create_headless_chrome(self, desired_capabilities, remote_url, options=None, service_log_path=None):
+        chrome_options = webdriver.ChromeOptions() if not options else options
```
pekkaklarck: I'd use
```python
options = options or webdriver.ChromeOptions()`
```
or
```python
if not options:
    options = webdriver.ChromeOptions()
```

src/SeleniumLibrary/keywords/webdrivertools.py
```diff
@@ -133,86 +136,108 @@ def _geckodriver_log(self):
         return log_file
 
     def create_headless_firefox(self, desired_capabilities, remote_url,
-                                ff_profile_dir, service_log_path=None):
-        options = webdriver.FirefoxOptions()
+                                ff_profile_dir, options=None, service_log_path=None):
+        ff_options = webdriver.FirefoxOptions() if not options else options
```
pekkaklarck: Same comment here as to Chrome options above.

src/SeleniumLibrary/keywords/webdrivertools.py
```diff
@@ -133,86 +136,108 @@ def _geckodriver_log(self):
...
+        if self._has_service_log_path(webdriver.Ie) and self._has_options(webdriver.Ie):
+            # service_log_path is supported from Selenium 3.14 onwards
+            # If can be removed when minimum Selenium version is 3.14.0 or greater
+            return webdriver.Ie(options=options, service_log_path=service_log_path, **desired_capabilities)
+        elif not self._has_service_log_path(webdriver.Ie) and self._has_options(webdriver.Ie):
+            # options is supported from Selenium 3.10 onwards
+            # If can be removed when minimum Selenium version is 3.10.0 or greater
+            logger.warn('This version of Selenium does not support service_log_path argument.')
+            return webdriver.Ie(options=options, **desired_capabilities)
+        logger.warn('This version of Selenium does not support options or service_log_path argument.')
```
pekkaklarck, line 168: These changes look somewhat unrelated to other changes.

src/SeleniumLibrary/keywords/webdrivertools.py
```diff
@@ -133,86 +136,108 @@ def _geckodriver_log(self):
...
         return webdriver.Ie(**desired_capabilities)
 
     def _has_service_log_path(self, web_driver):
         signature = inspect.getargspec(web_driver.__init__)
         return True if 'service_log_path' in signature.args else False
 
-    def create_edge(self, desired_capabilities, remote_url, service_log_path=None):
+    def _has_options(self, web_driver):
+        signature = inspect.getargspec(web_driver.__init__)
+        return True if 'options' in signature.args else False
```
pekkaklarck: It would be quite a bit better to use just
```python
return 'options' in signature.args
```

src/SeleniumLibrary/keywords/browsermanagement.py
aaltat: 1) Yes it would be meaningful and therefore it's a bad example. Will fix it. After reading that whole part, perhaps it would be better just to say that: `space after comma is allowed for readability.` Because that is the way it works. Also should check that parsing is aligned with the plugin API parsing because they do similar thing. Perhaps even see can both parsing live in same code.

2) Very true, something like `method:arg1` would be better.

src/SeleniumLibrary/keywords/browsermanagement.py
aaltat: There are some methods in Selenium options that takes two arguments in, example like https://selenium-python.readthedocs.io/api.html#selenium.webdriver.firefox.options.Options.set_capability and presenting two arguments as a list was an easiest option.

I know that it's an overkill when a method or an attribute has a single argument, but I initially didn't want to support two formats. I initially thought that it would be harder to explain in the documentation and thought that it would bring complicity to the code. But when I am now looking at the matter I am not anymore so sure. The code will not be a that much of a problem but what would be a good way to explain it?

src/SeleniumLibrary/keywords/browsermanagement.py
aaltat: Yes escaping can be needed, example for defining a proxy could have a colon. `--proxy:ip\\:port` is possible. But proxy definition was the only example which I could find and therefore escaping is done in very limited manner.

src/SeleniumLibrary/keywords/webdrivertools.py
aaltat: Ie and some other browser was difficult to make, because support for Selenium options and service_log_path came in different Selenium version. Therefore, depending on the Selenium version `Ie()` needs to be called with different arguments.

src/SeleniumLibrary/keywords/browsermanagement.py
aaltat: Also many methods can be called many times, like the https://selenium-python.readthedocs.io/api.html#selenium.webdriver.firefox.options.Options.add_argument and that was not possible in the dictionary only format.

aaltat: Explained in bad way, please ignore the above comments. Lets try again. 

The dictionary only format is not possible because many Selenium options methods are called (usually) many times. 
Example `{"add_argument": "value1", "add_argument": "value2"}` is not possible in the dictionary format. Therefore there is list in the outermost definition: `[{...}, {...}]`. 

Also there are Selenium options methods which have two arguments and therefore there is a list inside of the dictionary. But those methods are actually a minority and based on the found internet examples, not used that much. Therefore supporting two different formats could be be possible: 1) If there is only one argument, list inside of the dictionary would be optional, like: `[{"add_argument": "value1"}, {"add_argument": "value2"}]` 2) If there are two (or more) arguments, list inside of the dictionary is mandatory, like `[{"set_capability": ["key1", "value1"]}, {"set_capability": ["key2", "value2"]]`. I did initially reject the idea of two formats, because it would be harder to explain the documentation and would bring complexity to the code. But now when looking at the matter, code will not be complex, only one `if else` statement, but will it be easy to explain in the documentation.

Also in some point, I was thinking not the support the `[{...}, {...}]` format, but Selenium contains some stupid code, [like in headless](https://github.com/SeleniumHQ/selenium/blob/master/py/selenium/webdriver/chrome/options.py#L151) where Python `True` is explicitly required. It was also possible to make automatic conversion in the string format, but I did not want to go in that path, because it was hard to decide what format should be converted.

I agree that the  `[{...}, {...}]` format is messy and not easy to create in RF data. Would you see other and better options?

src/SeleniumLibrary/keywords/webdrivertools.py
aaltat: Also there was a bug which was fixed in drive by manner, but I do not anymore recall was it this or some other place.

aaltat:
Thank you for the comments, useful feedback. I did provide answers to the points which needed clarification. If I did not comment, I do agree with the comment.

src/SeleniumLibrary/keywords/browsermanagement.py
pekkaklarck: I don't understand. Why wouldn't `name, value = item.split(':', 1)` work? It wouldn't handle cases like `name:with:colon:this-is-the-value` where name is `name:with:colon` and value is `this-is-the-value`, but this is only a problem if names can have colons. Method or attribute names certainly cannot.

src/SeleniumLibrary/keywords/browsermanagement.py
pekkaklarck: This is clearly a complicated problem to solve. Hard to have a generic solution but keep the common case simple. Anyway, I still don't like the list of dicts format too much.

One problem I see is that it's not clear does `[{'method': 'x'}, {'method': 'y'}]` map to a single call with two arguments like `obj.method('x', 'y')` or to two individual calls with one argument like `obj.method('x'); obj.method('y')`.

Instead of a list with one item dictionaries the format could be a list with internal lists (or tuples). Then `[('method', 'x'), ('method', 'y', 'z')]` could be interpreted as `obj.method('x'); obj.method('y', 'z')`. This ought to be pretty generic. It would handle even cases where arguments needs to be a list such as `[('method', [1, 2, 3])]` that would be `obj.method([1, 2, 3])`.

If the common case is setting attributes or calling methods only once and with one argument, then it might be a good idea to support specifying methods/attributes as a dictionary like `{'method': 'arg', 'attribute': 'value'}`. In the code this could be trivially converted to a list of lists (technically tuples) with code like `if isinstance(item, Mapping): item = item.items()`.

pekkaklarck: Is there a way in the string format to call a method with more than one argument? Also, it's somewhat annoying if the API needing actual `True` makes it impossible to use this format.

What if the string format would be like `method('arg'); attr = 'value'; method('foo', True)`? This would be somewhat technical, but it would also look exactly like the same Selenium code in Python. It would require strings to be quotes but I don't think that's a huge problem. Parsing wouldn't be trivial but shouldn't be overly complicated either. I can help if needed.

Hmm, if this string format works well enough, then perhaps there wouldn't be need for the list of dicts/lists format at all. Externally created option objects probably should still be supported.

src/SeleniumLibrary/keywords/browsermanagement.py
aaltat: Some Selenium options methods take more than one argument in. Example
`method:arg, other_method:foo:bar` should be translated to 
```
options = webdriver.ChromeOptions()
options.method("arg")
options.other_method("foo", "bar")
```
But in 
`method:arg\:param`
it should be 
```
options = webdriver.ChromeOptions()
options.method("arg:param")
```
And in that proxy scenario argument contains a colon and it is the only case, which I have found, where colon is needed.

src/SeleniumLibrary/keywords/browsermanagement.py
aaltat: I agree that the `method('arg'); attr = 'value'; method('foo', True)` format is better. Although there is some more work to write it and also parsing is not trivial the benefits are greater than those two downsides. 

1) It is clear for the reader is user calling an attribute or a method.
2) There is a clear separation between Python booleans (Like `True`), numbers (although I haven't found an example where number would be needed) and strings.
3) This replaces the hard to make and read `[{...}, {...}]` format. 
4) Documentation will hopefully be less confusing, because there is only one format to explain. 

Yes external objects should be supported, but that is trivial.

aaltat:
Changed the options string format. All the other comments are not yet done.

aaltat:
I was thinking, that if there is this argument which has semicolon inside, example like this `method(" arg;after")` now the parsing will fail. This is because code will blindly split from the semicolon. This can be fixed in multiple ways, the semicolon can be escaped or splitting can be enhanced so that it doesn't split blindly. But it's worth of the effort, I can not recall seeing an argument which would have semicolon.

pekkaklarck:
I actually commented about handling semicolons that are part of arguments earlier. They can be handled by tokenizing the string and, assuming the string is Python, that can be done using [tokenize.generate_tokens](https://docs.python.org/2/library/tokenize.html#tokenize.generate_tokens). It has a really strange API, but [BuiltIn.evaluate](https://github.com/robotframework/robotframework/blob/master/src/robot/libraries/BuiltIn.py#L3009)  uses it succesfully. Here's a prototype:

```python
>>> from tokenize import generate_tokens
>>> import token
>>> from StringIO import StringIO    # Python 2
>>> options = "method('arg;with;colons'); attr = 'value;with;colons'"
>>> for toknum, tokval, start, _, _ in generate_tokens(StringIO(options).readline):
...     if tokval == ';' and toknum == token.OP:  # checking toknum may not be needed
...         row, col = start    # should we take into account multiline strings?
...         print("Found '%s' at index '%s'." % (tokval, col))
... 
Found ';' at index '25'.
>>> print(options[:25])
method('arg;with;colons')
>>> print(options[25+1:].lstrip())
attr = 'value;with;colons'
```

The same approach could possibly be used to separate `method('arg')` and `attr = 'value'` from each others and for parsing them in general, but that may be easier with regexps or just string methods once the option string has been split from `;`.

aaltat:
I did not see that it would be possible in that way. And true, it has really strange API. Will fix the code.

pekkaklarck:
I'd create a helper function/method that accepts the option string, finds semicolons, and yields individual options. Last item returned by generated_tokens is the logical line containing the token and the part before the `;` could be easily yield based on that. Such a helper would also be easy to unit test.

aaltat:
Improved the parser to only split between methods or attributes. Also fixed the documentation, but other comments are not yet fixed.

aaltat:
I did read the review comments all should be now fixed. Although reading documentation would be helpful.

aaltat:
One this is merged, I was planning to do new alpha release. Then fix last few opens issues and push release candidate out.

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -119,6 +119,69 @@ def open_browser(self, url, browser='firefox', alias=None,
         ``ff_profile_dir`` can also be instance of the
         [https://seleniumhq.github.io/selenium/docs/api/py/webdriver_firefox/selenium.webdriver.firefox.firefox_profile.html?highlight=firefoxprofile#selenium.webdriver.firefox.firefox_profile.FirefoxProfile|selenium.webdriver.FirefoxProfile].
 
+        Optional ``options`` argument allows to define browser specific
+        Selenium options. Example for Chrome, the ``options`` argument
+        allows defining the following
+        [https://seleniumhq.github.io/selenium/docs/api/py/webdriver_chrome/selenium.webdriver.chrome.options.html#selenium.webdriver.chrome.options.Options|methods and attributes]
+        and for Firefox these
```
emanlove, line 126: Just checking to make sure Firefox profile does not overlap or interfere with Firefox options ..? Read through the API docs and it looks like no they are mutually exclusive.

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -119,6 +119,69 @@ def open_browser(self, url, browser='firefox', alias=None,
...
+        the defining the methods or attributes. There are two exceptions.
+        In some Robot Framework test data formats, two or more spaces are
+        considered as cell separator and instead of defining a single
+        argument, two or more arguments may be defined. Spaces in string
+        arguments are not removed and are left as is. Example
+        `add_argument ( "--headless" )` is same as
+        `add_argument("--headless")`. But `add_argument(" --headless ")` is
+        not same same as `add_argument ( "--headless" )`, because
+        spaces inside of quotes are not removed.
+
```
emanlove, line 173: One difficulty with supporting a string formatted options is being backwards compatible over time or change the rules of how the string is formatted. You might try out the formatting with several users to see how it works for them.

src/SeleniumLibrary/keywords/browsermanagement.py
aaltat: Those two partially overlap but generally Firefox options allows to do more. But they are not exclusive, one can define both and all should work fine. I did briefly tested the scenario and did not find problems. Although I did not dive deeply on the matter and I don't know if same setting is defined in both, which one has priority over other.

src/SeleniumLibrary/keywords/browsermanagement.py
aaltat: That is a worry which I also have, but there are some safe guards. Calling the Selenium options is as much as possible backwards compatible, because behind the scenes it does a similar trick as the `Create WebDriver` keyword with Python `getattr`, `setattr` and `callable`, [more details in here](https://github.com/robotframework/SeleniumLibrary/pull/1393/files#diff-566132cdf5a534d7e99598e5acc77378R388). Therefore keyword leaves to the user to correctly defined and maintain the methods and attributes user wants to call.

But there is a part that worries me and that is the importing options. First Selenium options import path may change or the options class name may change. In the past, that part has been quite stable and I have not seen any changes in the upcoming Selenium 4.0 release either. But who knows what future will bring.

Also mobile devices, which are supported on the `Open Browser` are always done by using the  `webdriver.Remote`. The remote WebDriver supports options object, but those mobile devices don't have their own opinions object (Like example Chrome or Firefox has). Therefore the keyword does not support creating options object for those mobile devices by using the string format. Also there is feature/bug when Selenium options instance object is passed directly in, for mobile devices, [it will fail](https://github.com/robotframework/SeleniumLibrary/pull/1393/files#diff-566132cdf5a534d7e99598e5acc77378R353), but that feature can be lifted away. But there is some indication that for example Android might directly support Chrome options http://chromedriver.chromium.org/getting-started/getting-started---android Also I know that some users are using SeleniumLibrary for mobile testing because, according to some users feedback: "SeleniumLibrary has better keywords but is little bit tricky to setup." Therefore it might be interesting to support also options for mobile devices, but currently I lack the knowledge how that part actually works and the Travis testing infrastructure is not not setup (Although there seems to be [support for mobile](https://docs.travis-ci.com/user/languages/android/))

In any case, if the limitation in the string format is left as it is, it should be documented. Also we should remove the limitation when the Selenium options instance is passed in for mobile devices. That limitation should be easy to change, because now we support only two formats.

pekkaklarck: It's impossible to make this fully future compatible because Selenium doesn't have a stable API for specifying options. The current approach to support specifying options like `method('arg')` and `attr = 'value'` has a benefit that users can adjust their code if Selenium changes.

aaltat: That is true. I still need to think should we do some extra for the mobile users. But perhaps easiest way is to: 
1) Fix the bug with mobile and Selenium options
2) Android == Chrome for selenium options. + Need documentation. 

Make alpha release and ask specially for people using mobile to take a look.

src/SeleniumLibrary/keywords/browsermanagement.py
aaltat: Fixed.

src/SeleniumLibrary/keywords/browsermanagement.py
aaltat: Fixed.

src/SeleniumLibrary/keywords/browsermanagement.py
aaltat: Fixed.

aaltat:
Android == Chrome options but is not documented. Once that is done, this monster can be merged to the master.

Issue #1388, opened 2019-05-20
Enhance documentation to explain why `Window...` keywords do not always see other browsers windows


aaltat:
There was a long discussion in the Slack, why this fails in the last line: 
```
*** Settings ***
Library  SeleniumLibrary

Suite Teardown  Close all browsers

*** Variables ***
${BROWSER}  chrome

*** Test cases ***
Whatever
    ${one}=  open browser  about:blank  ${browser}
    ${two}=  open browser  about:blank  ${browser}
    ${three}=  open browser  about:blank  ${browser}

    @{identifiers}=  get window identifiers
    length should be  ${identifiers}  3
```
It was explained that each time browser is opened by `Open Browser` keyword it creates new WebDriver instance. Each WebDriver sees only the browser that it has opened and the child browsers. I never looked deeply how it actually works, but I believe that it's related to the process communication. It doesn't see the browsers opened by other WebDrivers.

Although we could change the SeleniumLibrary library functionality that it would list windows from all WebDrivers. And allows to change also between WebDriver`s 

But it would be also confusing for users, because it would change the active WebDriver and it could implicitly change how test works. I have never used those keywords that much, so I don't have strong opinion, which way would be better: 1) Current, listing only windows in the single WebDriver 2) Change to list all windows from all WebDrivers. Also change the active WebDriver in the SeleniumLibrary.

The consensus was that documentation must be enhanced, to explain the above functionality (currently it doesn't not explain). Also there has been discussion that should the keywords return all windows from all WebDrivers. But this issue is only for enhancing the documentation and if there new keywords needs to be created or existing one`s enhanced, that is done in a separate issue.

In the Slack, Bryan suggestion: _”current browser and its child windows”_ or something like that might be more clear.

PR #1412, opened 2019-07-10
Modified Window Handling to be able to work over all active Browsers.

Snooz82:
There are a few "Get" Keywords which return information about all windows of one browser. With this change it is also possible to get the window information of all windows of all browser.

Updated documentation about the relation of browsers and their windows.
Giving an example how to handle multiple browsers and multiple windows.

Cleanup:
Moving all keywords that where window related and not browser related from "BrowserManagementKeywords" to "WindowKeywords"


Snooz82:
fixes Issue: https://github.com/robotframework/SeleniumLibrary/issues/1388

Snooz82:
@aaltat 
i moved these Keywords from BrowserManagementKeywords to WindowKeywords:
If you want, i can undo this and move it to another Pull-Request.

def maximize_browser_window(self):
def get_window_size(self, inner=False):
def set_window_size(self, width, height,   inner=False):
def get_window_position(self):
def set_window_position(self, x, y):
def get_source(self):
def get_title(self):
def get_location(self):
def location_should_be(self, url,   message=None):
def location_should_contain(self,   expected, message=None):
def log_location(self):
def log_source(self, loglevel='INFO'):
def log_title(self):
def title_should_be(self, title,   message=None):
def go_back(self):
def go_to(self, url):
def reload_page(self):

aaltat:
I looked your code on mobile but I need to take a closer look with real PC. 

But in mean time, I think moving keywords did break some of the unit tests. Could you take a look what did actually break and fix it?

aaltat (changes requested):
That bash enforcement is not needed. Please remove it.

Snooz82:
@aaltat i modified the dynamic_content.html to be able to set the window title with a free text. in local Tests it had no negative effects to other atests.

I think it is ready to be reviewed... ;-)

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -350,14 +350,45 @@ def switch_browser(self, index_or_alias):
         self.debug('Switched to browser with Selenium session id %s.'
                    % self.driver.session_id)
 
+    @keyword
+    def get_browser_ids(self):
```
aaltat, line 354: I did not understand where this keyword would be useful? Let's discuss.

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -350,14 +350,45 @@ def switch_browser(self, index_or_alias):
...
+        """Returns index of all active browser as list.
+
+        See `Switch Browser` for more information and examples.
+
+        New in SeleniumLibrary 4.0
+        """
+        return self.drivers.active_driver_ids
+
+    @keyword
+    def get_browser_aliases(self):
```
aaltat, line 371: I can see this be useful for logging purposes. but does this keyword server some other purpose too?

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -350,14 +350,45 @@ def switch_browser(self, index_or_alias):
...
+
     @keyword
     def get_session_id(self):
         """Returns the currently active browser session id.
 
         New in SeleniumLibrary 3.2
         """
         return self.driver.session_id
-
+
```
aaltat: Nit pic, spaces in here are not allowed.

src/SeleniumLibrary/keywords/window.py
```diff
@@ -20,7 +20,7 @@
 
 from SeleniumLibrary.base import keyword, LibraryComponent
 from SeleniumLibrary.locators import WindowManager
-from SeleniumLibrary.utils import plural_or_not
+from SeleniumLibrary.utils import (plural_or_not, is_string)
```
aaltat: Nit pic: Parenthesis hare are not needed.

aaltat (commented):
I did not read all the code, but I have some questions first.

src/SeleniumLibrary/keywords/browsermanagement.py
Snooz82: The idea of this PR was to resolve complexity of window and browser handling.
1st reason for `get browser ids` and aliases was that we have corresponding Keywords for windows. This is for completeness.

2nd reason: It can happen, that browsers are opened in a very high level Suite Setup. Or that multiple browsers are open. If you want to Switch to a specific browser that has a specific site, you had no possibility to go through all Browsers. You could just iterate over some numbers, ` in range 0-10` but if some browsers are allready closed you will get problems.

src/SeleniumLibrary/keywords/browsermanagement.py
Snooz82: Logging and iterating over open browsers. 

Same as 'get id' this is also for be able to get what we set.

If some other tests opened a browser you can access with this keyword.

src/SeleniumLibrary/keywords/browsermanagement.py
Snooz82: I will fix this

src/SeleniumLibrary/keywords/window.py
Snooz82: Fixed

src/SeleniumLibrary/keywords/browsermanagement.py
Snooz82: Done

src/SeleniumLibrary/keywords/browsermanagement.py
aaltat: I see your point. Then there should be documentation for the users how the keyword can be used with other keywords. There could be also tests to cover the usage scenario, so that we do not accidentally break the feature. 

Same comment goes with the `get_browser_aliases` comment.

atest/acceptance/multiple_browsers_multiple_windows.robot
```diff
@@ -0,0 +1,147 @@
+*** Setting ***
+Documentation     These tests must open own browser because windows opened by
+...               earlier tests would otherwise be visible to Get Window XXX keywords
+...               even if those windows were closed.
+Suite Setup        Open 3 Browsers with Windows
+Suite Teardown     Close All Browsers
+Resource          resource.robot
+
+*** Variables ***
+@{BrowserA_exp_Titles}=    WindowA1    WindowA2    WindowA3
```
aaltat: Nit pick: we prefer to write global variables with capital letters. If it does not hurt readability, could global variables be written in upper case.

src/SeleniumLibrary/keywords/browsermanagement.py
Snooz82: There are atests that are covering all of these Keyword!

I will do an example.

src/SeleniumLibrary/__init__.py
```diff
@@ -223,6 +224,89 @@ class SeleniumLibrary(DynamicCore):
...
     See the `Add Location Strategy` keyword for more details.
 
+    = Browser and Window =
+
+    In this section the concept of Browsers or Webdrivers and its Windows
+    will be explained.
+
+    == Browser ==
+
+    In Selenium Library there are two ways to start a Browser and control it.
```
aaltat: Hmmm, I am hesitant about this chapter. Would it be better to explain something like this: 

_Each time the Open Browser or Create WebDriver keywords are called, it will create a new browser. Also Internally it creates a Selenium WebDriver_ [link here selenium docs would be nice] _instance. The new browser is, usually, independent from other browsers opened by the in the test, example they do not share cookies or example profile_

I would not explain the what keywords are preferred and about the keyword history, I think it confuses the reader. The above example is quite minimal and requires some more text.

src/SeleniumLibrary/__init__.py
```diff
@@ -223,6 +224,89 @@ class SeleniumLibrary(DynamicCore):
...
+    The first one is the `Open Browser` keyword. This one should be used as
+    default keyword.
+    The second one is the `Create Webdriver` keyword which was useful, when
+    browser options needed to be set. Since Selenium Library 4.0.x `Open Browser`
+    also supports option why `Create Webdriver` will be deprecated.
+    When opening a browser a new Webdriver process is startet and connected to
+    Selenium Library. When using Selenium Grid this browser may be started on any
+    Grid node that has matching capabilities.
+    It is possible to start multiple independent browsers at the same test case.
```
aaltat: I like this chapter, let's keep this in.

src/SeleniumLibrary/__init__.py
```diff
@@ -223,6 +224,89 @@ class SeleniumLibrary(DynamicCore):
...
+
+    == Browser ==
+
+    In Selenium Library there are two ways to start a Browser and control it.
+    The first one is the `Open Browser` keyword. This one should be used as
+    default keyword.
+    The second one is the `Create Webdriver` keyword which was useful, when
+    browser options needed to be set. Since Selenium Library 4.0.x `Open Browser`
+    also supports option why `Create Webdriver` will be deprecated.
+    When opening a browser a new Webdriver process is startet and connected to
```
aaltat: Would it be good idea to have table, that describer the terms
| WebDriver | Selenium Python class to control browser |
Window, browser, browser driver (example chrome driver) kind a needs their own explanations. 

Also when explaining this thing, would I am starting to think that `browser` is not a good word. Instead it could be cleaner to talk about WebDriver, window and browser driver.

src/SeleniumLibrary/keywords/webdrivertools.py
```diff
@@ -294,6 +294,18 @@ def active_drivers(self):
...
+    @property
+    def active_driver_ids(self):
+        open_driver_ids = []
+        for index, driver in enumerate(self._connections):
+            if driver not in self._closed:
+                open_driver_ids.append(index + 1)
+        return open_driver_ids
+
+    @property
+    def active_aliases(self):
```
aaltat: Instead of wrapping like this, it would be good to change `_alias` as a public method. `_alias` -> `alias`

atest/acceptance/multiple_browsers_multiple_windows.robot
Snooz82: changed them to capital. 
But, i let "BrowserA" at This case because it is also the alias and used multiple times. So it is more a Name that a variable?! 
like BrowserA_EXP_TITLES

src/SeleniumLibrary/__init__.py
Snooz82: i think then it is starting to confuse people again. 

we could explain a lot...
Webdriver is a REST-API and Selenium is the inplementation and a Webdriver Server it the thing at the Browser and the client is Selenium. etc...
Then we could explaint, that Grid nodes are servers and the hub is a Client. 
And nobody would understand whats about windows and Browsers.
i would say, if people want to understand Selenium they should read their docs.

i tried to explain that a "browser" has "windows". I didnt want to explain that a Webdriver instance has a connection to a browserdriver etc...

src/SeleniumLibrary/__init__.py
Snooz82: Did refactor this chapter.
More details on Webdriver and Selenium Stuff.

src/SeleniumLibrary/__init__.py
Snooz82: Did refactor this chapter.
More details on Webdriver and Selenium Stuff.

src/SeleniumLibrary/__init__.py
Snooz82: Did refactor this chapter.
More details on Webdriver and Selenium Stuff.

src/SeleniumLibrary/keywords/webdrivertools.py
Snooz82: But _aliases is a field of the Class ConnectionCache, which is part of robot.utils...
Do you wand me to make a pull request to this? This would brake a lot of libraries.
Also SSH Library uses _aliases etc...

May be i completely misunderstood.

Snooz82:
> That bash enforcement is not needed. Please remove it.
removed...

src/SeleniumLibrary/keywords/browsermanagement.py
Snooz82: It is hard for me to make an understandable pice of robot code.

Would be something like nested for loops. over browsers and windows.
And this is only possible with keywords in Keywords in robot.
Which makes it hard do document in Table.

Usecase would be switch to a window by its sontent. 
Of course i could do something like this:
```
Logging Window Titles for Each Browser
    @{browser_ids}=    Get Browser Ids
    FOR    ${id}    IN    @{browser_ids}
        @{window_titles}=    Get Window Titles    browser=${id}
        Log    Browser ${id} has these windows: ${window_titles}
    END
```

I could do this into the doc...
But it would not work as atest, because old Robot Version does not support this loop.

aaltat:
There is conflict that must be resolved before this can be merged.

src/SeleniumLibrary/keywords/browsermanagement.py
Snooz82: another example of how `Get Browser Ids` could be usefull:

```
*** Keywords ***
Switch Window By Location
    [Arguments]    ${selected_location}
    @{IDs}=    Get Browser Ids
    FOR     ${id}    IN    @{IDs}
        @{locations}=    Get Locations    browser=${id}   
        Run Keyword If    '${selected_location}' in $locations
        ...    Switch Window    url:${selected_location}    browser=${id}
    END
```

aaltat:
I made modifications to the documentation, so that it would be more easier to understand. I did went to a direction that hides most the Selenium and browser driver stuff, because it felt confusing. Now it talks about browser and windows (mostly.) Now I am bit scared that I did lost something when I did cut down the details, because for me it is clear but it makes me blind to some problems. Should we still change it someway to make it better?

src/SeleniumLibrary/keywords/webdrivertools.py
aaltat: I did not know that, no need to change.

src/SeleniumLibrary/locators/windowmanager.py
```diff
@@ -39,7 +39,45 @@ def __init__(self, ctx):
             'default': self._select_by_default
         }
 
-    def get_window_infos(self):
+    def get_window_handles(self, browser):
+        if is_string(browser) and browser == 'ALL':
+            handles = []
+            current_index = self.drivers.current_index
+            for index, driver in enumerate(self.drivers, 1):
```
aaltat, line 46: Hmm, are you sure that this will work, because drivers is a [set](https://docs.python.org/3.7/library/stdtypes.html#set-types-set-frozenset)

Snooz82: But set is iterable..

class set([iterable])

Or what exactly are you afraid is not working?

aaltat: My bad, drivers are not in set. Please ignore

aaltat (approved):

PR #1423, opened 2019-08-13
Added keywords for minimizing and fullscreening window

Harri:
Two new keywords for managing browser window size. Happy to add tests if there is a good idea how to implement those. :)

Maybe something like this:

    Maximize Window
        Set Window Size    200    400
        ${width}    ${height}=    Get Window Size
        Maximize Browser Window
        ${max width}    ${max height}=    Get Window Size
        Verify Window Size Does Not Equal
        ...    ${width}    ${height}    ${max width}    ${max height}

    Minimize Window 
        Set Window Size    200    400
        ${width}    ${height}=    Get Window Size
        Minimize Browser Window
        ${min width}    ${min height}=    Get Window Size
        Verify Window Size Does Not Equal
        ...    ${width}    ${height}    ${min width}    ${min height}

    Fullscreen Window
        Set Window Size    200    400
        ${width}    ${height}=    Get Window Size
        Fullscreen Browser Window
        ${full width}    ${full height}=    Get Window Size
        Verify Window Size Does Not Equal
        ...    ${width}    ${height}    ${full width}    ${full height}

Where Verify Window Size Does not Equal would be

    Verify Window Size Does Not Equal
        [Arguments]    ${orig width}    ${orig height}    ${new width}    ${new height}
        Should Not Be Equal As Numbers    ${orig width}    ${new width}
        Should Not Be Equal As Numbers    ${orig height}    ${new height}

### DIFF
--- a/atest/acceptance/windows.robot
+++ b/atest/acceptance/windows.robot
@@ -195,6 +195,40 @@ Select Window With Delay By Special Locator
     Select Window    main
     Title Should Be    Click link to show a popup window
 
+Maximize Window
+    [Documentation]    Maximizing the window does not change the window size
+    ...    in headlesschrome.
+    [Tags]    known issue headlesschrome
+    Set Window Size    301    302
+    ${width}    ${height}=    Get Window Size
+    Maximize Browser Window
+    ${max width}    ${max height}=    Get Window Size
+    Verify Window Size Does Not Equal
+    ...    ${width}    ${height}    ${max width}    ${max height}
+
+Minimize Window
+    [Documentation]    Minimizing the window does not change the window size
+    ...    in headlesschrome.
+    [Tags]    known issue headlesschrome
+    Set Window Size    301    302
+    Maximize Browser Window
+    ${width}    ${height}=    Get Window Size
+    Minimize Browser Window
+    ${min width}    ${min height}=    Get Window Size
+    Verify Window Size Does Not Equal
+    ...    ${width}    ${height}    ${min width}    ${min height}
+
+Fullscreen Window
+    [Documentation]    Fullscreening the window does not change the window size
+    ...    in headlesschrome.
+    [Tags]    known issue headlesschrome
+    Set Window Size    301    302
+    ${width}    ${height}=    Get Window Size
+    Fullscreen Browser Window
+    ${full width}    ${full height}=    Get Window Size
+    Verify Window Size Does Not Equal
+    ...    ${width}    ${height}    ${full width}    ${full height}
+
 *** Keywords ***
 Open Popup Window, Select It And Verify
     [Arguments]    ${window_id}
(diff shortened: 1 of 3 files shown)


Harri:
I'm assuming that the CI failure is not due the my changes, rather something else. These changes should not affect "test_no_libraries" test in any way, which is the failing one.

aaltat:
So you want to expose Selenium `minimize_window` and `fullscreen_window` as keywords in the SeleniumLibrary. I think it is good idea and changes will certainly fit in the next release, which is SeleniumLibrary 4.0.  

All keywords should have test. Unit testing is not needed because keywords are just wrappers for Selenium API, but acceptance test are required.  Something that you did write in the first message should good enough. Have you seen the [README](https://github.com/robotframework/SeleniumLibrary/blob/master/atest/README.rst) for the acceptance tests? It explains how test can run locally and where they should be placed.

For the CI failure, this PR is reason for the failing [test](https://github.com/robotframework/SeleniumLibrary/blob/master/utest/test/api/test_plugins.py#L25). Every time keywords count is modified that test fails. It is to ensure that plugin API does not interfere with the keyword discovery. But perhaps that test should be modified something like:
`self.assertTrue(len(sl.get_keyword_names()) > 100)`

Harri:
Headless Chrome keeps the window size the same when maximize, minimize of fullscreen is called. Headless Firefox changes the window size. Both browsers work just fine non-headless. Should I add the tests even though it will cause failures with headlesschrome? Should I add the tests with 

    [Documentation]    Fullscreening the window does not change the window size
    ...    in headlesschrome.
    [Tags]    known issue headlesschrome

or something totally different?

aaltat:
Hmmm, that feels like a bug in the Chrome or ChromeDriver side. It would be great if you could report the problem in ChromeDriver project side.

Yes writing test is required and adding tag like in your example will make the test runner to mark test as noncritical. CI runs only with Chrome, but uses headless and normal mode to run tests. Also adding documentation to the test is a good idea.

aaltat:
There was error with pypy and Python 2: 
```
FAIL: Acceptance.Windows.Minimize Window
WebDriverException: Message: unknown error: failed to change window state to minimized, current state is normal
  (Session info: chrome=76.0.3809.100
```
I did retrigger the jobs and lets see the results. Have you seen similar result with Python 2? It is little bit strange, because the Chrome, ChromeDriver and Selenium versions are the same and only thing that changes is the Python version. And Python version should not affect in this case but for some odd reason it does affect.

Harri:
I ran the tests only with py3.6 before pushing, but now I ran

    Minmaxfull Window
        FOR    ${index}    IN RANGE    1    100
            Minimize Browser Window
            Maximize Browser Window
            Fullscreen Browser Window
        END

for few times with py2.7.15 and headlesschrome and it always passed.

aaltat:
Did you happen to `headlesschrome` or `chrome` as a browser? The failing test tests in Travis uses `chrome` as a browser (my assumption for the Python version was wrong, it was the browser type).

aaltat:
Just wanted to say that I am aiming to push out the release candidate at the end of the this week or early next week. After release candidate is made, new features (like this one) are delayed in the next release. What is your preference, would you like to see this in the SL 4.0 or is it OK to delay to the release after SL 4.0?

Harri:
It's ok to delay.

aaltat:
Do you still plan to work with the pull request or can we close the PR?

Harri:
Let's close this. If I ever feel like working on this again, I'll consider re-opening this or creating a new one. Thanks.

aaltat:
Closed by request.

PR #1434, opened 2019-08-27
Complex example for creating new Open Browser keywords and using SL f…

aaltat:
…unctionality


aaltat:
It looks like I did something which is not compatible with Python 2. Need to run tests locally and see what it's.

PR #1440, opened 2019-09-05
Use Python 3.7 in Travis instead of Python 3.7 dev

aaltat:


aaltat:
Need to switch to Travis bionic ( Ubuntu 18.04 LTS ) before this can be done.

PR #1441, opened 2019-09-06
Updating Travis to use bionic image

aaltat:
To move Python 3.7 and 3.8


aaltat:
I am really starting to hate Jython, it doesn't work with Selenium on never images. I am so waiting 1.1.2020 so that I can delete Python 2 and Jython from Travis.

Good news is that once Jython is gone, moving to never images is pretty straight forward.

PR #1460, opened 2019-09-27
Feature/support chromium apps

JonKoser:
Adds a keyword to allow users to open up Chromium-based applications easier.

To open up an application instead of a normal browser, you can use:
`| Open Chromium App | /path/to/app | remote-debugging-port`

The remote debugging port is how ChromeDriver interacts with the application. See:
https://bitbucket.org/chromiumembedded/cef/wiki/UsingChromeDriver

### DIFF
--- a/src/SeleniumLibrary/keywords/browsermanagement.py
+++ b/src/SeleniumLibrary/keywords/browsermanagement.py
@@ -254,11 +254,11 @@ class BrowserManagementKeywords(LibraryComponent):
                           alias=None, desired_capabilities=None,
                           options=None, service_log_path=None):
         """
-        Similar to Open Browser except opens a specified chromium_based application
+        Similar to Open Browser except opens a specified chromium_app application
         instead of a browser. The desired_capabilities, remote_url, alias,
         options, and service_log_path arguments are the same and are all optional.
 
-        ``app_binary`` is the path to the desired chromium_based desktop
+        ``app_binary`` is the path to the desired chromium_app desktop
         application. This includes Qt-based applications with QtWebEngine content
         from Qt versions >= 5.6.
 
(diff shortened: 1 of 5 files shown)


aaltat:
Have you looked at the enhanced Open Browser keyword in SeleniumLibrary 4.0 release? By a quick glance of the PR, it looks to me that there is better support in the latest release.

JonKoser:
@aaltat 
Yup, I looked at the 4.0 release. It's technically do-able, it just takes some wrangling with options. I acknowledge as much in my issue #1461. Maybe there's something else I'm missing? Anyway, in lieu of something a little cleaner, this change just makes life a little easier so I figured I might as well share.

aaltat:
Well, this should work:
`| Open Browser | URL | Chrome | options=binary_location=/path/to/binary;add_argument("remote-debugging-port=debug_port") |` with zero code changes.

JonKoser:
Yeah, good call. However, I really don't like having to go to a URL upon opening. It technically "works" but causes my application to flash whatever random website I put in there first, then redirect to the actual web content it's supposed to display. Perhaps an easy solution would be to just make the URL argument optional and not force users to immediately go to a URL?

JonKoser:
Ah, about:blank works fine.
Thanks!

pekkaklarck:
Should `url` have `about:blank` as the default? Usage would then be
```
Open Browser    browser=Chrome    options=...
```

Also, it's great to see enhanced `Open Browser` working so well!

aaltat:
It gets sometimes asked that keyword should not always force navigation to a URL. We could change it so that if URL is false, we do not perform the navigation.

pekkaklarck:
Perhaps `None` rather than `False`. And it could be the default.

aaltat:
Yes, anything that goes through from `is_noney` is probably better.

pekkaklarck:
This starts to a bit off topic, but once you drop Python 2, you should also drop RF 3.0 to be able to use type hints (explicit and implicit based on defaults). No need for `is_noney` afterwards!

More on topic, I think it would be good to submit an issue about `url=None`. Might be a good idea to submit another issue, or perhaps re-open this one, about documenting how to use SL with Chromium apps. Even though no code changes are required, the usage doesn't seem to be obvious.

JonKoser:
Definitely sounds good to me. I'll write that issue when I get some time today.

PR #1462, opened 2019-09-27
Native click action

JonKoser:
Allows a user to click an element using a native (ActionChain) click without the need to specify coordinates. Keyword is:

`| Click Element Native | <locator> |`


JonKoser:
Probably don't want to add a new keyword. Cancelling PR until a better solution is decided upon. See issue #1463

PR #1467, opened 2019-09-29
Add warning for Python 2

aaltat:


pekkaklarck:
I don't think anyone sees that. When installing with pip, setup.py isn't used at all. Even if it was, someone would only see it if they had enabled Python warnings . I'm fairly certain everyone who knows how to do that is already on Python 3...

PR #1468, opened 2019-09-29
Add methot to replace Python format for PY2 and PY3

### LINKED ISSUE #1452: If output dir contains multiple { and } characters pairs, then opening browser with Firefox will fail.
If robot is run with command: `robot -d {foo}_dir path/to/suite` then `Open Browser` keyword will fail with Firefox on: `KeyError: 'foo'`. This is because Python [format](https://github.com/robotframework/SeleniumLibrary/blob/895c1d69d0165b9d4e59ecf73c4902f2fda73314/src/SeleniumLibrary/keywords/webdrivertools/webdrivertools.py#L288) is used. Same problem also applies to [Capture Page Screenshot](https://github.com/robotframework/SeleniumLibrary/blob/895c1d69d0165b9d4e59ecf73c4902f2fda73314/src/SeleniumLibrary/keywords/screenshot.py#L133). 

Build a wrapper that handle formatting for Python 2 and Python 3. For Python 3: https://docs.python.org/3/library/stdtypes.html#str.format_map is nice. For Python 2 build a something like: https://stackoverflow.com/questions/17215400/python-format-string-unused-named-arguments

aaltat:
Fixes #1452

### DIFF
--- a/atest/acceptance/keywords/screenshots.robot
+++ b/atest/acceptance/keywords/screenshots.robot
@@ -53,12 +53,12 @@ Set Screenshot Directory set back to previous value
 Capture page screenshot with unique index
     [Setup]    Remove Directory    ${OUTPUTDIR}/screenshot-and-index    recursive
     ${file1} =    Capture Page Screenshot    ${OUTPUTDIR}/screenshot-and-index/other-{index}-name.png
-    ${file2} =    Capture Page Screenshot    ${OUTPUTDIR}/screenshot-and-index/some-other-name-{index}.png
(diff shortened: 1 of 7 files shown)


aaltat:
1) Needs more testing with different kinds of combinations. Like `{ind@ex}` or `{ind ex}` or `{index}-{index}`.
2) Helper method is in wrong module, because it is needed also in [WebDriverCreator](https://github.com/robotframework/SeleniumLibrary/blob/master/src/SeleniumLibrary/keywords/webdrivertools/webdrivertools.py#L35).

Issue #1463, opened 2019-09-27
Should have an ActionChain click for Click Element keyword


JonKoser:
As it is now, the only way to use a native ActionChain-based click on an element is by using the `Click Element At Coordinates` keyword. It would be nice to not need to specify coordinates and just do a native click on the center of a given element. This has been implemented in PR#1462.

aaltat:
Hmmm, click at coordinates keyword has a special purpose, but I don't quite understand what problem you are trying to solve. Could you example clarify the why `Click Element` keyword is not suitable for your needs?

JonKoser:
My problem is that, for some reason due to asynchronous calls and promises, I'm unable to click on elements in my Chromium-based application using selenium's <element>.click() function. It gives me this error:
`unknown error: unhandled inspector error: {"code":-32000,"message":"Result of the evaluation is not a promise"} (Session info: content shell=)`

Unfortunately, that's the method that `Click Element` uses. The way I've been able to get around it though is by calling a click action using ActionChains. The only keyword that uses the action chain method is `Click Element At Coordinates`.

pekkaklarck:
It would be good to investigate why that happens. Perhaps there's a bug somewhere that could be fixed. If not, there still are other solutions than adding a new keyword:
- Change `Click Element` and possibly also other `Click` keywords to use action chains.
- Change them to use action chains if `click()` fails.
- Add an option like `action_chain=False` to make this behavior configurable.

JonKoser:
That's fair. In lieu of finding the cause of the failure, since that might take some time, is there one of those options you think might be a good stop-gap? I feel like using action chains if `click()` fails or adding the option are both good routes. I see no reason to change everything to use action chains. I'll close my PR until this is decided.

JonKoser:
I have a branch where I created a helper function which centralizes click actions (like in `Click Button`, `Click Link`, `Click Image`, `Click Element`). It first tries to click the web element with `web_element.click()`. If that attempt fails with a `WebDriverException` it then tries to click the element with an ActionChain click. Seems to work pretty darn well. If you like that approach I'd be happy to create a PR based off that. Otherwise, I'd be happy to try something else.

aaltat:
[...]

PR #1469, opened 2019-09-30
Added action-chain click option

JonKoser:
This PR introduces an option to `Click Element` which enables a user to use Selenium ActionChains to issue a click action on an element instead of using the <web_element>.click() function. This enables a more "native" click of any element in the UI. This is in response to issue #1463

### DIFF
--- a/atest/acceptance/keywords/click_element.robot
+++ b/atest/acceptance/keywords/click_element.robot
@@ -40,7 +40,7 @@ Click Element Action Chain
     [Tags]    NoGrid
     [Documentation]
     ...    LOB 1:1 INFO        Clicking 'singleClickButton' using an action chain.
-    ...    LOG 2:5 DEBUG GLOB: *actions {"actions": [{"type": "pointer",*
+    ...    LOG 2:5 DEBUG GLOB: *actions {"actions": [{*
     Click Element    singleClickButton      action_chain=True
     Element Text Should Be    output    single clicked
 
--- a/atest/acceptance/keywords/click_element.robot
+++ b/atest/acceptance/keywords/click_element.robot
@@ -37,7 +37,9 @@ Double Click Element Error
     Double Click Element    id:öööö
 
 Click Element Action Chain
-    [Documentation]   LOG 2:5 DEBUG GLOB: *actions {"actions": [{"type": "pointer",*
+    [Documentation]
+    ...    LOB 1:1 INFO        Clicking 'singleClickButton' using an action chain.
+    ...    LOG 2:5 DEBUG GLOB: *actions {"actions": [{"type": "pointer",*
     Click Element    singleClickButton      action_chain=True
     Element Text Should Be    output    single clicked
 
--- a/atest/acceptance/keywords/click_element.robot
+++ b/atest/acceptance/keywords/click_element.robot
@@ -38,6 +38,7 @@ Double Click Element Error
 
 Click Element Action Chain
     [Documentation]   Clicks an element using the 'action_chain' argument without a modifier
+    ...               LOG 2     REGEXP: Clicking '.*' using an action chain.
     Click Element    singleClickButton      action_chain=True
     Element Text Should Be    output    single clicked
 
--- a/atest/acceptance/keywords/click_element.robot
+++ b/atest/acceptance/keywords/click_element.robot
@@ -37,6 +37,7 @@ Double Click Element Error
     Double Click Element    id:öööö
 
 Click Element Action Chain
+    [Tags]    NoGrid
     [Documentation]
     ...    LOB 1:1 INFO        Clicking 'singleClickButton' using an action chain.
     ...    LOG 2:5 DEBUG GLOB: *actions {"actions": [{"type": "pointer",*
--- a/atest/acceptance/keywords/click_element.robot
+++ b/atest/acceptance/keywords/click_element.robot
@@ -37,7 +37,7 @@ Double Click Element Error
     Double Click Element    id:öööö
 
 Click Element Action Chain
-    [Documentation]   LOG 2     REGEXP: Clicking '.*' using an action chain.
+    [Documentation]   LOG 2:5 DEBUG GLOB: *actions {"actions": [{"type": "pointer",*
     Click Element    singleClickButton      action_chain=True
     Element Text Should Be    output    single clicked
 
--- a/atest/acceptance/keywords/click_element.robot
+++ b/atest/acceptance/keywords/click_element.robot
@@ -37,8 +37,7 @@ Double Click Element Error
     Double Click Element    id:öööö
 
 Click Element Action Chain
-    [Documentation]   Clicks an element using the 'action_chain' argument without a modifier
-    ...               LOG 2     REGEXP: Clicking '.*' using an action chain.
+    [Documentation]   LOG 2     REGEXP: Clicking '.*' using an action chain.
     Click Element    singleClickButton      action_chain=True
     Element Text Should Be    output    single clicked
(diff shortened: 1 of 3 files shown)


src/SeleniumLibrary/keywords/element.py
```diff
@@ -586,16 +586,29 @@ def click_element(self, locator, modifier=False):
         [https://seleniumhq.github.io/selenium/docs/api/py/webdriver/selenium.webdriver.common.keys.html#selenium.webdriver.common.keys.Keys.ALT|ALT key]
         . If ``modifier`` does not match to Selenium Keys, keyword fails.
 
+        The ``action_chain`` argument can be set to True to use an ActionChain
+        based click instead of the <web_element>.click() function. In this case,
+        an action chain is created which locates the element, moves the mouse to
+        that element, then clicks the mouse.
```
aaltat: Needs explanation about order of arguments, which one overrules which one. Now the implementation is done that `action_chain` is applied before `modifier` argument. But I think arguments are applied in wrong order. If `action_chain` and `modifier` are defined with valid values, then `modifier` should be applied and `action_chain` be ignored.

src/SeleniumLibrary/keywords/element.py
```diff
@@ -586,16 +586,29 @@ def click_element(self, locator, modifier=False):
...
         The ``modifier`` argument is new in SeleniumLibrary 3.2
         """
         if is_falsy(modifier):
-            self.info("Clicking element '%s'." % locator)
-            self.find_element(locator).click()
+            if is_falsy(action_chain):
+                self.info("Clicking element '%s'." % locator)
+                self.find_element(locator).click()
+            else:
+                action = ActionChains(self.driver)
```
aaltat: I would move this logic to an own method.

aaltat (changes requested):
Needs also testing, in addition to comments. Most likely easiest way is to write few acceptance tests and use status checker for verifying correct messages are displayed in the logs.

JonKoser:
Added tests and addressed comments

src/SeleniumLibrary/keywords/element.py
JonKoser: Good call

src/SeleniumLibrary/keywords/element.py
JonKoser: Fix in

aaltat:
The test are good start but they need to look deeper under the hood. Now Test ensures that click happens, but tests do not ensure that click happens by using action chains. Usually in this kind of situation, it is easiest to use [statuschecker](https://github.com/robotframework/statuschecker/) to verify that logs from Selenium are from actions chains and not from normal `click()`. Example [Choose File](https://github.com/robotframework/SeleniumLibrary/blob/master/atest/acceptance/keywords/choose_file.robot#L35) keyword uses it for same purposes.

JonKoser:
Tests guarantee log output now

atest/acceptance/keywords/click_element.robot
```diff
@@ -36,6 +36,12 @@ Double Click Element Error
     [Setup]    Go To Page "javascript/click.html"
     Double Click Element    id:öööö
 
+Click Element Action Chain
+    [Documentation]   Clicks an element using the 'action_chain' argument without a modifier
+    ...               LOG 2     REGEXP: Clicking '.*' using an action chain.
```
aaltat: Better but not quite on what I had in my mind. Test are done, among other things, to safe guard maintainers from stupid mistakes. Now I can change the action chains implementation and it's not always certain that test would capture the accidental change. 

The test now ensures that keyword logs by using the Robot Framework API (which is relevant too) and some sort of click happens, but it doesn't ensure that action chains click happens. 

Please add more checking that keyword calls the correct Selenium API methods.

Same comment applies also to the test in below.

atest/acceptance/keywords/click_element.robot
```diff
@@ -36,6 +36,12 @@ Double Click Element Error
     [Setup]    Go To Page "javascript/click.html"
     Double Click Element    id:öööö
 
+Click Element Action Chain
+    [Documentation]   Clicks an element using the 'action_chain' argument without a modifier
```
aaltat: Remove the documentation saying what the test does, it doesn't bring value and only adds maintenance burden. If you want, test name could say: `Click Element With Action Chains`

atest/acceptance/keywords/click_element_modifier.robot
```diff
@@ -42,6 +42,13 @@ Click Element Wrong Modifier
     ...    ValueError: 'FOOBAR' modifier does not match to Selenium Keys
     ...    Click Element    Button    Foobar
 
+Click Element Action Chain and modifier
+    [Documentation]   Clicks an element with both a valid modifier and action_chain argument. Modifier should should take precedent
```
aaltat: Please remove the documentation saying what the test does, it doesn't bring value and only adds maintenance burden. If you want, test name could say: `In Click Element Modified Argument Must 
Overrule Action Chains argument`

src/SeleniumLibrary/keywords/element.py
```diff
@@ -586,18 +586,37 @@ def click_element(self, locator, modifier=False):
...
+        are valid, the click will be performed using ``modifier`` and ``action_chain``
+        will be ignored.
+
         Example:
         | Click Element | id:button | | # Would click element without any modifiers. |
         | Click Element | id:button | CTRL | # Would click element with CTLR key pressed down. |
         | Click Element | id:button | CTRL+ALT | # Would click element with CTLR and ALT keys pressed down. |
+        | Click Element | id:button | action_chain=True | # Clicks the button using an action chain |
 
         The ``modifier`` argument is new in SeleniumLibrary 3.2
```
aaltat, line 601: Add line here saying: ``action_chain`` is new in SeleniumLibrary 4.1

aaltat (changes requested):
Better but needs few minor enhancements.

atest/acceptance/keywords/click_element.robot
JonKoser: True. Unfortunately, ActionChains click() and WebElement.click() register the same JavaScript MouseEvent action so I can't tell them apart that way. I'll see what I can do but if you have any ideas, please let me know!

JonKoser:
Pushed up documentation changes. Still trying to figure out how to tell if it was an "action chain click" or a "web element click"

atest/acceptance/keywords/click_element.robot
JonKoser: I'm really at a loss for what I can do with this. I've tried a whole bunch of different things with mockito's spy, spy2, patch, etc. and I can't seem to figure out something that'll work. Do you have any guidance here? The trouble comes from the ActionChains object being created and acted upon only in the scope of `_click_with_action_chain()` - it makes it very difficult to say what was called on that object and what wasn't.

aaltat:
No worries I can take a look tomorrow evening.

aaltat:
Sorry, I did have busy day and I can look today. I will try to find time as soon as possible. Thank yo for being patient.

aaltat (approved):

aaltat:
If Travis shows green after the builds this is ready to be merged.

aaltat:
Looks like Chrome and Travis are not playing together and builds are failing. Need to fix Travis first.

aaltat:
It looks like I broke something with Python 2, need to investigate closely.

aaltat:
Thank you for the PR.

JonKoser:
Thank you for helping out with this!

Issue #1464, opened 2019-09-28
Users should not be required to define a URL to Open Browser keyword.


JonKoser:
Currently, the `Open Browser` keyword requires users to pass a URL to navigate to upon opening the browser.  However, in some cases that may not be desired and the user might simply want to open the browser. (as in the case of Chromium-based applications). 

Users should not need to specify this argument and the URL should be `None` by default. If `None` or not specified, navigation shouldn't occur.

I don't have a PR for this yet but can have one as soon as I can get back to my computer. Also sorry about the title "pay" - autocorrect :)

PR #1470, opened 2019-09-30
Optional URL in Open Browser

JonKoser:
Makes the URL in the `Open Browser` keyword optional. This means that you can just open a browser without immediately navigating to a URL. In response to issue #1464


aaltat (changes requested):
Code is looking good but needs test.

JonKoser:
Added a unit test where `Open Browser` is called without any arguments. No .robot acceptance tests added since I noticed that there were no other default value `Open Browser` tests in that suite.

utest/test/keywords/test_keyword_arguments_browsermanagement.py
```diff
@@ -31,3 +31,8 @@ def test_open_browser(self):
         alias = self.brorser.open_browser(url, alias='None',
                                           remote_url=remote_url)
         self.assertEqual(alias, None)
+
```
aaltat, line 35: Please write a new test, because this is testing totally different thing than the original test.

utest/test/keywords/test_keyword_arguments_browsermanagement.py
```diff
@@ -31,3 +31,8 @@ def test_open_browser(self):
...
+        when(self.brorser)._make_driver('firefox', None,
+                                        None, False, None, None).thenReturn(browser)
+        alias = self.brorser.open_browser()
+        self.assertEqual(alias, None)
```
aaltat: Instead of writing this assert, something like this could be better:
```
from mockito import verify, ANY

verify(brorser, times=0).get(ANY)
```

Written without testing the code, so errors might be present, Please let me know, if you want me to try out provide more accurate example.

aaltat (changes requested):
The unit test is a good start, but it needs to look deeper what gets called from the WebDriver.

utest/test/keywords/test_keyword_arguments_browsermanagement.py
JonKoser: Just for clarity's sake - you want me to 1) change the assertEqual into the mokito verify and 2) add another test?

aaltat: 1) Yes, 
2) I was unclear. sorry. I want to that new part you added to do into a new unit test. So that it is not appended to the existing test.

JonKoser: Okidoke - that should do it :)

JonKoser:
New unit test to check that `get()` is never called when no URL is specified

aaltat:
Looking pretty good from a mobile screen. I need to take another look from a bigger screen to make sure that I did not miss anything.

aaltat (approved):

aaltat:
This is ready to be merged. Thank you for the contribution.

I will take a look your other PR during the weekend.

aaltat:
Darn that close button is too near Comment button.

JonKoser:
Thanks for helping out with this!

Issue #1461, opened 2019-09-27
Update Open Browser keyword documentation to better demonstrate how to use Selenium options


JonKoser:
Selenium options is quite complicated system and it seems that Open Browser keyword documentation could be enhanced to better demonstrate how Selenium options can be used. Use real example with Chromium based app.

JonKoser:
Re-opening issue as discussed in PR#1460 (now closed) comments. 

`Open Browser` by itself actually works quite well for Chromium-based apps. However, the documentation for that usecase could be significantly improved. Providing examples of how to use options to pass a binary location and remote debugging port would be quite helpful.

JonKoser:
Re-opening issue as discussed in PR#1460 (now closed) comments. 

`Open Browser` by itself actually works quite well for Chromium-based apps. However, the documentation for that usecase could be significantly improved. Providing examples of how to use options to pass a binary location and remote debugging port would be quite helpful.

aaltat:
Done in #1471

PR #1471, opened 2019-09-30
Chromium-based app example

JonKoser:
Adds an example of how to open up an application built with the Chromium Embedded Framework instead of an internet browser. In response to issue #1461


aaltat (commented):
Looks pretty good, I need to read it from bigger screen.

JonKoser:
Do you want me to put in a line break?

aaltat:
No, reading PR from mobile screen is difficult. Screen size is just too small to see.

aaltat:
I did modify the documentation a little bit. But I usually make lots of typos when using GitHub online editor and therefore proof reading would be nice.

Issue #1482, opened 2019-10-10
"Table Should Contain" does not work properly


Pawci3oo:
Example code:
*** Settings ***
Library    SeleniumLibrary    #4.0.0

*** Variables ***
${url}    https://www.w3schools.com/html/html_tables.asp
${customer_table}    //table[@id='customers']

*** Test Cases ***
Sample test
    Open Browser    ${url}    browser=chrome
    Maximize browser window
    Element Should Be Visible    ${customer_table}
    Table Should Contain    ${customer_table}    Each

## Expected behavior and actual behavior

Expected: FAIL
Actual: PASS

If there is given text on the page and even I provide ID of table for this keyword, then this keyword returns True. (table does not contain text)
`Table Should Contain` is useless in that case because I can also use `Page Should Contain` to find text on the page. In my project, I have two tables on the page where one record is moved to the second table and I need to verify that.

## Environment

Browser: Chrome 77.0.3865.90
Browser driver: ChromeDriver 2.42.591088
Operating System: Windows 10 17763.740
Libraries 
- Robot Framework: 3.1.2
- Selenium: 3.141.0
- SeleniumLibrary: 4.0.0
- Interpreter: Python 3.7.4

aaltat:
Please update the ChromeDriver, it is pretty old compared to the Chrome and try does it resolve the problem.

Pawci3oo:
No changes.
Current version: ChromeDriver 77.0.3865.40

aaltat:
I can reproduce the problem, need to investigate more closely where the problem is.

aaltat:
Found the bug, it is in [table element finder](https://github.com/robotframework/SeleniumLibrary/blob/master/src/SeleniumLibrary/keywords/tableelement.py#L197) and Selenium documentation [explains](https://selenium-python.readthedocs.io/api.html#selenium.webdriver.remote.webelement.WebElement.find_elements_by_xpath) why it happens. Good catch, it has been there since 3.0 release.

Pawci3oo:
I found that bug because I did upgrade my framework (RF and python). I also changed Selenium2Library to SeleniumLibrary and a lot of my tests are failing since this upgrade.

When I can expect a fix? :)

aaltat:
Already fixed locally. Bug also revealed that that there is many other bugs in the same class and I am fixing all of those.

Pawci3oo:
I understand. Anyway thanks :)

[...]

PR #1485, opened 2019-10-10
Fixed table *  should contains keywords

aaltat:
Since SeleniumLibrary 3.0 release, `Table Should Contain`, `Table Header Should Contain` and `Table Footer Should Contain` have searched text not only from table but from the whole page. 

This PR fixes the problem, described in #1482

### DIFF
--- a/src/SeleniumLibrary/keywords/tableelement.py
+++ b/src/SeleniumLibrary/keywords/tableelement.py
@@ -194,13 +194,13 @@ class TableElementKeywords(LibraryComponent):
                                  % (locator, expected))
 
     def _find_by_content(self, table_locator, content):
-        return self._find(table_locator, '//*', content)
+        return self._find(table_locator, 'xpath:.//*', content)
 
     def _find_by_header(self, table_locator, content):
-        return self._find(table_locator, '//th', content)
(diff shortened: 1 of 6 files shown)


src/SeleniumLibrary/keywords/tableelement.py
```diff
@@ -194,13 +194,13 @@ def table_should_contain(self, locator, expected, loglevel='TRACE'):
...
-        return self._find(table_locator, '//*', content)
+        return self._find(table_locator, 'xpath:.//*', content)
 
     def _find_by_header(self, table_locator, content):
-        return self._find(table_locator, '//th', content)
+        return self._find(table_locator, 'xpath:.//th', content)
 
     def _find_by_footer(self, table_locator, content):
-        return self._find(table_locator, '//tfoot//td', content)
+        return self._find(table_locator, 'xpath:.//tfoot//td', content)
```
pekkaklarck, line 203: If support for chained locators is added so that it can be used also externally, not only internally, `.//xxx` should be recognized as an xpath expression automatically.

aaltat: Good point and perhaps `.//xxx` should be recognized as xpath. Should be easy to add, but in separate PR.

PR #1486, opened 2019-10-12
Fix close browser not delete alias

### LINKED ISSUE #1416: Close Browser does not delete alias
Code
 
```
*** Settings ***
Library           SeleniumLibrary
*** Test Cases ***

Open browser

    Open browser   https://github.com/  alias=zombie
    Go to    http://google.se
    Close Browser

Switch to browser
    Switch browser    zombie
    Go to  http://google.fi
```

Expected: Switch browser should fail.
Actual: Switch browser pass and Go to fail.

Close browser leaves the alias of the browser as a valid session resulting in weird behavior when using switch browser

yo-ga:
I make a workaround to remove the index in the alias cache. If there is a better way to get the alias of the current index, please let me know.

Fix #1416 

Signed-off-by: Jia-wei Yu <art96962006@gmail.com>

### DIFF
--- a/atest/acceptance/open_and_close.robot
+++ b/atest/acceptance/open_and_close.robot
@@ -35,10 +35,9 @@ Switch to closed browser is possible
     Close Browser
     Switch Browser    Browser 3
     Page Should Contain    Name:
-    Switch Browser    Browser 2
     Run Keyword And Expect Error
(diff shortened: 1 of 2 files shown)


aaltat:
Please fix the failing test.

Issue #1500, opened 2019-10-28
'Handle Alert' keyword treats all exceptions as timeout


Zeckie:
## Steps to reproduce the issue

If the browser crashes or the window is closed the keyword 'Handle Alert' is waiting for an alert to appear, the error "Alert not found in (time)" is raised, where (time) is the timeout configured, but not reached. For example, if timeout is 2 minutes the error shows as `Alert not found in 2 minutes.`  despite the keyword failing after only a few seconds.

I've created a demo at: https://github.com/Zeckie/AlertBug/tree/master/RobotProject

## Error messages and additional information
From test DelayedClose (source https://github.com/Zeckie/AlertBug/blob/master/RobotProject/Alerts.robot)
```
Alert not found in 2 minutes. 
Traceback (most recent call last):
  File "C:\Users\Zeckie\AppData\Local\Programs\Python\Python37\Lib\site-packages\SeleniumLibrary\__init__.py", line 467, in run_keyword
    return DynamicCore.run_keyword(self, name, args, kwargs)
  File "C:\Users\Zeckie\AppData\Local\Programs\Python\Python37\Lib\site-packages\SeleniumLibrary\base\robotlibcore.py", line 102, in run_keyword
    return self.keywords[name](*args, **kwargs)
  File "C:\Users\Zeckie\AppData\Local\Programs\Python\Python37\Lib\site-packages\SeleniumLibrary\keywords\alert.py", line 117, in handle_alert
    alert = self._wait_alert(timeout)
  File "C:\Users\Zeckie\AppData\Local\Programs\Python\Python37\Lib\site-packages\SeleniumLibrary\keywords\alert.py", line 138, in _wait_alert
    % secs_to_timestr(timeout)) 
```

## Expected behavior and actual behavior

I expect the test to fail, but for the error message to not be misleading. Ie. it should say that the window was closed (or that it couldn't communicate with the browser).

## Environment

Browser: Chrome 78
Browser driver: chromedriver 77
Operating System: Windows 10 
Eclipse: 2019-09
Libraries
- Robot Framework: 3.1.2
- Selenium: 3.141.0
- SeleniumLibrary: 4.1.0
- Interpreter: Python 3.7.4

I'm working on a fix - will submit a pull request soon.

aaltat:
I think I understand what you are saying, but I need to also see the results from your examples. I will try to find time to do it, when I have reviewed #1497

aaltat:
[...]

PR #1501, opened 2019-10-28
Handle errors other than Timeout in keyword 'Handle Alert'

Zeckie:
Handle TimeoutException separately in AlertKeywords._wait_alert
Raise the current exception only for TimeoutException and add new except block for WebDriverException.
https://github.com/robotframework/SeleniumLibrary/issues/1500


aaltat:
I understand what you have done and I can see the reasoning behind. Only thing to do is to write some test for the chance. What you think, is this easy to test in the unit or in the acceptance tests

src/SeleniumLibrary/keywords/alert.py
```diff
@@ -133,6 +133,9 @@ def _wait_alert(self, timeout=None):
...
         try:
             return wait.until(EC.alert_is_present())
-        except WebDriverException:
+        except TimeoutException:
             raise AssertionError('Alert not found in %s.'
-                                 % secs_to_timestr(timeout))
+                             % secs_to_timestr(timeout))
+        except WebDriverException as e:
+            raise AssertionError('An exception occurred waiting for alert: %s'
+                                 % str(e))
```
pekkaklarck: `str(e)` fails if the error contains non-ASCII characters on Python 2. Good news is that `str()` can just be removed and it ought to work fine.

`e` alone perhaps looks a bit odd and could be renamed to `err` at the same time. I think that's used more in SL code base.

aaltat:
It looks like something broken in the alert handling and test in CI fails. Could you take a look and fix the issue.

aaltat:
In some of the environments there are constant failures:
```
FAIL: Acceptance.Keywords.Alerts.Handle Alert when popup window closes
Expected error 'STARTS:An exception occurred waiting for alert' but got 'Alert not found in 1 second
```
Unfortunately the log.html is not saved anywhere from the Travis runs.

I wonder is this a timing issue and would this be easier to test in the unit level.

Zeckie:
> Unfortunately the log.html is not saved anywhere from the Travis runs.
> 
> I wonder is this a timing issue and would this be easier to test in the unit level.

Yes, it would be much easier to troubleshoot if we could see the log.html 
I agree that it looks like a timing issue - will try increasing the timeout & put a teardown so the new test failing won't cause other tests to fail.

Zeckie:
oops, didn't mean to close

src/SeleniumLibrary/keywords/alert.py
Zeckie: I've made those changes in commit https://github.com/robotframework/SeleniumLibrary/pull/1501/commits/d61a8030d313bb2f697e07c4334e7992ba9570e3

aaltat:
Fixed one minor thing and then this ready to go. Thank you for the contribution.

PR #1508, opened 2019-11-02
Prevent Input Password to reveal password from Selenium logs

### LINKED ISSUE #1454: Input Password keyword should also prevent password being visible from the Selenium logging
When Robot Framework log level is high enough the password is visible from the Selenium level logging. Something like this is usually seen: 
```
DEBUG | POST http://127.0.0.1:51593/session/0e87ad26dbe3d303d8b7937309367b4a/element/4beff912-9198-4dad-9d0d-46f844ddd3eb/value {"text": "FooBar", "id": "4beff912-9198-4dad-9d0d-46f844ddd3eb", "value": ["F", "o", "o", "B", "a", "r"]}
```
SeleniumLibrary  keyword should hide such entry being visible output files of Robot Framework. SeleniumLibrary will not prevent the value being visible in the keyword argument, because that is logged by Robot Framework core.

aaltat:
Fixes #1454

### DIFF
--- a/src/SeleniumLibrary/keywords/formelement.py
+++ b/src/SeleniumLibrary/keywords/formelement.py
@@ -16,6 +16,8 @@
 
 import os
 
+from robot.libraries.BuiltIn import BuiltIn
+
 from SeleniumLibrary.base import LibraryComponent, keyword
 from SeleniumLibrary.errors import ElementNotFound
 from SeleniumLibrary.utils import is_noney, is_truthy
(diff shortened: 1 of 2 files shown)


src/SeleniumLibrary/keywords/formelement.py
```diff
@@ -231,16 +233,15 @@ def input_password(self, locator, password, clear=True):
...
+        the TRACE level and tests must not be executed using level below
+        DEBUG if the password should not be logged in any format.
 
-        The `clear` argument is new in SeleniumLibrary 4.0
+        The `clear` argument is new in SeleniumLibrary 4.0. Hiding password
+        logging from Selenium logs is new in SeleniumLibrary 4.2.
         """
         self.info("Typing password into text field '%s'." % locator)
-        self._input_text_into_text_field(locator, password, clear)
+        self._input_text_into_text_field(locator, password, clear, True)
```
pekkaklarck, line 244: Passing Boolean values (or None) as-is is not that good style. Better to name the argument like `disable_logging=True`.

src/SeleniumLibrary/keywords/formelement.py
```diff
@@ -421,8 +422,15 @@ def _get_value_from_radio_buttons(self, elements):
                 return element.get_attribute('value')
         return None
 
-    def _input_text_into_text_field(self, locator, text, clear):
+    def _input_text_into_text_field(self, locator, text, clear, disable_log):
```
pekkaklarck, line 425: `clear` and `disable_log` should have default values. They provide kind of type hints and at least the latter wouldn't need to be always passed.

aaltat:
Comments fixed in: https://github.com/robotframework/SeleniumLibrary/pull/1510

PR #1511, opened 2019-11-10
Document and test options with Windows paths

### LINKED ISSUE #1473: Open Browser keyword and Selenium options with Windows path needs double escaping
If there is a need to do add windows path by using `Open Browser` keywords `options` argument, then windows path needs double escaping. Example: 
```
| Open Browser | url | firefox | options=add_argument("-profile"); add_argument("C:\\\\path\\to\\\\profile")
```
This is both untested and not documented. Write few tests and improve the documentation.

aaltat:
Fixes #1473


src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -177,7 +178,16 @@ def open_browser(self, url=None, browser='firefox', alias=None,
...
         `add_argument("--headless")`. But `add_argument(" --headless ")` is
         not same same as `add_argument ( "--headless" )`, because
-        spaces inside of quotes are not removed.
+        spaces inside of quotes are not removed. Please note that if
+        options string contains backslash, example a Windows OS path,
+        the backslash needs escaping both in Robot Framework data and
+        in Python side. This means single backslash must be writen using
+        four backslash characters. Example, Windows path:
+        "C:\\path\\to\\profile" must be written as
+        "C:\\\\\\\\path\\\\\\to\\\\\\\\profile". Another way to write
```
pekkaklarck, line 187: `\\\\\\\\` looks somewhat strange here in source code but is probably right. Prefixing this docstring with `r` would allow using `\\\\` instead.

pekkaklarck: Could also mention that most of the time `/` works as a path separator also on Windows.

aaltat: Good idea, I will try it out.

aaltat: For the first comment, it looks really strange, but I generated kw doc's and it did render correctly.

PR #1537, opened 2020-01-19
Executable path for Open Browser keyword

### LINKED ISSUE #1536: Add possibility to configure executable_path in the Open Browser keywords
The `executable_path` is used to configure webdriver executable, like chromedriver path, when webdriver instance is created. This is useful feature example [webdriver manager](https://github.com/rasjani/webdrivermanager) and for cases when altering `PATH` is not easy.

aaltat:
Fixes #1536


aaltat:
Missing documentation, then should be ready to merge.

Issue #1559, opened 2020-03-10
"Wait Until Location Is" error message reads "Location did not is <url> in <wait time>"


testventure:
Steps to reproduce the issue:
Give "Wait Until Location Is" a parameter that will fail.

Observed: Location did not is (url) in (wait time)
Expected: Something like:
- Location did not become (url) in (wait time)
- Location did not change to (url) in (wait time)

aaltat:
Good catch, the error message definitely needs improving. Would you like to make a PR to fix it?

testventure:
Sure I'm willing to do that.

aaltat:
Fixed in #1560

PR #1560, opened 2020-03-10
fix to the error message for wait until location is

testventure:


aaltat:
There is a test [Wait Until Location Is Fails](https://github.com/robotframework/SeleniumLibrary/blob/master/atest/acceptance/keywords/location.robot#L75) which tests the error message. That should be also changed.

aaltat:
Fixes #1559

aaltat:
Thank you for the fix.

PR #1608, opened 2020-06-12
Use correct directory for screenshots when EMBED is default

osrjv:
It's possible to set EMBED as the default option for screenshots and override it in a per keyword basis to actual file paths. It worked correctly by accident with absolute paths, because of how os.path.join() works, but created an 'EMBED' folder with relative paths.

Note: The acceptance test documentation fields seemed to contain some XML, but wasn't sure what the purpose was. Left it empty for now.


aaltat:
Fixes #1614

aaltat:
Thank you for the contribution.

PR #1666, opened 2020-10-01
Improve type hints

aaltat:


src/SeleniumLibrary/keywords/alert.py
```diff
@@ -52,7 +53,7 @@ def alert_should_be_present(
         self,
         text: str = "",
         action: str = ACCEPT,
-        timeout: Union[float, str, None] = None,
+        timeout: Optional[timedelta] = None,
```
pekkaklarck, line 56: `Optional[]` is unnecessary when there's `None` as a default value. Even MyPy doesn't need it unless you run it with `--no-implicit-optional`:
https://mypy.readthedocs.io/en/stable/command_line.html?highlight=optional#cmdoption-mypy-no-implicit-optional

src/SeleniumLibrary/keywords/alert.py
```diff
@@ -76,7 +78,7 @@ def alert_should_be_present(
 
     @keyword
     def alert_should_not_be_present(
-        self, action: str = ACCEPT, timeout: Union[float, str, None] = 0
+        self, action: str = ACCEPT, timeout: Union[timedelta] = 0
```
pekkaklarck, line 81: This is wrong because an argument with type `timedelta` cannot really have an integer as a default. You probably also meant `Optional[timedelta]` not `Union[timedelta]` (but I'd leave `Optional[]` away as well).

src/SeleniumLibrary/keywords/alert.py
mkorpela: I think Optional is best practice if None is used. Types should not contain None as part of the type and if they do, then that should *explicitly* be stated.
See for example https://en.wikipedia.org/wiki/Void_safety for more details on this subject.

mkorpela: And the default is also changing in mypy https://github.com/python/mypy/issues/9091

src/SeleniumLibrary/keywords/alert.py
aaltat: Hmmm, what would be suitable default? The easiest would be to change it to `None` but that would be somewhat backwards incompatible (Because it would default to library default timeout.) Would: 
```
def alert_should_not_be_present(self, action: str = ACCEPT, timeout = timedelta(seconds=0);
```
be better?

Issue #1653, opened 2020-09-25
Drag And Drop does not work with event firing


sebastianciupinski:
# Steps to reproduce the issue

```*** Setting ***
Library    SeleniumLibrary

*** Variables ***
${SELENIUM_SPEED}		1

*** Test Cases ***

A
    Open Browser   http://the-internet.herokuapp.com/  gc     options=add_argument("--ignore-certificate-errors")
    Set Selenium Speed     ${SELENIUM_SPEED}
    Capture Page Screenshot
    Click Link     Dropdown
    Go Back
    Click Link     Drag and Drop
    ${a}=       Get Webelement     column-a
    ${b}=       Get Webelement     column-b
    Drag And Drop    ${a}    ${b}
    Drag And Drop    ${b}    ${a}
    Close Browser
```

Such code works, slowly and requires cursor to be in browser window, but works
${a} and ${b} are `selenium.webdriver.remote.webelement.WebElement`

While having the same code with

```
Library    SeleniumLibrary    event_firing_webdriver=${CURDIR}${/}SomeScript.py
```

code does not work, because Drag And Drop requires webelement object, while having event firing driver, ${a} and ${b} are `selenium.webdriver.support.event_firing_webdriver.EventFiringWebElement` and result is: `AttributeError: move_to requires a WebElement`

```
robotframework==3.2.2
robotframework-seleniumlibrary==4.5.0
selenium==3.141.0
chrome 85.0.4183.121 on windows 10
```

aaltat:
This is actually bug in Selenium side. But we have been making workarounds for similar type of problems also in other places in SeleniumLibrary code  so we could fix it. 

As a quick fix, you could try out Selenium 4 pre releases. In there this problem should be fixed. Please remember that Selenium is still in Alpha phase, so it might not work correctly always, but is worth of try.

aaltat:
Fixed in #1679

PR #1679, opened 2020-11-03
Initial commit of additional patching for selenium 3 move_to bug.

emanlove:
First, I'll note my workaround is slightly different. Instead of catching attribute errors,
I pass in both the driver and the element(s) checking first if this is indeed an
EventFiringWebDriver and not Selenium 4 or greater. I did this because I didn't see the other
try/catch method till after I wrote this. I can modify these to go that path. My one concern is
that in all the additional places discovered which are affected I was worried that we might
need multiple try/catch within a single method or one a large block of code. The other concern
was that we may get a AttributError for other reasons.

I added in acceptance tests (but no unit tests yet) for the additional changes. I copied pasted
into the current selenium_move_to_workaround.robot test suite. This feels a little hackish to me
and I would like to either copy paste into multiple test suites under a folder (still hackish).
Or what I would really like to do is to use the existing test suites and accomplish this with
tags and some mechanism to run with a differnt open browser setup. This may be too complicated
for now (mostly in time to figure out that solution) so open for ideas here ..

I have not yet tested with selenium 4 due to a bug I encountered when running the acceptance tests.
Need to look at that now.

Welcome any feedback...


src/SeleniumLibrary/utils/events/event.py
```diff
@@ -15,9 +15,30 @@
...
+        https://github.com/SeleniumHQ/selenium/issues/7467
+        https://github.com/SeleniumHQ/selenium/issues/6604
+
+    """
+    if not isinstance(ef_driver,EventFiringWebDriver) or selenium_major_version() >= 4:
+        return elements
+    unwrapped_elements = ef_driver._unwrap_element_args(elements)
+    return unwrapped_elements
+
+def selenium_major_version():
```
aaltat, line 41: The distutils [LooseVersion](https://github.com/python/cpython/blob/master/Lib/distutils/version.py#L269) could be handy.

src/SeleniumLibrary/utils/events/event.py
```diff
@@ -15,9 +15,30 @@
...
+        References:
+        https://github.com/SeleniumHQ/selenium/issues/7877
+        https://github.com/SeleniumHQ/selenium/pull/8348
+        https://github.com/SeleniumHQ/selenium/issues/7467
+        https://github.com/SeleniumHQ/selenium/issues/6604
+
+    """
+    if not isinstance(ef_driver,EventFiringWebDriver) or selenium_major_version() >= 4:
+        return elements
+    unwrapped_elements = ef_driver._unwrap_element_args(elements)
```
aaltat: There is also [public method](https://selenium-python.readthedocs.io/api.html#selenium.webdriver.support.event_firing_webdriver.EventFiringWebElement.wrapped_element) and same for the [driver](https://selenium-python.readthedocs.io/api.html#selenium.webdriver.support.event_firing_webdriver.EventFiringWebDriver.wrapped_driver). But do not know is there need to unwrap the driver.

aaltat (commented):
I like your idea with code, but there are few things in the unwrap method which could be enhanced.

src/SeleniumLibrary/utils/events/event.py
emanlove: I was trying to avoid having to require another Python package for something temporary. Right now it doesn't look like we require setuptools (which it appears LooseVersion now a part of) in either the core or the library. Although I though there might be something out there I went with this solution because of that. It does look like LooseVersion is in c-python but had not checked iron python or any other build (and don't recall which ones are supported still).

src/SeleniumLibrary/utils/events/event.py
emanlove: I came across that initially and then lost it till I saw the fix you had in there as I tested. One thing about this private method I liked was that it could handle a single element or a list of elements. I wasn't sure if any of the keywords I was patching had the possibility of multiple elements. Two other things I worried about. One was that there could be other reasons for an AttributeError exception (have no definitive reasons but also no way to prove that only this issue could cause one). The other concern was having too many try catch to pick up all the lines in the element.py file.

I'm not unwrapping the webdriver but just the webelements, using the private method on the EventFiringWebDriver. I see I could possibly just bring the element(s) and see if there instance of event firing webelement version. I guess I brought in the driver to have a way check the instance is a event firing web element.

src/SeleniumLibrary/utils/events/event.py
aaltat: Next version, 5.0, will support Python 3 and PyPy 3 only. No need to worry about other interpreters. But it is just an idea and your code works as well.

src/SeleniumLibrary/utils/events/event.py
aaltat: Usually action chains work with single element at the time. Therefore it would be safer to use the public attribute.

aaltat:
About writing acceptance tests. I guess we could change this [import](https://github.com/robotframework/SeleniumLibrary/blob/master/atest/acceptance/resource.robot#L2) to have EventFiringWebDriver. This would run all the acceptance test with it.

src/SeleniumLibrary/utils/events/event.py
emanlove: Sounds good. I will go back and take a look at this.

emanlove:
> 
> 
> About writing acceptance tests. I guess we could change this [import](https://github.com/robotframework/SeleniumLibrary/blob/master/atest/acceptance/resource.robot#L2) to have EventFiringWebDriver. This would run all the acceptance test with it.

I was looking to see if there is a way to run the same tests suites with different configurations. I would think running all tests with eventfiringwebdriver would be testing something different and thus not the best. I was just curious if there was a model for running similar or the same suite but externalize the configuration so once it runs with and another time without.

aaltat:
> > About writing acceptance tests. I guess we could change this [import](https://github.com/robotframework/SeleniumLibrary/blob/master/atest/acceptance/resource.robot#L2) to have EventFiringWebDriver. This would run all the acceptance test with it.
> 
> I was looking to see if there is a way to run the same tests suites with different configurations. I would think running all tests with eventfiringwebdriver would be testing something different and thus not the best. I was just curious if there was a model for running similar or the same suite but externalize the configuration so once it runs with and another time without.

How about setting it with variable and let CI run the tests?

emanlove:
Using a variable and CI may be a solution.. will take a look at that.

src/SeleniumLibrary/utils/events/event.py
emanlove: Switched over to the public attribute. Looking at the acceptance test now ..

emanlove:
> How about setting it with variable and let CI run the tests?

Tried that with something like the following
```robotframework
*** Settings ***
Documentation    Can be deleted when minimum Selenium version 4.0
Library          SeleniumLibrary     event_firing_webdriver=${event_firing_or_none}
Resource         resource_event_firing_webdriver.robot
Force Tags       NoGrid
...
```
but it won't recognize ${CURDIR} if I run with
```
python atest\run.py --nounit firefox --suite=selenium_move_to_workaround -v event_firing_or_none:${CURDIR}/MyListener.py
```
but drawing a blank on getting it to accept and read `${CURDIR}` as a variable not just the string "${CURDIR}"

aaltat:
> > How about setting it with variable and let CI run the tests?
> 
> Tried that with something like the following
> 
> ```
> *** Settings ***
> Documentation    Can be deleted when minimum Selenium version 4.0
> Library          SeleniumLibrary     event_firing_webdriver=${event_firing_or_none}
> Resource         resource_event_firing_webdriver.robot
> Force Tags       NoGrid
> ...
> ```
> 
> but it won't recognize ${CURDIR} if I run with
> 
> ```
> python atest\run.py --nounit firefox --suite=selenium_move_to_workaround -v event_firing_or_none:${CURDIR}/MyListener.py
> ```
> 
> but drawing a blank on getting it to accept and read `${CURDIR}` as a variable not just the string "${CURDIR}"

I think Robot Framework variables are not resolved from command line. At least ${CURDIR} is a special variables which exists only in passing time. 

But I think we could enhance the run.py and have a command line argument that sets the variable correctly and resolves the path in Python side. Something like 
`python atest\run.py --nounit firefox --suite=selenium_move_to_workaround --evet-firing-webdriver`

emanlove:
> But I think we could enhance the run.py and have a command line argument that sets the variable correctly and resolves the path in Python side. Something like
> `python atest\run.py --nounit firefox --suite=selenium_move_to_workaround --evet-firing-webdriver`

I'll explore that some more. Thanks for the feedback!

aaltat:
> > But I think we could enhance the run.py and have a command line argument that sets the variable correctly and resolves the path in Python side. Something like
> > `python atest\run.py --nounit firefox --suite=selenium_move_to_workaround --evet-firing-webdriver`
> 
> I'll explore that some more. Thanks for the feedback!

Similar as the [--nounit](https://github.com/robotframework/SeleniumLibrary/blob/master/atest/run.py#L284-L289).

emanlove:
I like the idea but wondering if that should be explored in a different pull request?

aaltat:
> I like the idea but wondering if that should be explored in a different pull request?

it can be, the only thing that I am interested is that CI stays green. For me one PR can contain multiple fixes and it should not be that long that it is daunting to read. This PR is still pretty short.

emanlove:
Pushed up a new argument for the test runner. Plan on cleaning up tests as a result. Hoping to remove the copied selenium move to work around and use the main set using CI and arguments.

emanlove:
@aaltat Thinking about this last night can statuschecker work on different expected logs based upon runtime conditions/arguments?

aaltat:
> @aaltat Thinking about this last night can statuschecker work on different expected logs based upon runtime conditions/arguments?

Well, not easily. 

1) Use regex. Done already in few tests. 
2) Do not run the tests, if they are not important 
2.5) Duplicate test and write different check. And run the test which is meaningful.

emanlove:
> 2.5) Duplicate test and write different check. And run the test which is meaningful.
This is what the selenium on move workaround tests were doing which made sense. But with expanding it to a much larger group of duplicate tests I was hoping to avoid the duplication by using different run time conditions. But that conflicts with the usage of statuschecker ..

aaltat:
Regex can do wonders too, perhaps looking in to that direction too?

emanlove:
I've noticed where the issue was and fixed a few items. I also learned more about this system of checks and how to review them and download the results. If I need to make any changes please go ahead and share that. Have a great day!

aaltat:
Looking good from mobile screen, I  need to take a look from bigger screen. Is there reason for draft status in the PR or is this ready for merge?

emanlove:
Originally I did it as a draft merge because I wasn't sure whether or not this idea would be a good one. As for merging do I need to add anything to the changes log or elsewhere? Other than that I think it is good to go.

aaltat:
If all looks good from bigger screen,  the only thing is for me to press merge.

aaltat:
Fixes #1653

emanlove:
Kiitos!

PR #1698, opened 2021-03-09
Update webdrivertools.py

remontees:
Support [Edge webdriver for Linux](https://developer.microsoft.com/fr-fr/microsoft-edge/tools/webdriver/).

### DIFF
--- a/src/SeleniumLibrary/keywords/webdrivertools/webdrivertools.py
+++ b/src/SeleniumLibrary/keywords/webdrivertools/webdrivertools.py
@@ -285,7 +285,7 @@ class WebDriverCreator:
         remote_url,
         options=None,
         service_log_path=None,
-        executable_path="MicrosoftWebDriver.exe",
+        executable_path="msedgedriver",
     ):
         if remote_url:
             defaul_caps = webdriver.DesiredCapabilities.EDGE.copy()


aaltat:
This breaks one of the unit tests, please fix it.

remontees:
> This breaks one of the unit tests, please fix it.

For sure, because the test is already expecting the .exe file instead of the cross-platform executable: https://github.com/robotframework/SeleniumLibrary/blob/6025235f6b24b4843a753dc90fb4c2a814c27227/utest/test/keywords/test_webdrivercreator.py#L475

idxn:
@remontees Why do we need to update the code? Can you just specify the executable_path argument in the Open browser keyword as an example below?
Open Browser | http://example.com | Edge | executable_path=msedgedriver
-- | -- | -- | --

aaltat:
Because then it by default will work with larger community.

remontees:
> Because then it by default will work with larger community.

@idxn 
That's it: if you try to open an Edge web browser in your test in a Linux environment, you would expect it to work (or say that the **LINUX** webdriver has not been installed), not that the Windows executable is missing.
In the meantime, thks for the trick.

emanlove:
Rechecked pull request against updated unit/acceptance tests [with passing results](https://github.com/emanlove/robotframework-seleniumlibrary/actions/runs/3212001693).

PR #1709, opened 2021-03-22
Type convert fix

### LINKED ISSUE #1708: "Element Attribute Should Have Value" no longer supports ${None}
## Steps to reproduce the issue

```
*** Settings ***
Library     SeleniumLibrary    timeout=30   implicit_wait=5   run_on_failure=Capture Page Screenshot

*** Test Cases ***
Test
    ${chrome_options} =    Evaluate    sys.modules['selenium.webdriver'].ChromeOptions()    sys
    Call Method    ${chrome_options}    add_argument   --headless
    Create Webdriver  Chrome  my_alias  chrome_options=${chrome_options}  executable_path=/usr/local/bin/chromedriver
    Go To    https://www.google.com
    ${S2L}          Get Library Instance    SeleniumLibrary
    ${WEBDRIVER} =    Set Variable    ${S2L.driver}

    Element Attribute Value Should Be    //img[contains(@alt, "Google")]    madeupattr    ${None}
```

```
==============================================================================
Test
==============================================================================
Test                                                                  | FAIL |
Element '//img[contains(@alt, "Google")]' attribute should have value 'None' but its value was 'None'.
Test                                                                  | FAIL |
1 test, 0 passed, 1 failed
==============================================================================
Output:  /Volumes/Workspace/git/fe/q2smart_automation/output.xml
Log:     /Volumes/Workspace/git/fe/q2smart_automation/log.html
Report:  /Volumes/Workspace/git/fe/q2smart_automation/report.html
```

With some manual testing, it looks like providing the str hint is converting the none to string. I think we should use `Union[None, str]` instead of just str for the expected argument. 

## Expected behavior and actual behavior

I expect this test to pass because Get Element Attribute returns ${None}. I have tests that follow this pattern that have started failing after the latest robotframework-seleinum update.

## Environment

Browser: Chrome Version 89.0.4389.90 (Official Build) (x86_64)
Browser driver: headless chrome=89.0.4389.90
Operating System: Mac 10.14.6
Libraries
- Robot Framework: Version 4.0
- Selenium: Version 3.141.0
- SeleniumLibrary: Version 5.1.1
- Interpreter: Python 3.7.6

aaltat:
Fixes #1708

### DIFF
--- a/src/SeleniumLibrary/keywords/element.py
+++ b/src/SeleniumLibrary/keywords/element.py
@@ -25,6 +25,7 @@ from selenium.webdriver.remote.webelement import WebElement
 
 from SeleniumLibrary.base import LibraryComponent, keyword
 from SeleniumLibrary.errors import ElementNotFound
+from SeleniumLibrary.utils.types import type_converter
 
 
 class ElementKeywords(LibraryComponent):
@@ -414,7 +415,7 @@ class ElementKeywords(LibraryComponent):
         self,
         locator: Union[WebElement, str],
         attribute: str,
-        expected: str,
+        expected: Union[None, str],
         message: Optional[str] = None,
     ):
         """Verifies element identified by ``locator`` contains expected attribute value.
@@ -431,8 +432,9 @@ class ElementKeywords(LibraryComponent):
         if current_expected != expected:
             if message is None:
                 message = (
-                    f"Element '{locator}' attribute should have value '{expected}' but "
-                    f"its value was '{current_expected}'."
+                    f"Element '{locator}' attribute should have value '{expected} "
+                    f"({type_converter(expected)})' but its value was '{current_expected}' "
+                    f"({type_converter(current_expected)})."
                 )
             raise AssertionError(message)
         self.info(
--- a/src/SeleniumLibrary/keywords/element.py
+++ b/src/SeleniumLibrary/keywords/element.py
@@ -55,7 +55,7 @@ class ElementKeywords(LibraryComponent):
     def element_should_contain(
         self,
         locator: Union[WebElement, str],
-        expected: str,
+        expected: Union[None, str],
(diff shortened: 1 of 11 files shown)


src/SeleniumLibrary/keywords/element.py
```diff
@@ -325,7 +326,7 @@ def element_should_not_be_visible(
     def element_text_should_be(
         self,
         locator: Union[WebElement, str],
-        expected: str,
+        expected: Union[None, str],
```
pekkaklarck, line 329: Doesn't this now mean that you cannot test that element text is string `none`, case-insensitively?

This is to some extend related to the open issue about changing `Union` conversion semantics (robotframework/robotframework#3897). If we'd change the logic so that conversion isn't done if a non-string argument has one of the accepted types, using `Union[str, None]` here would allow using `${NONE}` (i.e. real Python `None`) but strings were never converted to `None`. If we'd change the logic so that also strings aren't converted if there's `str` in `Union`, then also the current variant would work.

aaltat: Yes, string `None` can not be tested. But that how it did work when SL did have the custom `is_noney` in place. Also the difficult part is that Selenium will return Python `None` when something is not found. 

Also there is some logic in SL that relies that string None is converted to Python `None ` in some other keywords.

pekkaklarck: Not being able to test string `None` sounds like an annoying limitation to me. Let's discuss on Slack how we can make that work. This will help deciding about robotframework/robotframework#3897 as well.

aaltat: I agree lest discuss in Slack, it is easier. But this is mostly just returning functionality of SeleniumLibrary did work before RF 4 and SL 5.0.

PR #1710, opened 2021-03-23
execute_javascript code type changed to Union

vincenzo-gasparo:
I'm facing an issue after the RF4 release. The keyword `execute_javascript` treats WebElements passed after ARGUMENTS as str, by changing the type declaration `def execute_javascript(self, *code: str) -> Any` in SeleniumLibrary/keywords/javascript.py with `Union[WebElement, str] `the problem seems to be solved


aaltat:
Thank you for the contribution, this will fix #1711

PR #1714, opened 2021-03-24
Headless edge option

Kim-Hietikko:
Headless edge option with msedge selenium tools.

pip install msedge-selenium-tools selenium==3.141 needs to be installed in order it to run.

### DIFF
--- a/src/SeleniumLibrary/keywords/webdrivertools/webdrivertools.py
+++ b/src/SeleniumLibrary/keywords/webdrivertools/webdrivertools.py
@@ -327,6 +327,7 @@ class WebDriverCreator:
         options.use_chromium = True
         options.add_argument("headless")
         options.add_argument("disable-gpu")
+        options.add_argument("no-sandbox")
         return self.create_edge(
             desired_capabilities,
             remote_url,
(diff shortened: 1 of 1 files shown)


rasjani:
Few things to consider:

1. This cannot be merged as-is because it depends on the 3rd party tooling that is not fetched as part of installation and depending on msedge (for now) doesn't sound good approach because ->
2. Even if no import/dependency errors, i'd check first what sort of support upstream selenium provides for handling edge chromium options. However, this would probably only land with Selenium 4. 

However, support for  new chrome based Edge is definitely something that should be available for in SeleniumLibrary.

There are approaches that could be done to support this with SeleniumLibrary plugin interface. Backporting the support for current Selenium/SeleniumLibrary could be done first via plugin, that will inherit original WebDriverCreator class of SeleniumLibrary and override/modify those parts that are needed and then in the plugin's constructor, overwrite `_webdriver_creator` property but I'll get back into this a bit later .. 

As parting words, one of the features i'd like to see in SeleniumLibrary is the ability to modify/inject code into how and what gets instantiated so the issue is close to my aims ..

jyogesh27:
Thanks for sharing

jyogesh27 (approved):

emanlove:
With the current state of Selenium and the SeleniumLibrary, Chromium based Edge, and the End-Of-Life for Microsoft Edge Legacy for which MicrosoftWebDriver.exe is paired with, this pull request is no longer valid. In addition one can get an Headless (Chromium-based) Edge with a Open Browser keyword like

```robotframework
Open Browser  https://robotframework.org/  Edge  options=add_argument("headless")
```

This doesn't mean having a `headlessedge` like we have a `headlessfirefox` and `headlesschrome` might be helpful but that will need to be in a different pull request.

PR #1717, opened 2021-03-30
Fix some inconsistencies on dev tooling

rasjani:
* Add flake8 to dev dependencies
* Add possibility to run unittests via inv
* include tasks.py in lint target
* fixed some typos


tasks.py
```diff
@@ -208,7 +208,20 @@ def atest(ctx, suite=None):
...
+@task
+def utest(ctx, suite=None):
+    """Runs utest/run.py
+
+    Args:
+        suite: Select which suite to run.
+    """
+    command = "python utest/run.py"
+    if suite:
+        command = f"{command} --suite {suite}"
```
rasjani, line 226: doesnt seem to work ..

aaltat: That should work, at least last week it did work

rasjani: Could be pebcak error. But I was running --suite locators and console log was reporting plenty of other tests. Maybe they where skipped but still shown in terminal as I didn't pay attention to status.

rasjani: yeah, it was error between chair and monitor :D

PR #1718, opened 2021-03-30
Support 'data-' attributes as locators

### LINKED ISSUE #1474: Support 'data-' attributes as locators for SeleniumLibrary
The data attribute is a well defined attribute in the HTML5 specification.

It would be nice to have a '_shortcut_' in SeleniumLibrary, so we could use it as a locator. 

**For an Example:**
If we had an element such as: `<div data-automation="helptext">ABC 123</div>`
Instead of using something like: `css:div[data-automation="helptext"]`
or instead of: `//*[@data-automation="helptext"]`

We could do something like this instead: `data:automation:helptext`

_Submitted as per discussion in the seleniumlibrary channel in the RobotFramework Slack workspace._

rasjani:
Adds new locator by data-* id's with syntax:

data:id_name:actual_id
Fixes #1474

### DIFF
--- a/src/SeleniumLibrary/__init__.py
+++ b/src/SeleniumLibrary/__init__.py
@@ -136,6 +136,7 @@ class SeleniumLibrary(DynamicCore):
     | link         | Exact text a link has.              | ``link:The example``           |
     | partial link | Partial link text.                  | ``partial link:he ex``         |
     | sizzle       | Sizzle selector deprecated.         | ``sizzle:div.example``         |
+    | data         | Element ``data-*`` attribute        | ``data:id:my_id``              |
     | jquery       | jQuery expression.                  | ``jquery:div.example``         |
     | default      | Keyword specific default behavior.  | ``default:example``            |
 
@@ -171,6 +172,9 @@ class SeleniumLibrary(DynamicCore):
       the system under test contains the jQuery library.
     - Prior to SeleniumLibrary 3.0, table related keywords only supported
       ``xpath``, ``css`` and ``sizzle/jquery`` strategies.
+    - ``data`` strategy is conveniance locator that will construct xpath from the parameters.
+      If you have element like `<div data-automation="automation-id-2">`, you locate the element via
+      ``data:automation:automation-id-2``. This feature was added in SeleniumLibrary 5.2.0
 
     === Implicit XPath strategy ===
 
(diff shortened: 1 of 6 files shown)


src/SeleniumLibrary/__init__.py
```diff
@@ -171,6 +172,9 @@ class SeleniumLibrary(DynamicCore):
       the system under test contains the jQuery library.
     - Prior to SeleniumLibrary 3.0, table related keywords only supported
       ``xpath``, ``css`` and ``sizzle/jquery`` strategies.
+    - ``data`` strategy is conveniance locator that will construct xpath from the parameters.
+      If you have element like `<div data-automation="automation-id-2">`, you locate the element via
+      ``data:automation:automation-id-2``
```
pekkaklarck: Should add a note like `New in SeleniumLibrary <version>.` or `New in v.<version>.` somewhere.

src/SeleniumLibrary/locators/elementfinder.py
```diff
@@ -220,6 +221,12 @@ def _find_by_tag_name(self, criteria, tag, constraints, parent):
             parent.find_elements(By.TAG_NAME, criteria), tag, constraints
         )
 
+    def _find_by_data_locator(self, criteria, tag, constraints, parent):
+        name, value = criteria.split(":", 2)
+        local_criteria = f'//*[@data-{name}="{value}"]'
+        return self._find_by_xpath(local_criteria, tag, constraints, parent)
+
+
```
pekkaklarck: `name, value = criteria.split(":", 2)` fails with somewhat cryptic ValueError if `criteria` has no `:` in it. Should probably check that (or catch ValueError) and provide a better error to users.

Also unnecessary empty line.

src/SeleniumLibrary/__init__.py
rasjani: Yeah, i was thinking about that as there's other references about when the feature was introduced. However,  need to sync up with @aaltat to get a bit more understanding about releasing.

rasjani: added 5.2.0 as released version, and added it to the gh milestone too ..

rasjani:
tests pass, initial comments addressed.. Anything else ?

PR #1721, opened 2021-04-01
update Edge driver name from legacy to current

skorzhevsky:
According to https://developer.microsoft.com/en-us/microsoft-edge/tools/webdriver/, "MicrosoftWebDriver.exe" is a legacy now, new name is "msedgedriver.exe"


emanlove:
This was done with #1698 and #1795.

PR #1722, opened 2021-04-13
Use RF 4.0.1

aaltat:


rasjani:
Those failures look somewhat alarming @aaltat  .. did you have mechanism in place that aggregates or atleast allows easy access to robot's own logs and not just console on the pr ? zip file is generated  but is it visible anywhere ?

aaltat:
The result are archived in the GItHub actions, see below in this page: https://github.com/robotframework/SeleniumLibrary/actions/runs/743757850 

This is caused by a change in https://github.com/robotframework/statuschecker/releases/tag/v2.0.0 It has backwards incompatible change how it count keywords in tests.

aaltat:
Let see does this fix it https://github.com/robotframework/SeleniumLibrary/pull/1723

PR #1723, opened 2021-04-14
Use robotstatuschecker 2.0

aaltat:


rasjani:
random comment: seeing status on GitHub app feels like a "Killer feature" for GH actions :)

aaltat:
I think I did find bug in statuschekker. There is difference in checker functionality between RF 4.0 and 3.2.2

PR #1727, opened 2021-05-06
Add Wait Until keywords for Textarea

### LINKED ISSUE #1726: Provide support for waiting for a value in a textarea element
# Feature request

While evaluating RobotFramework + SeleniumLibrary I found a use case where
the value of a textarea is changed and I would like to wait until the value matches some expected text.

Generally, you would expect Wait Until Element Contains would fill this scenario. But issue #167 showcases that textareas behave a little differently.

I believe that textareas would need their own Wait Until keyword to handle the .value property they contain.

nathanhannig:
Wait Until Textarea Contains - wait_until_textarea_contains
Wait Until Textarea Does Not Contain - wait_until_textarea_does_not_contain
Wait Until Textarea Value Is - wait_until_textarea_value_is
Wait Until Textarea Value Is Not - wait_until_textarea_value_is_not

Closes #1726

### DIFF
--- a/src/SeleniumLibrary/keywords/waiting.py
+++ b/src/SeleniumLibrary/keywords/waiting.py
@@ -417,6 +417,106 @@ class WaitingKeywords(LibraryComponent):
             error,
         )
 
+    @keyword
+    def wait_until_textarea_contains(
+        self,
+        locator: Union[WebElement, None, str],
+        text: str,
+        timeout: Optional[timedelta] = None,
+        message: Optional[str] = None,
+    ):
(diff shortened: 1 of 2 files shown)


nathanhannig:
Closes #1727 1726

nathanhannig:
Closes #1726

emanlove:
Would like to see if this can be done without adding addition keywords. Otherwise we start to expand the library with keywords for every condition and every element type.

emanlove:
To limit the number of keywords, which for waits there could be many, these keywords can be accomplished by existing keywords including `Wait For Condition`.

Issue #1746, opened 2021-09-30
Keyword `Input Password` throws an exception when run with Python (no robotframework)


mikahanninen:
# For issues
## Steps to reproduce the issue

```python
from SeleniumLibrary import SeleniumLibrary
from time import sleep

library = SeleniumLibrary()

def minimal_task():
    try:
        library.open_browser('https://docs.robocorp.com')
        library.input_password("//input[@aria-label='Search']", "playwright")
        sleep(5)
    finally:
        library.close_all_browsers()
    print("Done.")


if __name__ == "__main__":
    minimal_task()
```

## Error messages and additional information

```shell
Traceback (most recent call last):
  File "task.py", line 23, in <module>
    minimal_task()
  File "task.py", line 15, in minimal_task
    library.input_password("//input[@aria-label='Search']", "playwright")
  File "C:\Users\User\AppData\Local\robocorp\live\199e494e4c733ef3\lib\site-packages\SeleniumLibrary\keywords\formelement.py", line 269, in input_password
    self._input_text_into_text_field(locator, password, clear, disable_log=True)
  File "C:\Users\User\AppData\Local\robocorp\live\199e494e4c733ef3\lib\site-packages\SeleniumLibrary\keywords\formelement.py", line 509, in _input_text_into_text_field
    previous_level = BuiltIn().set_log_level("NONE")
  File "C:\Users\User\AppData\Local\robocorp\live\199e494e4c733ef3\lib\site-packages\robot\libraries\BuiltIn.py", line 3006, in set_log_level
    old = self._context.output.set_log_level(level)
  File "C:\Users\User\AppData\Local\robocorp\live\199e494e4c733ef3\lib\site-packages\robot\libraries\BuiltIn.py", line 64, in _context
    return self._get_context()
  File "C:\Users\User\AppData\Local\robocorp\live\199e494e4c733ef3\lib\site-packages\robot\libraries\BuiltIn.py", line 69, in _get_context
    raise RobotNotRunningError('Cannot access execution context')
robot.libraries.BuiltIn.RobotNotRunningError: Cannot access execution context
```

## Expected behavior and actual behavior

Keyword `Input Password` works with and without Robot Framework

## Environment

Browser: default by Open Browser keyword
Browser driver: Gecko
Operating System: Windows 10
Libraries
- Robot Framework: NOT USED
- Selenium: 3.141.0
- SeleniumLibrary:  robotframework-seleniumlibrary==5.1.3
[...]

PR #1747, opened 2021-10-05
Fixed failure when inputting text and RF is not running

Harri:
Fixes issue mentioned in https://github.com/robotframework/SeleniumLibrary/issues/1746


emanlove:
Isn't this issue also in other places where we try to log and robot is not running? I ask because i am wondering if the issue should be resolved at a higher level in a singular spot for all the library.

emanlove:
Closing as won't implement. If one feels this is still an issue please open a new ticket.

PR #1749, opened 2021-10-26
Strip and collapse spaces from element text should be

jannek76:
Sometimes element has spaces that can be accepted for the test case. Specially leading and trailing spaces can occur in element.

These functionalities has added to [BuiltIn Should Be Equal](https://robotframework.org/robotframework/latest/libraries/BuiltIn.html#Should%20Be%20Equal) keywords in RF 4.0 and RF 4.1 and would be nice to have in SeleniumLibrary too.

### DIFF
--- a/src/SeleniumLibrary/keywords/element.py
+++ b/src/SeleniumLibrary/keywords/element.py
@@ -13,12 +13,14 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import re
+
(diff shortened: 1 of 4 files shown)


src/SeleniumLibrary/keywords/element.py
```diff
@@ -329,6 +331,8 @@ def element_text_should_be(
         expected: Union[None, str],
         message: Optional[str] = None,
         ignore_case: bool = False,
+        strip_spaces: Union[bool, str] = False,
```
rasjani, line 334: why this is union and collapse_spaces is just bool  ?

Issue #1535, opened 2020-01-08
Add API to set page load timeout


0xLeon:
Today, I stumbled upon the lack of ability to set the page load timeout defined by the W3C WebDriver standard [1]. The underlying Python Selenium implementation does provide a corresponding `set_page_load_timeout` method in `selenium.webdriver.remote.webdriver.WebDriver` [2].

Too me, it is not completely clear how this should be handled on the library level. SeleniumLibrary currently has two kinds of timeouts. The library timeout is both the library-own default timeout for the waiting keywords as well as the script timeout from the WebDriver standard. The implicit wait timeout is only a direct pass-through to the implicit wait timeout from the WebDriver standard. I don't know exactly why the WebDriver standard has a separate page load timeout, but it should be exposed on the library level as well, I think.

The question is, should SeleniumLibrary just use the existing library-level timeout for the page load timeout as well? This would mean that you basically loose the separation on the WebDriver level and have a single configuration parameter for script timeout and page load timeout.  
The other solution would be to add another set of parameters and keywords for the page load timeout itself. This way, the separation from the WebDriver level is also exposed on the library level. But it would probably mean an API-breaking change is necessary, adding another parameter to the library `__init__` as well as some more keywords specifically for the page load timeout.

[1] https://w3c.github.io/webdriver/#timeouts
[2] https://github.com/SeleniumHQ/selenium/blob/5d0099de8401c3e5b58f58bd227c46d95da7c8a7/py/selenium/webdriver/remote/webdriver.py#L967

0xLeon:
One more note: I'd be happy to provide a pull request for either solution, but I can't say for sure when this would be possible due to the necessary code contribution approval process from my employer.

aaltat:
Page load timeout is (usually) bigger than the other timeouts in the library. Usually current timeouts are less then 60 seconds (more close to 15 seconds in my usual use) and page load timeout can be multiple of minutes. Of course this highly depends on the web application performance and numerous other factors. 

At least we should add keyword to set the page load timeout, in similar functionality [Set Selenium Timeout](https://robotframework.org/SeleniumLibrary/SeleniumLibrary.html#Set%20Selenium%20Timeout). But should page load timeout be also library level import? It is question which does not have clear answer. Page load timeout is not needed that often than other timeouts, but for consistency reason it would be good to have in the library init. I would say that consistency is good thing in this matter and lets add it in the library init too. 

[...]

PR #1754, opened 2021-12-06
 Add API to set page load timeout (#1535) 

robinmatz:
Hi there,

this PR addresses [issue 1535](https://github.com/robotframework/SeleniumLibrary/issues/1535).

The issue requested there be an api to set selenium page load timeout.

Currently, SeleniumLibrary supports 
  - a global timeout, which is used mostly as a poll for assertions
  - implicit_wait, which is the time to wait for locators

In addition to that, the selenium project offers a page load timeout, which is the amount of time to wait for page to complete loading before an error is thrown.

As @aaltat argued

> At least we should add keyword to set the page load timeout, in similar functionality Set Selenium Timeout. But should page load timeout be also library level import? It is question which does not have clear answer. Page load timeout is not needed that often than other timeouts, but for consistency reason it would be good to have in the library init. I would say that consistency is good thing in this matter and lets add it in the library init too. [1]

Therefore, I created both the init argument as well as two new keywords, ``Set Selenium Page Load Timeout`` and ``Get Selenium Page Load Timeout``.

Note that this is my first contribution to this project, and while I tried to follow the instructions with due diligence, there are most certainly things that I have forgotten or misunderstood.

So please reach out in case there are any questions or suggestions.

Regards,
Robin

[1] https://github.com/robotframework/SeleniumLibrary/issues/1535#issuecomment-572144557

### DIFF
--- a/src/SeleniumLibrary/__init__.py
+++ b/src/SeleniumLibrary/__init__.py
@@ -355,6 +355,17 @@ class SeleniumLibrary(DynamicCore):
 
     See `time format` below for supported syntax.
 
+    == Page load ==
+    Page load timeout is the amount of time to wait for page load to complete until error is raised.
+
+    The default page load timeout can be set globally
+    when `importing` the library with the ``page_load_timeout`` argument
+    or by using the `Set Selenium Page Load Timeout` keyword.
+
+    See `time format` below for supported timeout syntax.
+
+    Support for page load is new in SeleniumLibrary 6.1
+
     == Selenium speed ==
 
     Selenium execution speed can be slowed down globally by using `Set
@@ -431,6 +442,7 @@ class SeleniumLibrary(DynamicCore):
         self,
         timeout=timedelta(seconds=5),
         implicit_wait=timedelta(seconds=0),
+        page_load_timeout=timedelta(seconds=10),
         run_on_failure="Capture Page Screenshot",
         screenshot_root_directory: Optional[str] = None,
         plugins: Optional[str] = None,
@@ -442,6 +454,8 @@ class SeleniumLibrary(DynamicCore):
           Default value for `timeouts` used with ``Wait ...`` keywords.
         - ``implicit_wait``:
           Default value for `implicit wait` used when locating elements.
+        - ``page_load_timeout``:
+          Default value to wait for page load to complete until error is raised.
         - ``run_on_failure``:
           Default action for the `run-on-failure functionality`.
         - ``screenshot_root_directory``:
@@ -456,6 +470,7 @@ class SeleniumLibrary(DynamicCore):
         """
         self.timeout = _convert_timeout(timeout)
         self.implicit_wait = _convert_timeout(implicit_wait)
+        self.page_load_timeout = _convert_timeout(page_load_timeout)
         self.speed = 0.0
         self.run_on_failure_keyword = RunOnFailureKeywords.resolve_keyword(
             run_on_failure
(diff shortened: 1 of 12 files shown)


MaximeAbout:
Hi guys, I was looking on how to change the default timeout on a Go To keyword and found this MR, I'm really interested, if it can be merged please. Thx !

amitlohar:
Is there any update on this? With the recent docker updates, our very 1st (initial) test case is going into hang state, where the url takes couple of minutes to load.
The test case just hangs on Open browser call.
I think, this proposed keyword is really required and should be part of RF since we do not have a work around for this.

src/SeleniumLibrary/__init__.py
```diff
@@ -431,6 +442,7 @@ def __init__(
         self,
         timeout=timedelta(seconds=5),
         implicit_wait=timedelta(seconds=0),
+        page_load_timeout=timedelta(seconds=10),
```
robinmatz: @emanlove , should this be placed at the bottom for backwards compatibility reasons?

robinmatz:
@emanlove , I've just updated this PR to resolve the merge conflicts and to match version 6.1. Looking forward to the code review :wink:

robinmatz:
Just fixed 2 failing tests. Now all tests pass without failure.

I am thinking of a way to also add some atests for this.

emanlove:
I'll take a look. I want to merge this is in and make a release this week.

robinmatz:
I just updated the acceptance tests for this. The pipeline fails because apparently, the following tests suites depend on a browser being opened already.

I thought, the tests accounted for that now, but I still seem to be missing something.

robinmatz:
@emanlove I was thinking about changing two things:
1. Move the new import parameter `page_load_timeout` to the bottom of the exiting ones to keep backwards compatibility
2. Increase the default page load timeout to 30 seconds because this was the default with selenium from what I have seen in the execution logs for Open Browser.

What do you think?

emanlove:
1. I like the idea of moving that to the bottom. This added library import was what was causing me some hesitation.  In the next release the initial plan is to include a generic wait for condition so thus we may have even more timing/waiting based arguments. I was thinking that we might make a single argument (here in the near future / 6.2 release) that is a dictionary of waits and timing values. So I wasn't sure about adding two more library import arguments and then changing it again. That is really a me issue and not a code issue. I did ask @yuriverweij  and @lisacrispin and both expressed or asked about backwards compatibility. So you have resolved that and if we do change in the future we can make a announced depreciation/change if needed.

2. 30 seconds sounds good as a default. Honestly not sure what I would have defaulted to so using selenium as a reference sounds good.

aaltat:
SL starting to have many arguments in the library import and those arguments are positional. If one wants to change only the last one all must be defined. Instead of using positional arguments using named only arguments would be better. Then arguments order doesn’t matter and only arguments which are needed can be defined.

robinmatz:
I totally agree with what Tatu said. However, since deprecating positional import arguments does not seem like a trivial change, I think it's worth to pursue it in a dedicated issue.

For this one, I suggest we leave the new import parameter at the bottom. 

One thing I am not sure about: I just realized that I misread the default page load timeout. It's actually 300000 Ms, i.e. 5 minutes, which seems like a lot. Then again if we were to set a lower value as the default for SeleniumLibrary, might this cause issues in user experience?

emanlove:
@robinmatz ~~Can you add the `Triage` tag to the "Handle Alert when popup window closes" test [as shown here](https://github.com/robotframework/SeleniumLibrary/commit/0f35ba2e3763992d0f4234eeac28fe78726a3e7b)? It looks like the error message has changed with alerts. I don't want to hang up merging this in but also want to see the test results.~~

I am going ahead and going to merge this in. I have checked the tests against a branch were this test is dropped/triaged and all pass. I opened another issue to review that test. For no let's leave that test as is so this issue can stay in the forefront.

PR #1762, opened 2021-12-12
Fix #1459: Add Capture Full Page Screenshot

### LINKED ISSUE #1459: Capture Fullpage Screenshot
This is not a bug, but rather a new feature. 
Often the browser page is longer that is shown and the user needs to scroll down to see the rest of the page. 

Robot Framework often takes a screenshot on failure, but the screenshot captured only shows half of the page and might not include the problematic area. 
It would be nice to have a keyword that captures a whole length image.

This is my python script to implement such functionality 
```
def capture_fullpage_screenshot(url):
    """ Creating a Full-Page screenshot.

    Use this keyword for debugging purposes.
    Creates an image called full_page_screenshot.png.
    """
    options = webdriver.ChromeOptions()
    options.add_argument("start-maximized")
    options.add_argument("enable-automation")
    options.add_argument("--headless")
    options.add_argument("--no-sandbox")
    options.add_argument("disable-infobars")
    options.add_argument("--disable-extensions")
    options.add_argument("--disable-gpu")
    driver = webdriver.Chrome(options=options)
    current_window = driver.get_window_size()

    driver.get(url)

    time.sleep(12)

    height = driver.execute_script("return document.body.scrollHeight")
    driver.set_window_size(1920, height)
    driver.save_screenshot("full_page_screenshot.png")

    driver.set_window_size(current_window["width"], current_window["height"])
```

However, it is not yet properly working, as it does not work inside docker. 
And it has a hardcoded file name.

kumy:
Please find PR to allow capture full page screenshot. It use the functions linked per @aaltat in https://github.com/robotframework/SeleniumLibrary/issues/1459#issuecomment-535932114

This add a new keyword `Capture Full Page Screenshot`. It also add a boolean parameter `full_screen` to  `Capture Page Screenshot`.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -159,7 +159,7 @@ Tests
 When submitting a pull request with a new feature or a fix, you should
 always include tests for your changes. These tests prove that your
 changes work, help prevent bugs in the future, and help document what
-your changes do. Depending an the change, you may need
+your changes do. Depending on the change, you may need
 ``acceptance tests``\ *, ``unit tests``* or both.
 
 Make sure to run all of the tests before submitting a pull request to be
--- a/src/SeleniumLibrary/keywords/screenshot.py
+++ b/src/SeleniumLibrary/keywords/screenshot.py
@@ -64,7 +64,7 @@ class ScreenshotKeywords(LibraryComponent):
         return previous
 
     @keyword
-    def capture_page_screenshot(self, filename: str = DEFAULT_FILENAME_PAGE) -> str:
+    def capture_page_screenshot(self, filename: str = DEFAULT_FILENAME_PAGE, full_screen: bool = False) -> str:
         """Takes a screenshot of the current page and embeds it into a log file.
 
         ``filename`` argument specifies the name of the file to write the
@@ -78,6 +78,10 @@ class ScreenshotKeywords(LibraryComponent):
         is embedded as Base64 image to the log.html. In this case file is not
         created in the filesystem.
 
+        If ``full_screen`` is True, then a full page screenshot will be taken.
+        Only firefox driver support this. On other drivers, this option has
+        no effect.
+
         Starting from SeleniumLibrary 1.8, if ``filename`` contains marker
         ``{index}``, it will be automatically replaced with an unique running
         index, preventing files to be overwritten. Indices start from 1,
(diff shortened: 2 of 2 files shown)


kumy:
Anyone able to review please?

aaltat:
I am not anymore maintaining this library, ping @emanlove But in my opinion having new keyword is not good idea, just have argument for the existing one is better.

kumy:
Thanks for your reply. If there is no new keyword, then is there a way to set the argument to `full_screen=True` on library import? :thinking: 

I was using it like this:
```
Library    SeleniumLibrary    timeout=10    implicit_wait=0    run_on_failure=capture full page screenshot
```

emanlove:
I agree that an additional separate keyword would not be preferred. One problem I see is adding a parameter to the keyword (`Capture Page Screenshot`) that itself is a parameter to the run_on_failure library argument. Tatu has mentioned to me that there is code within BrowserLibrary which allows parameters on parameters within library import arguments. This might be better than the alternative of using the `Register Keyword To Run On Failure` keyword within the script. One question I have there is are we making the library import even more complicated if we do this and have plug-ins?

The other issue is that `get_full_page_screenshot_as_base64` is only a method on the Firefox driver and not on any of the others. There should be some code to protect against that. I know there is desire for full page screenshot across browser but think that should be addressed in a different issue (possibly?).

kumy:
So if I understand correctly,

1. remove the new keyword, we then have to user `Register Keyword To Run On Failure`, should be quite easy.
2. protect new `full_screen` parameter to be used on other driver than Firefox.

For 2., I've not yet any idea how to do that. Do you have existing code example from which I could learn from? Thanks

emanlove:
One technique would be to use the [hasattr function](https://docs.python.org/3/library/functions.html#hasattr) to check and if if the `driver` object has the `get_full_page_screenshot_as_base64` method.

kumy:
@emanlove Hello, I've finally done the requested changes :smile: 

- New keyword removed
- Check function presence or fallback to normal mode.

Could be used like:
```robot
*** Settings ***
Library  SeleniumLibrary  run_on_failure=Capture Full Page Screenshot     # […]

# […]

*** Keywords ***
Capture Full Page Screenshot
    Capture Page Screenshot     full_screen=True
```
Or by using

```robot
Register Keyword To Run On Failure      Capture Full Page Screenshot
```

Thanks for your review

FlyingSpaceCow:
Forgive my ignorance (I am very new to all of this, but have been stumbling through pretty well up until this point). 

I was hoping you could point me to how someone would save the "Capture Full Page Screenshot" as a .png in a desired directory. 

I'm getting the error: ("name used for saved screenshot does not match file ")

I'm not quite sure how to add the argument to tell "Capture Full Page Screenshot" where to save the image and what to name it.

kumy:
Hi, is there anything else I can do? Thanks

aaltat:
@emanlove why the CI did not run?

@kumy tests are needed.

emanlove:
Not sure why the CI tests didn't run. Initial quick review doesn't seem to indicate anything. Doing some more in depth looking ..

hrnaltnts:
Hello 
Currently, do have any solution to take full page screenshots using Selenium Library?

PR #1765, opened 2021-12-30
Switching browsers - Minimize departure session and maximize arrival.

gerodim:
This is a change that enables switching among multiple open browsers when executing this type of test scenarios.
This is particularly important when there are multiple parallel sessions (e.g. automating a scenario where multiple users are logged in in an application and they have to interact).

### DIFF
--- a/src/SeleniumLibrary/keywords/browsermanagement.py
+++ b/src/SeleniumLibrary/keywords/browsermanagement.py
@@ -430,6 +430,11 @@ class BrowserManagementKeywords(LibraryComponent):
         """
         try:
             self.drivers.switch(index_or_alias)
+            handle_of_the_window = self.driver.current_window_handle
+            self.driver.switch_to.window(handle_of_the_window)
+            self.driver.minimize_window()
+            self.driver.set_window_rect(0, 0)
+            self.driver.maximize_window()
         except RuntimeError:
             raise RuntimeError(
                 f"No browser with index or alias '{index_or_alias}' found."


rasjani (commented):
@emanlove  Would it make more sense to have the newly added functionality behind a toggle ?

emanlove:
> Would it make more sense to have the newly added functionality behind a toggle ?

In other words add an argument that the user can choose to do these steps on switching browsers? I think yes. This one in particular gives me hesitation in that it forces the user to do these steps without asking. One can do this already with existing keywords. This is just a convenience that although helpful forces everyone to follow this action as currently implemented.

emanlove:
As I did a quick review of the open pull requests, this is one I was leaning towards rejecting because (1) it forces people to do steps they might not want to do (2) it adds to the overall keyword count when it can be done with existing keywords (3) I am not sure how widely needed or requested it is (again given it can be done with existing keywords).

rasjani:
> As I did a quick review of the open pull requests, this is one I was leaning towards rejecting because (1) it forces people to do steps they might not want to do (2) it adds to the overall keyword count when it can be done with existing keywords (3) I am not sure how widely needed or requested it is (again given it can be done with existing keywords).

My Pov:  Either have this sort of  extra functionality behind a optional keyword argument *or* move the functionality to separate plugin - seleniumtestability definitely could take this - but even there, it would be good to be behind feature flag ..

emanlove:
This is too much of a unique sequence of operation for a broad set of users. As noted above one can use the extension capability to add this functionality if desired.

PR #1782, opened 2022-09-02
Remove Opera

Brownies:
Opera is [no longer supported](https://github.com/SeleniumHQ/selenium/blob/trunk/py/CHANGES) and causes unit tests to fail.
[Chrome limits cookie expiry to max 400 days](https://chromestatus.com/feature/4887741241229312) since version 104 so adjusted the expiry times in cookie tests. 

Also, Selenium version 4.1.4 changes logging slightly and that breaks some of the acceptance tests so I have limited the version to  `<= 4.1.3` for now.


emanlove:
Merged in these changes plus fixes to the acceptance test logging issues in another pull request. I will add you to the changelog for those.

PR #1793, opened 2022-10-09
Remove deprecated rebot option

### LINKED ISSUE #1788: Acceptance tests: rebot option `--noncritical` is deprecated since RF 4
[`--noncritical`](https://github.com/robotframework/SeleniumLibrary/blob/master/atest/run.py#L100) is deprecated and ignored in RF 4 and a warning is issued. The option is completely removed in RF 5 and causes tests to fail.

This doesn't show up in Actions because they only test RF 3 & 4 with Chrome and even the tests tagged `Known Issue Chrome` actually pass.

Brownies:
Fixes #1788 

Also updates `checkout` action to `v3` and `setup-python` action to `v4`.
Actions based on Node12 now emit warnings like
`Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions/checkout, actions/setup-python, actions/setup-python, actions/checkout`

### DIFF
--- a/atest/run.py
+++ b/atest/run.py
@@ -97,8 +97,6 @@ ROBOT_OPTIONS = [
 REBOT_OPTIONS = [
     "--outputdir",
     RESULTS_DIR,
-    "--noncritical",
-    "known issue {browser}",
 ]
 
 
@@ -208,6 +206,7 @@ def execute_tests(interpreter, browser, rf_options, grid, event_firing):
     options.extend([opt.format(browser=browser) for opt in ROBOT_OPTIONS])
     if rf_options:
         options += rf_options
+    options += ["--exclude", f"known issue {browser.replace('headless', '')}"]
     command = runner
     if grid:
         command += [
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -12,9 +12,9 @@ jobs:
         rf-version: [3.2.2, 4.1.3]
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Set up Python ${{ matrix.python-version }} with Robot Framework ${{ matrix.rf-version }}
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
     - name: Start xvfb
--- a/atest/acceptance/windows.robot
+++ b/atest/acceptance/windows.robot
@@ -95,14 +95,14 @@ Get and Set Inner Window Size with Frames
     ...    Set Window Size         ${400}    ${300}    ${True}
 
 Get and Set Window Position
-    [Tags]  Known Issue Chrome    Known Issue Safari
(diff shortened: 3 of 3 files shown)


atest/run.py
```diff
@@ -208,6 +206,7 @@ def execute_tests(interpreter, browser, rf_options, grid, event_firing):
     options.extend([opt.format(browser=browser) for opt in ROBOT_OPTIONS])
     if rf_options:
         options += rf_options
+    options += ["--exclude", f"known issue {browser.replace('headless', '')}"]
```
Brownies, line 209: Do you think this is fine or should this be done the same way as  `browser` is inserted into `ROBOT_OPTIONS`?

rasjani: looks good imho

atest/run.py
```diff
@@ -97,8 +97,6 @@
 REBOT_OPTIONS = [
     "--outputdir",
     RESULTS_DIR,
-    "--noncritical",
-    "known issue {browser}",
```
emanlove, line 101: There are more "Known Issue" tags beyond just chrome in these tests so I don't think `known issue {browser}` should be removed.

In addition the question of headless versus headed is a good one I'll address in the other change.

atest/run.py
emanlove: I don't think we can assume it would be headless. Instead I would treat headless as separate. As I write this I was hoping that I could get some ease to having to write both Known Issue Chrome and Known Issue Headless Chrome each and every time. But realized this is the same assuming that a chrome issue affect both headed and headless. 

Would a method be that if it affect only a headed or headless operation of the browser to say `Known Issue Chrome headless only`? Can we post process this with the current toolset or is this a bit more work?

emanlove: @Brownies Would you want to separate this question of known issues from the critical fix? Not saying you have to but I was assuming that the non critical is what is failing the test against Robot Framework version > 5.0.  If it is then separating this might allow us to resolve that and check newer RF versions while working through this question of known issues. Just a thought and open to differing opinions on this.

atest/run.py
Brownies: I don't quite understand why you would not want to remove this bit.

`--noncritical` is an option that takes a tag argument and `known issue {browser}` is the argument. If it is left there rebot will interpret it as the input file.

atest/run.py
Brownies: I'm quite sure handling headed and headless issues separately with a single tag will require some more effort.

I don't really want to separate this from the `--noncritical` fix because this is a replacement for that functionality. I can remove it if you really want to but it will cause tests with known issues to be executed unless users manually `--exclude` them. This wouldn't break Github actions though because all tests should pass with Chrome.

emanlove: I may have understood this wrong. I will relook at it sometime over the next few days.

PR #1798, opened 2022-10-10
Use python interpreter that executed atest/run.py instead of python

### LINKED ISSUE #1796: atest task loses python interpreter when running with virtualenv under Windows
# For issues
## Steps to reproduce the issue

Under Windows

```msbatch
mkdir atest-python-interpreter
cd atest-python-interpreter
virtualenv -p C:\Python39\python.exe venv-py39
venv-py39\Scripts\activate
git clone git@github.com:robotframework/SeleniumLibrary.git rf-sl
cd rf-sl
pip install -r requirements-dev.txt
inv atest
```

The last command results in an error
```
(venv-py39) D:\devsandbox\atest-python-interpreter\rf-sl>inv atest
  File "D:\devsandbox\atest-python-interpreter\rf-sl\utest\run.py", line 9, in <module>
    from pytest import main as py_main
ModuleNotFoundError: No module named 'pytest'
Not running acceptance test, because unit tests failed.
```
## Error messages and additional information

The last command results in an error
```
(venv-py39) D:\devsandbox\atest-python-interpreter\rf-sl>inv atest
  File "D:\devsandbox\atest-python-interpreter\rf-sl\utest\run.py", line 9, in <module>
    from pytest import main as py_main
ModuleNotFoundError: No module named 'pytest'
Not running acceptance test, because unit tests failed.
```

I am running under Windows 10 using Python 3.9.9 within a virtual env

## Expected behavior and actual behavior

I was expecting the runner to use the virtual env I have executed with. But instead it appeas to run `python` from a process not realizing it is a virtualenv that it was started under.

## Environment

Browser: Name and version (Usually available from the about dialogue.)
Browser driver: Name and version
Operating System: Windows 10
Libraries
- Robot Framework: 5.0.1
- Selenium: 4.5.0
- SeleniumLibrary: 6.1.0.dev1
- Interpreter: Python 3.9.9 under a virtualenv

```
(venv-py39) C:\projects\SeleniumLibrary\atest-python-interpreter\rf-sl>pip list
Package                      Version
---------------------------- -----------
[ ... snip ...]
pytest                       7.1.3
pytest-approvaltests         0.2.4
pytest-mockito               0.0.4
[ ... snip ...]
```

emanlove:
When running under a virtualenv, using `python` as the interpreter forgets the virtualenv python which executed atest.run.py. Instead using sys.executable which will execute with same interpreter used to execute atest/run.py. I also later check to make sure the interpreter is not None nor an empty string as doc states [1] it may be if "Python is unable to retrieve the interpreter".

Fixes #1796

[1] https://docs.python.org/3.10/library/sys.html#sys.executable

### DIFF
--- a/atest/run.py
+++ b/atest/run.py
@@ -110,7 +110,7 @@ def acceptance_tests(
     os.mkdir(RESULTS_DIR)
     if grid:
         hub, node = start_grid()
-    with http_server():
+    with http_server(interpreter):
         execute_tests(interpreter, browser, rf_options, grid, event_firing)
     failures = process_output(browser)
     if failures:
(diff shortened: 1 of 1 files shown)


Brownies (approved):

emanlove:
As this stands it resolves the issue of the executor but does not emulate the same as a virtualenv does. That adds some additional env variables as well.

PR #1799, opened 2022-10-12
Unify "index" type of keyword arguments.

### LINKED ISSUE #1791: Typing issues when using SeleniumLibrary directly from python.
select_from_list_by_index  expects the index to be \<str\> .. calling from RF side, this is not an issue but doesnt make sense when called from python .. 

create_webdriver also returns \<str\> but register_webdriver then returns \<int\> 

Action: 

Go thru the type annotations that refer to index and allow passing of \<int\>  instead of only  \<str\> 

PS. Typically RF itself doesnt get issues with this sort of type annotations - issue is mainly now apparent because plenty of Robocorp customers are making "pure python robots".

rasjani:
create_webdriver was marked as returning a str, however, its return value is coming from register_driver which is documented to return int. These where changed to return actual int.

switch_driver was only accepting str as "index_or_alias". While aliases are str's since create_driver now returns the actual index as int, accepting both types makes more sense.

select and unselect list keywords where typehinted to receive only str as index type - when called from python side, this does not make much sense so changed those to int as RF should do type conversion automatically.

Fixes #1791


rasjani:
no tests have been written nor executed ..

emanlove:
One overarching thought is I would want this to be backwards compatible. Such that this change does not force people to change all their select from list by index calls. So I would want the Robot Type conversion from the robot side to take those strings and automatically convert those to ints in all cases that switch from incoming str to incoming int.

Adding some inline comments as well ..

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -344,7 +344,7 @@ def _make_new_browser(
     @keyword
     def create_webdriver(
         self, driver_name: str, alias: Optional[str] = None, kwargs={}, **init_kwargs
-    ) -> str:
+    ) -> int:
```
emanlove, line 347: This one makes sense to me although I am needing to relook at the doc to see what one does with this return value. I read 

> Returns the index of this browser instance which can be used later to switch back to it. Index starts from 1 and is reset back to it when `Close All Browsers` keyword is used. See `Switch Browser` for an example.

What troubles me most here is that index of one. Just considering all the math one might need to have an one based index instead of a zero based index. But in the end changing that to zero based I do think would be a bad backwards compatibility change.

I will also add this one I am not too worried about change and being backwards incompatible. Not sure why I am not as concerned about this one in particular. But it seems like one I would be willing to break. (Now not sure if that would force this one into a major release with some fore warning .. which would be best if it was backwards incompatible)

src/SeleniumLibrary/__init__.pyi
```diff
@@ -133,7 +133,7 @@ class SeleniumLibrary:
     def select_all_from_list(self, locator: Union[selenium.webdriver.remote.webelement.WebElement, str]): ...
     def select_checkbox(self, locator: Union[selenium.webdriver.remote.webelement.WebElement, str]): ...
     def select_frame(self, locator: Union[selenium.webdriver.remote.webelement.WebElement, str]): ...
-    def select_from_list_by_index(self, locator: Union[selenium.webdriver.remote.webelement.WebElement, str], *indexes: str): ...
+    def select_from_list_by_index(self, locator: Union[selenium.webdriver.remote.webelement.WebElement, str], *indexes: int): ...
```
emanlove, line 136: Does this need to be `Union[str, int]` for Robot to do the type conversion?

emanlove: And if robot core does not do the type conversion I would be fine with both 

Union[str, int] as the argument type and then inrternally in all spots needed, something like

```
if isinstance(indexes, str):
    int(indexes)
```

.. although now I see these can be and are *indexes and not index, meaning it could be multiple indexes .. thus complicating my conversion strategy a bit ..

emanlove: [Noting these comments apply to all instances of these str to ints]

rasjani: Corner cases are not know but rf will do type conversion to int

Union is used only for with alias_or_index because alias is user defined string  bu index is int returned by create_driver and register_driver

src/SeleniumLibrary/keywords/browsermanagement.py
rasjani: the last line in this file is:         return self.ctx.register_driver(driver, alias)

This method his here: https://github.com/robotframework/SeleniumLibrary/blob/master/src/SeleniumLibrary/__init__.py#L530

and it is documented to return *int* @ https://github.com/robotframework/SeleniumLibrary/blob/master/src/SeleniumLibrary/__init__.py#L538

and the actual code that "generates" the index is result of len() call here: https://github.com/robotframework/robotframework/blob/master/src/robot/utils/connectioncache.py#L64

So, its pretty safe to say that regarding the driver, plain old index is always *int*

pekkaklarck:
Don't have time to review this right now, but I have one possible idea related to this. We recently changed Robot's Process library to [return actual process objects](https://github.com/robotframework/robotframework/issues/4104), not handles (e.g. ints), and it has worked great. Could `Create Webdriver` return the created WebDriver instance as well?

src/SeleniumLibrary/keywords/browsermanagement.py
emanlove: Ok. This looks good then.

Issue #1767, opened 2022-01-13
Add keywords for dealing with stale elements


Brownies:
For my use case it would be nice to have some keywords to be able to deal with elements that get removed from the DOM. Something along the lines of
- `Wait Until Element Is Stale`
- `Get Element Staleness`
- `Element Should (Not) Be Stale`

I wrote a rudimentary Python plugin at work using Selenium's [`staleness_of`](https://www.selenium.dev/selenium/docs/api/py/webdriver_support/selenium.webdriver.support.expected_conditions.html?highlight=staleness_of#selenium.webdriver.support.expected_conditions.staleness_of) condition. I got the OK to use the plugin as a base and spend some time so I'd be willing to try and provide a PR as well if this feature makes sense and would be a good fit for the library.

emanlove:
I know personally I have been hesitant to add an exhaustive set of keywords which specifically wait until some state. I was just now talking with Tatu about a question someone raised about waiting. He showed me this solution

```python
element = WebDriverWait(self.driver, 30, ignored_exceptions=(StaleElementReferenceException,)).until(
                EC.element_to_be_clickable((By.ID, "description"))
            )
```

I was thinking a generic Wait until Element State Is/Is Not might be the solution. There is an [full list of expected conditions](https://www.selenium.dev/selenium/docs/api/py/webdriver_support/selenium.webdriver.support.expected_conditions.html) that we could take advantage of .. I was just going to put in a reminder ticket for myself and saw your request. Do you think this (with a `Get Element State` and `Element State Should (Not) Be` keywords) fulfill your request?

Brownies:
Yeah that would be great!

PR #1808, opened 2022-12-22
Fix tests on firefox

### LINKED ISSUE #1804: RemoteDriverServerException was removed from Selenium
https://github.com/SeleniumHQ/selenium/commit/5fba3a2b2d24a02356def84ffc5618bda73cc09f breaks `utest/test/keywords/test_webdrivercache.py`

Brownies:
* Drop Robot Framework 3 and PythonLibCore 2 support, new version of one combined with older version of the other causes type conversion problems.
* Add Firefox to CI runs
* Adjust tests and tags so that they run on Chrome and Firefox


.github/workflows/CI.yml
```diff
@@ -44,24 +44,34 @@ jobs:
...
+
     - name: Run tests with headless Chrome and with PyPy
       if: matrix.python-version == 'pypy-3.7'
       run: |
         xvfb-run --auto-servernum python atest/run.py --nounit --zip headlesschrome
-    - name: Run tests with headless Chrome and with Python != 3.8
-      if: matrix.python-version != '3.8' && matrix.python-version != 'pypy-3.7'
+
+    - name: Run tests with normal Chrome and with Python 3.7
+      if: matrix.python-version == '3.7'
```
emanlove, line 54: This appears to be running headless Chrome and not "normal" Chrome.

emanlove: [Thinking out loud ..] Wondering about the matrix of python versions and browsers versions. Not sure the origin of these specific combinations but it may be worth checking overall what we are testing to see if there are gaps. For example here it look like we have switched from testing normal Chrome with Python 3.8 to Python 3.7.

And with the expanding versions of Python to 3.9, 3.10 how does these fit into the mix. @aaltat always had the strong opinion of limiting the overall testing to just the latest versions. He has very good reason for this - essentially limited resources and time- although I keep wanting expand the testing matrix. I'll say I have no strong reasons for this nor have I overcome the real issue of limited resources. One idea I had was to make this a separate GitHub workflow file which would not "block" - this stil doesn't resolve the real issue of support. We both agree that being able to publicly state what is supported - i.e. what we have tested against - is important and should be done.
This background info I think is important just so all understand all the previous thinking and work that has gone into the current test setup. This way we can build upon that experience and not struggle through the same without changes or improving upon that experience. I'm not opposed to change but am wanting to learn from others experiences in order to make better.

atest/acceptance/2-event_firing_webdriver/selenium_move_to_workaround.robot
```diff
@@ -38,7 +38,7 @@ Click Element
 Double Click Element
     [Documentation]    LOG 1 Double clicking element 'doubleClickButton'.
     [Setup]    Initialize Page For Click Element
-    [Tags]    Known Issue Safari    Known Issue Firefox
+    [Tags]    Known Issue Safari
```
emanlove, line 41: [More thinking out loud ..] Wondering if it is both helpful and practical to be able to, if available, attach an external bug issue somewhere, somehow to some of these so we can track them. For example I have no idea about these issues, when they were reported, really nothing. Nor do I know how to look nor see if they have been repaired. Of course I could run these tests ignore the known issue tag but just because the pass doesn't give me a deep confidence around the issue resolution.

atest/acceptance/keywords/location.robot
```diff
@@ -106,8 +106,8 @@ Wait Until Location Is Not Fails With Timeout
     ${orig_timeout}=    Set Selenium Timeout    2 s
     Click Element   button
     Run Keyword And Expect Error
-    ...     Location is 'http://localhost:7000/html/javascript/wait_location.html' in 1 second.
-    ...     Wait Until Location Is Not     http://localhost:7000/html/javascript/wait_location.html   timeout=1 s
+    ...     Location is 'http://localhost:7000/html/javascript/wait_location.html' in 750 milliseconds.
+    ...     Wait Until Location Is Not     http://localhost:7000/html/javascript/wait_location.html   timeout=750ms
```
emanlove, line 110: Just curious .. why the change here?

atest/acceptance/windows.robot
```diff
@@ -169,7 +175,7 @@ Select Popup Window With Delay By Excluded List
     @{excluded_handle_list}=    Get Window Handles
     Click Button     id:MyButton
     Switch Window    ${excluded_handle_list}    timeout=5
-    Title Should Be    Original
+    Wait Until Keyword Succeeds    5s    200ms    Title Should Be    Original
```
emanlove, line 178: I'm a bit cautious with this approach and change. It is not surprising to me that over time, for tests that were once running fine timing issues appear. I guess what I want to ask is that is a timing issue does appear is there any deeper understanding as to why. I suspect in most cases there is not. I just don't want to be masking issues in the tests and timing issues, especially when the appear where they were once not there, is something that throws alarm bells in my head.

I would also like our tests to be a model for how others should right tests, as best as we can. I know this is an ever difficult battle to fight or steep hill to climb. I'm not a fan of Wait Until Keyword Succeeds (WUKS) but as we don't have yet implemented the [generic wait for element state](https://github.com/robotframework/SeleniumLibrary/issues/1767#issuecomment-1016653989) then this is fine.

requirements.txt
```diff
@@ -1,3 +1,3 @@
 selenium >= 4.0.0
-robotframework >= 3.2.2
-robotframework-pythonlibcore >= 2.2.1
+robotframework >= 4.1.3
```
emanlove, line 2: Is there are hard requirement for version increase? Or is this a preferred/suggested one?

src/SeleniumLibrary/keywords/webdrivertools/webdrivertools.py
```diff
@@ -228,7 +228,7 @@ def _geckodriver_log(self):
         log_file = self._get_log_path(
             os.path.join(self.log_dir, "geckodriver-{index}.log")
         )
-        logger.info(f"Firefox driver log is always forced to to: {log_file}")
+        logger.trace(f"Firefox driver log is always forced to to: {log_file}")
```
emanlove, line 231: Should this be debug as compared to trace level? I'm pretty sure we don't have published guidelines so almost asking myself here why one over the other ..

emanlove (changes requested):
Added some inline comments ...

requirements.txt
emanlove: Sorry, sometimes I am a bit forgetful even of things/changes I have made myself. Yes, we had bumped up the tested version in the matrix dropping out version 3.2.2 due to the many issues that were occurring and due to the fact we are currently on rf 6.x.

One question I have which I want to ask, if not to myself, is why raise the minimum to 4.1.3 and not 4.1.0 or more simply >4.x? Really I am asking what are difference amongst the point releases of the core? Are they security ones such that it makes sense to jump to the latest? or are they a feature releases that giving a lower version number support a widest audience - which as I type this I hear Tatu telling just the latest.

src/SeleniumLibrary/keywords/webdrivertools/webdrivertools.py
Brownies: Having this at info or debug breaks a whole bunch of acceptance tests for Firefox, I didn't see an obvious way to make robotstatuschecker ignore these.

requirements.txt
Brownies: 3.2.2 was dropped because it has problems with pythonlibcore 4, I didn't test plc 3 though.


There's no specific reason for 4.1.3, I just picked the latest 4.x and that's also what we test.
I checked just now and there are some [small backwards incompatibilities between 4.0.x and 4.1.x](https://github.com/robotframework/robotframework/blob/master/doc/releasenotes/rf-4.1.rst#id38). However, I think upgrading shouldn't be too difficult if some people have stuff that depends on the earlier behaviour. Also, there are some nice bug fixes and features in 4.1.x series. 4.1.1 adds official support for Python 3.10 and 4.1.2 has some improvements to memory use. 4.1.3 just fixes a regression introduced in 4.1.2. Overall I think users would benefit from running 4.1.3 compared to other 4.x versions.

atest/acceptance/keywords/location.robot
Brownies: This is another timing issue / reliability improvement.
[The page](https://github.com/Brownies/SeleniumLibrary/blob/fix-tests-on-firefox/atest/resources/html/javascript/wait_location.html#L5) has a 1000 millisecond timer that changes the location. This was mostly fine on CI but sometimes had problems locally.

atest/acceptance/windows.robot
Brownies: I think this was fine on Chrome but Firefox had timing problems, I will try to confirm this later. The window is selected correctly, just that sometimes it would read the title of the old window. My guess is that switching windows is a blocking action with chromedriver but not blocking with geckodriver.

atest/acceptance/2-event_firing_webdriver/selenium_move_to_workaround.robot
Brownies: It sure would be helpful to attach issue references or bugtracker links when available, either as `[Documentation]` or inline comments. I don't know anything about this problem either, just that it works now.

Github has a "View git blame" feature that lets you see changes to a particular line but that doesn't really give any info other than who made the change and whatever is in the commit message.

src/SeleniumLibrary/keywords/webdrivertools/webdrivertools.py
emanlove: Makes sense and good enough reason (for now) to drop these down into trace level

atest/acceptance/2-event_firing_webdriver/selenium_move_to_workaround.robot
emanlove: I'm going to open a new issue, mostly as a reminder, to look at the project's process  to see if we can improve upon this and provide a way to track these if possible.

atest/acceptance/keywords/location.robot
emanlove: Make sense. Here we want a failure due to the timeout. But the given timeout is the same as the time it should take. So an very small fluctuation could cause a false error.

atest/acceptance/windows.robot
emanlove: Just thinking this through

- This test case purposes states via its test case name "With Delay". And there is another similar test case; one without delay "Select Popup Window By Excluded List". What then is the intended difference?
- We want/expect it to be successful. This can be said because of any lack of expect error or other validation code to the opposite. But is this test testing that with the delay it should be "old title" that we are expecting. Or  are we expecting a blocking action in all cases and thus the title should be the "new title".
- With the timeout argument on the Switch Window call being 5[secs] should the timeout on WUKS be around the same? or more? or less? Basically how do these two interact if at all?

Brownies: The difference between `Select Popup Window By Excluded List` and `Select Popup Window With Delay By Excluded List` is that "With Delay" clicks a button that sets a timer for 1000 milliseconds that then opens the popup, while the first test clicks a link that opens it immediately.

Both tests first navigate to [popupwindow.html](https://github.com/Brownies/SeleniumLibrary/blob/fix-tests-on-firefox/atest/resources/html/javascript/popupwindow.html) and the title of that page is `Click link to show a popup window`. The popup that is opened is [dynamic_content.html](https://github.com/Brownies/SeleniumLibrary/blob/fix-tests-on-firefox/atest/resources/html/javascript/dynamic_content.html) that, somewhat confusingly for these tests, has title `Original`.

The WUKS timeout probably doesn't matter too much as long as it retries at least a couple of times. I don't think it has any interaction directly with `Switch Window`.

.github/workflows/CI.yml
Brownies: Good catch, I will fix this to normal chrome. 3.8 was not tested previously, despite what some of the task names state. This is because 3.8 was not(and is not) included in the matrix.

I would prefer to get rid of this `if python-version == xx && yy=zz` and just use the matrix to define browsers and python+rf versions. Maybe even Selenium versions in the future, though this is not realistic at the moment. I realize this results in a lot of jobs very quickly but as far as I can tell Github Actions are free for public repositories and the only limit is 20 concurrent jobs. I think this would further encourage making current tests more robust or not-flaky by fixing timing issues and such, and also to help create robust tests in the future.

emanlove: The conditions on python version I believe were to limit the overall matrix. I'm good with expanding it. I would say possibly in another pull request.

atest/acceptance/windows.robot
emanlove: Thus the idea that Switch Window wouldn't return to the title had adjusted. Ok, let's leave the WUKS in there.

requirements.txt
emanlove: I'm good with 4.1.3. For anyone interested [here was a discussion](https://robotframework.slack.com/archives/C0K0240NL/p1672429142861239) we had about versions within the Robot Framework Slack community.

.github/workflows/CI.yml
Brownies: Yeah definitely not doing that in this pull request.

PR #1810, opened 2023-01-01
Remove se v3 bug workaround

emanlove:
This pulls out the code for working around a known issue within Selenium version 3.x It has been resolved under Selenium 4.x and as we have a minimum requirement of v4.0.0 we can now remove this code.


yuriverweij (commented):
Looks good to me

PR #1814, opened 2023-02-19
Update Page should contain doc

### LINKED ISSUE #1785: Review Page Should Contain documentation
With the change under Selenium Version 4.1.4, the logging was changed. This keyword doc talks about the default level the error message is run under. With the change I am not sure this is still the same. Need to review keyword documentation.

emanlove:
Corrects the default custom log level to `TRACE` and adds some additional tests proving out this change and some other statements made within the keyword documentation.

Resolves #1785

### DIFF
--- a/atest/acceptance/keywords/content_assertions.robot
+++ b/atest/acceptance/keywords/content_assertions.robot
@@ -37,6 +37,22 @@ Page Should Contain With Text Having Internal Elements
     Go to page "links.html"
     Page Should Contain    Relative with text after
 
+Page Should Contain With Custom Log Level INFO
+    [Tags]    NoGrid
+    [Documentation]    Html content is shown at INFO level.
+    ...    FAIL Page should have contained text 'non existing text' but did not.
+    ...    LOG 1:18 INFO REGEXP: (?i)<html.*</html>
+    ...    LOG 1:19 FAIL Page should have contained text 'non existing text' but did not.
+    Page Should Contain    non existing text    INFO
+
+Page Should Contain With Custom Log Level WARN
+    [Tags]    NoGrid
+    [Documentation]    Html content is shown at WARN level.
+    ...    FAIL Page should have contained text 'non existing text' but did not.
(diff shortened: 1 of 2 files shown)


atest/acceptance/keywords/content_assertions.robot
```diff
@@ -27,6 +27,16 @@ Page Should Contain
     Page Should Contain    This is the haystack
     Page Should Contain    non existing text
 
+Page Should Contain Using Default Custom Log Level
```
yuriverweij, line 30: See comment for line 74, this test is basically a duplicate.

atest/acceptance/keywords/content_assertions.robot
```diff
@@ -47,14 +73,30 @@ Page Should Contain With Custom Log Level DEBUG
 
 Page Should Contain With Custom Log Level TRACE
```
yuriverweij, line 76: This test is now basically duplicate for "Page Should Contain Using Default Custom Log Level' on line 30.
For consistency would it not make more sense to use the title from this test("Page Should Contain With Custom Log Level Trace")? but use the re-setting of original Log Level like in the other tests? `${old_level}= Set Log Level TRACE`

emanlove: This is a very good catch and great feedback. There is a subtle difference in the tests. The first doesn't set the loglevel (Line 39*: `Page Should Contain non existing text`). Whereas the second explicitly specifies the loglevel to be TRACE (Line 83*: `    Page Should Contain    non existing text    TRACE`). It took me a while to go back through this code and understand the changes I had initially made here. I recalled something about the default level and how the doc specified one level but execution said another. Taking your feedback I have updated the documentation on a few of these tests. Let me know if this is any clearer.

* lines numbers are reference updated code [27e0602](https://github.com/robotframework/SeleniumLibrary/pull/1814/commits/27e06024cd75c18f2a93bfe31bcc4f29da761c26)

yuriverweij (approved):

Issue #1784, opened 2022-09-29
Scroll Element in to view not working in seleniumlibrary


everlighting:
Scroll Element in to view is not working .I don't see any new updates in seleniumlibrary after january,on the other hand browser library is releasing in every weeks !!!

emanlove:
If you could give a bit more information about what error message you may be getting, what system information you are running with and an example which reproduces the problem that would be helpful.

Brownies:
Probably a duplicate of #1780

everlighting:
@Brownies anyone looking in to this issue?

Brownies:
Not that I know of.

Issue #1780, opened 2022-08-22
'Scroll Element Into View' is broken on Firefox


Brownies:
## Steps to reproduce the issue
* Open a page in Firefox such that some element is not inside the viewport
* Try to scroll to the element with 'Scroll Element Into View'

## Expected behavior and actual behavior

When I follow those steps, I see...
[Gist with trace level log output, stack trace and sample .robot and .html files](https://gist.github.com/Brownies/72f9d5b42a14174bcfd25bb9080ec6ef)

I was expecting the element to be scrolled into view. This does work on Chromium, both headless and normal.

## Environment

Browser: Firefox 103.0
Browser driver: geckodriver 0.31.0
Operating System: Linux, Ubuntu 20.04
Libraries
- Robot Framework: 5.0
- Selenium: 4.1.3 / 4.4.3
- SeleniumLibrary: 6.0.0
- Interpreter: CPython 3.8.10

Brownies:
A likely culprit seems to be the [keyword itself](https://github.com/robotframework/SeleniumLibrary/blob/c6e28bdda61abefefa875f4ef8661d7e2a6f81ca/src/SeleniumLibrary/keywords/element.py#L739) which uses `ActionChains.move_to_element` instead of `ActionChains.scroll_to_element`.

everlighting:
Is some one following up? When this issue will be released?  Once this issue is fixed we can remove the JS in our script

Brownies:
I'll see if I can fix this and get the acceptance tests running on Firefox too.

Brownies:
Bad news, switching to `ActionChains.scroll_to_element` does not fix the problem. It's also present with RF 6 and selenium 4.7.2 with Firefox 108.
Maybe this is a Selenium or Geckodriver problem?

Brownies:
Yeah, this is an ancient Geckodriver issue and they will not fix it until W3C updates the WebDriver spec: https://github.com/mozilla/geckodriver/issues/776#issuecomment-355086173
Might as well close this issue because the problem is not in SeleniumLibrary and judging by the rate of recent commits the spec update is going to take an eternity.

PR #1816, opened 2023-02-20
Add `scroll_element_into_view` workaround for Firefox "MoveTargetOutOfBoundsException" error

jnhyperion:
Running `scroll_element_into_view` on a dropdown list may have different results on `Firefox` and `Chrome`, on `Firefox` it will raise `MoveTargetOutOfBoundsException` in certain cases while on `Chrome` it works well.

probably fix:

https://github.com/robotframework/SeleniumLibrary/issues/1784
https://github.com/robotframework/SeleniumLibrary/issues/1780


emanlove:
Scrolling in Firefox is broken due to a long standing issues with Firefox. Going to move this into #1780.

PR #1840, opened 2023-06-15
Imporve the indexing function for the pages screenshot.

AmirHdm:
Add indexing functionality based on datetime for capture screen page method inside ScreenshotKeywords class.


emanlove:
@AmirHdm Just wanted to let you know that your pull request has been noticed and within the queue to be reviewed. My plan was to bring this to the team to review and discuss. I will update you sometime within the next couple weeks.

AmirHdm:
> @AmirHdm Just wanted to let you know that your pull request has been noticed and within the queue to be reviewed. My plan was to bring this to the team to review and discuss. I will update you sometime within the next couple weeks.

@emanlove great ! And thanks for the feedback

lisacrispin:
@AmirHdm Is this changing the naming of screenshot files from simply numbering them sequentially based on the index, to using the timestamp in the name? That does seem like a good idea, if so.

AmirHdm:
> 

@lisacrispin yeath that's right because using timestamps in naming is more helpful and significantly

emanlove:
@AmirHdm  I like the idea of providing different type of "indexing" or filenames. One concern I do have though is breaking any existing systems that rely on numerical indexing. Relooking at [the code](https://github.com/robotframework/SeleniumLibrary/blob/master/src/SeleniumLibrary/keywords/screenshot.py#L129) and [keyword documentation](https://robotframework.org/SeleniumLibrary/SeleniumLibrary.html#Capture%20Page%20Screenshot) I see for singular Capture Page Screenshot there is a filename argument which allows one to format the filename. I'm curious as to whether we could use this "syntactical idea" to allow for something like `filename = selenium-screenshot-{timestamp}.png` or `filename = selenium-screenshot-{timestamp:%Y-%m-%d-%H-%M-%S}.png` or maybe even `filename = selenium-screenshot-{random}.png` and apply it to the default capture screenshot on error functionality?

AmirHdm:
> @AmirHdm I like the idea of providing different type of "indexing" or filenames. One concern I do have though is breaking any existing systems that rely on numerical indexing. Relooking at [the code](https://github.com/robotframework/SeleniumLibrary/blob/master/src/SeleniumLibrary/keywords/screenshot.py#L129) and [keyword documentation](https://robotframework.org/SeleniumLibrary/SeleniumLibrary.html#Capture%20Page%20Screenshot) I see for singular Capture Page Screenshot there is a filename argument which allows one to format the filename. I'm curious as to whether we could use this "syntactical idea" to allow for something like `filename = selenium-screenshot-{timestamp}.png` or `filename = selenium-screenshot-{timestamp:%Y-%m-%d-%H-%M-%S}.png` or maybe even `filename = selenium-screenshot-{random}.png` and apply it to the default capture screenshot on error functionality?

@emanlove sure i guess we could apply your idea and actually it's a great feature then let me see what i could do and will give a quick feedback

AmirHdm:
The user now is able to set 3 indexing options on the keyword capture page screenshot :
* filename: str = 'selenium-screenshot-{index}.png'
* filename: str = 'selenium-screenshot-{random}.png'
* filename: str = 'selenium-screenshot-{timestamp}.png'

src/SeleniumLibrary/keywords/screenshot.py
```diff
@@ -201,16 +215,31 @@ def _get_screenshot_path(self, filename):
...
+                if formatted == filename or not os.path.exists(path):
+                    return path       
+            elif 'index' in filename:
+                index+=1
+                formatted = _format_path(filename, index)
+                path = os.path.join(directory, formatted)
+                # filename didn't contain {index} or unique path was found
+                if formatted == filename or not os.path.exists(path):
+                    return path       
+
```
emanlove, line 242: Just a quick read of this I believe there is a logic error here where if none of the three are in the filename one will get themselves into an infinite loop.

PR #1850, opened 2023-08-12
New attribute property keywords 1822

emanlove:
Added `Get DOM Attribute` and `Get Property` keywords and several acceptance tests. I know these test can be cleaned up a bit. One bit of review I am looking for is whether all the tests are there? Or if there might be something missing? I also added some comments within the tests. One issue I have been coming across is that when I go into an internal test (either acceptance or unit) some of them I have a tough time trying to see exactly what is is testing, and what should I expect it to do and verify. So I started to play around a bit with comments. I don't think there that great and need improvement. But this starts to highlight one idea for explaining tests better.


atest/acceptance/keywords/elements.robot
```diff
@@ -60,6 +60,55 @@ Get Element Attribute
...
+
+Get non existing DOM Attribute
+    ${class}=    Get DOM Attribute    link:Link with id    class
+    Should Be Equal    ${class}    ${NONE}
+
+More DOM Attributes
+    [Setup]    Go To Page "forms/enabled_disabled_fields_form.html"
+    # Test get empty attribute
+    ${disabled}=    Get DOM Attribute    css:input[name="disabled_input"]    disabled
+    Should Be Equal    ${disabled}    true   # ${True}
```
emanlove: The comment here at the end was what I was expecting the result to be; a boolean True. I left here for now to remind me that when I do the write up on how to use these new keywords to talk about what one should expect and to expect maybe the unexpected.

atest/acceptance/keywords/elements.robot
```diff
@@ -60,6 +60,55 @@ Get Element Attribute
...
+    ${disabled}=    Get DOM Attribute    css:input[name="disabled_password"]    disabled
+    Should Be Equal    ${disabled}    true   # disabled
```
emanlove: Same here; the comment here at the end was what I was expecting the result to be.

atest/acceptance/keywords/elements.robot
```diff
@@ -60,6 +60,55 @@ Get Element Attribute
...
+    # Test empty string as the value for the attribute
+    ${empty_value}=    Get DOM Attribute    css:input[name="disabled_password"]    value
+    Should Be Equal    ${empty_value}    ${EMPTY}
+    ${disabled}=    Get DOM Attribute    css:input[name="enabled_password"]    disabled
+    Should Be Equal    ${disabled}    ${NONE}   # false
```
emanlove: .. as with here.

src/SeleniumLibrary/keywords/element.py
```diff
@@ -409,6 +409,37 @@ def get_element_attribute(
         """
         return self.find_element(locator).get_attribute(attribute)
 
+    @keyword
+    def get_dom_attribute(
+        self, locator: Union[WebElement, str], attribute: str
+    ) -> str:
+        """Returns the value of ``attribute`` from the element ``locator``. `Get DOM Attribute` keyword
+        only returns attributes declared within the element's HTML markup.
```
emanlove: The last sentence here,

> `Get DOM Attribute` keyword only returns attributes declared within the element's HTML markup.

come from the Selenium Python code. Technically this is not actually correct. I found with the tests above, and it would be good to have someone else verify, that if the requested attribute is not there it returns `${None}`. In part that was what I was exploring with the tests. Would it throw an error or return a NaN type response (which Python generally doesn't distinguish from None)? So I suspect we should change this to something else; still explaining what dom attribute is but clarify expected results.

atest/acceptance/keywords/elements.robot
```diff
@@ -60,6 +60,55 @@ Get Element Attribute
     ${class}=    Get Element Attribute    ${second_div}    class
     Should Be Equal    ${class}    Second Class
 
+Get DOM Attribute
+    ${id}=    Get DOM Attribute    link:Link with id    id
+    Should Be Equal    ${id}    some_id
+    # Test custom attribute
+    ${existing_custom_attr}=   Get DOM Attribute    id:emptyDiv  data-id
+    Should Be Equal    ${existing_custom_attr}    my_id
+    ${doesnotexist_custom_attr}=   Get DOM Attribute    id:emptyDiv  data-doesnotexist
```
yuriverweij, line 95: I'd split these tests up (existing, existing custom attr, non existing custom attr) for the "normal" non-existing attr there is already a separate test below.

atest/acceptance/keywords/elements.robot
```diff
@@ -60,6 +60,55 @@ Get Element Attribute
...
+    ${isConnected}=   Get Property  css:input[name="readonly_empty"]     isConnected
+    Should Be Equal    ${isConnected}    ${True}
+
+    # ToDo: nned to test own versus inherited property
+    # ToDo: Test enumerated property
+    # Test proprty which returns webelement
+    ${children}=    Get Property    id:table1    children
+
+
+Get "Attribute" That Is Both An DOM Attribute and Property
```
yuriverweij, line 131: Is it possible to add a test where the property is different than the attribute?

src/SeleniumLibrary/keywords/element.py
yuriverweij: Might be good to mention this behavior in the comments.

yuriverweij (commented):
Looks good to me. I've added a few comments.

.github/workflows/CI.yml
```diff
@@ -27,7 +36,7 @@ jobs:
         python -m pip install --upgrade pip
         pip install -r requirements-dev.txt
     - name: Install dependencies for pypy
-      if: matrix.python-version == 'pypy-3.7'
+      if: matrix.python-version == 'pypy-3.9'
```
yuriverweij: Should something like the same comment as somewhere further on be added here? 
` # Recognize for the moment this will NOT run as we aren't using CPython 3.9`

.github/workflows/CI.yml
```diff
@@ -36,25 +45,28 @@ jobs:
...
         pip install -U --pre robotframework==${{ matrix.rf-version }}
+    - name: Install drivers via selenium-manager
+      run: |
+        SELENIUM_MANAGER_EXE=$(python -c 'from selenium.webdriver.common.selenium_manager import SeleniumManager; sm=SeleniumManager(); print(f"{str(sm.get_binary())}")')
+        echo "$SELENIUM_MANAGER_EXE"
+        echo "WEBDRIVERPATH=$($SELENIUM_MANAGER_EXE --browser chrome --debug | awk '/INFO[[:space:]]Driver path:/ {print $NF;exit}')" >> "$GITHUB_ENV"
+        echo "$WEBDRIVERPATH"
     - name: Generate stub file for ${{ matrix.python-version }}
-      if: matrix.python-version != 'pypy-3.7'
+      if: matrix.python-version != 'pypy-3.9'
```
yuriverweij: same here
` # Recognize for the moment this will NOT run as we aren't using CPython 3.9`

.github/workflows/CI.yml
```diff
@@ -36,25 +45,28 @@ jobs:
...
       run: |
         invoke gen-stub
-    - name: Debugging
-      if: matrix.python-version != 'pypy-3.7'
-      run: |
-        which python
 
     - name: Run tests with headless Chrome and with PyPy
-      if: matrix.python-version == 'pypy-3.7'
+      if: matrix.python-version == 'pypy-3.9'
```
yuriverweij: and here 
` # Recognize for the moment this will NOT run as we aren't using CPython 3.9`

.github/workflows/CI.yml
```diff
@@ -36,25 +45,28 @@ jobs:
...
       run: |
         xvfb-run --auto-servernum python atest/run.py --nounit --zip headlesschrome
 
-    - name: Run tests with normal Chrome and with Python 3.7
-      if: matrix.python-version == '3.7'
+    - name: Run tests with normal Chrome if CPython
```
yuriverweij: This does not seem to be correct?  name says 'if CPython', but comparison in next line says '!= pypy-3-9'

.github/workflows/CI.yml
```diff
@@ -65,12 +77,12 @@ jobs:
...
         xvfb-run --auto-servernum python atest/run.py --zip firefox
 
-    - name: Run tests with Selenium Grid
-      if: matrix.python-version == '3.8' && matrix.rf-version == '3.2.2' && matrix.python-version != 'pypy-3.7'
-      run: |
-        wget --no-verbose --output-document=./selenium-server-standalone.jar http://selenium-release.storage.googleapis.com/3.141/selenium-server-standalone-3.141.59.jar
-        sudo chmod u+x ./selenium-server-standalone.jar
-        xvfb-run --auto-servernum python atest/run.py --zip headlesschrome  --grid True
+    # - name: Run tests with Selenium Grid
+    #   if: matrix.python-version == '3.11' && matrix.rf-version == '3.2.2' && matrix.python-version != 'pypy-3.7'
```
yuriverweij: update to 'pypy-3-9' ?

.github/workflows/CI.yml
```diff
@@ -65,12 +77,12 @@ jobs:
       run: |
         xvfb-run --auto-servernum python atest/run.py --zip firefox
 
-    - name: Run tests with Selenium Grid
-      if: matrix.python-version == '3.8' && matrix.rf-version == '3.2.2' && matrix.python-version != 'pypy-3.7'
-      run: |
-        wget --no-verbose --output-document=./selenium-server-standalone.jar http://selenium-release.storage.googleapis.com/3.141/selenium-server-standalone-3.141.59.jar
-        sudo chmod u+x ./selenium-server-standalone.jar
-        xvfb-run --auto-servernum python atest/run.py --zip headlesschrome  --grid True
+    # - name: Run tests with Selenium Grid
```
yuriverweij: Should all this be commented out? These will not run anyway because 'matrix.rf-version 3.2.2' is not in the list. Maybe also add a comment like the others?

PR #1851, opened 2023-08-12
Added `Print Page As PDF` Keyword

emanlove:
Some initial ideas and sketches for the `Print Page As PDF` keyword. I wanted to seewhat it may look like and how it may work. So this is just some preliminary ideas. Seeing a need to handle Print Options, need to document options and what are valid values, how to deal with the return encoded base64 string, etc.


src/SeleniumLibrary/keywords/screenshot.py
```diff
@@ -235,3 +237,20 @@ def _embed_to_log_as_file(self, path, width):
...
+        """ Print the current page as a PDF
+
+        """
+        if not print_options:
+            print_options = PrintOptions()
+            print_options.page_ranges = ['-']
+
+        base64code  = self.driver.print_page(print_options)
+        pdfdata = b64decode(base64code)
+        with open('test.pdf', mode='wb') as pdf:
```
emanlove: This is very preliminary. For example, here I just use a test file. I expect a fair more code and allowing the user to specify the optional filename or filename syntax when printing page as pdf.

atest/acceptance/keywords/print_page.robot
```diff
@@ -0,0 +1,18 @@
+*** Settings ***
+Documentation    Suite description
+Suite Setup       Go To Page "non_ascii.html"
+Resource          ../resource.robot
+
+*** Test Cases ***
+Print Page As PDF Without Print Options
+    Print Page As PDF
+
+Provide Print Options From Module
```
emanlove: With this test I just wanted to do a preliminary exploration of how to set print options. A while back for when one opens a browser there were ways to customize how the browser was started. To do this one had to do something similar, which is go into the python module and make an instance of the python class of options. Then add to that robot variable as class instance with a method on that class to add attributes. It was fairly ugly and the syntax was all python. But doable. Tatu really cleaned that up with just having a string of options. I wanted to see if one could add print options that old school way. Here I left in my failed attempts as a reminder of what failed and what worked so if we do explain it with this then I can say what and what not to do.

src/SeleniumLibrary/keywords/screenshot.py
```diff
@@ -235,3 +237,20 @@ def _embed_to_log_as_file(self, path, width):
             f'<a href="{src}"><img src="{src}" width="{width}px"></a>',
             html=True,
         )
+    
+    @keyword
+    # def print_page_as_pdf(self, print_options: Optional[PrintOptions]=None,):
```
emanlove: Trying to figure out the proper type hints and type conversion signature for this method ..

src/SeleniumLibrary/keywords/screenshot.py
yuriverweij: That was the first thing I noticed, thanks for clarifying.

yuriverweij: Would it make sense to be able to set a default set of PrintOptions? Meaning the user can set the PrintOptions once, and those settings will be used every time the print_page_as_pdf will be called?
Something like with setting the timeouts  
`${old_value}  Set Print to PDF options  scale=0.5 password=${pdfpassword}`
or something like that

yuriverweij (commented):
Haven't been able to try this yet, but added some thoughts/comments

src/SeleniumLibrary/keywords/screenshot.py
```diff
@@ -241,16 +241,64 @@ def _embed_to_log_as_file(self, path, width):
     @keyword
     # def print_page_as_pdf(self, print_options: Optional[PrintOptions]=None,):
     # def print_page_as_pdf(self, print_options: Optional[Union[PrintOptions, dict]]=None):
-    def print_page_as_pdf(self, print_options: Union[PrintOptions, dict, None]=None):
+    # def print_page_as_pdf(self, print_options: Union[PrintOptions, dict, None]=None):
```
yuriverweij: comments can be removed

src/SeleniumLibrary/keywords/screenshot.py
emanlove: Initially accepting the all the print options within keyword arguments.

src/SeleniumLibrary/keywords/screenshot.py
emanlove: Removed. Added some more, and eventually removed those as well. [Laughing]

atest/acceptance/keywords/print_page.robot
```diff
@@ -0,0 +1,24 @@
+*** Settings ***
+Documentation    Suite description
+Suite Setup       Go To Page "non_ascii.html"
+Resource          ../resource.robot
+
+Test Setup    Remove Files    ${OUTPUTDIR}/selenium-page-*.pdf
+
+*** Test Cases ***
+Print Page As PDF Without Print Options
```
aaltat, line 9: Is this same as next test?

emanlove: I guess it is but not not intended to be. The first was intended to make sure that a page would print if no arguments were passed to it and iusing some of the default settings.

The second one is to check the automatic indexing of the saved filename. But you are right that a single test would check both of these scenarios.

emanlove: Updated the second test using various arguments to make it sloightly different than the first.

src/SeleniumLibrary/keywords/screenshot.py
emanlove: It does makes sense to have a context for print options. For the initial release I am going to leave that out though.

PR #1859, opened 2023-10-29
Mutable default value 1817

emanlove:
Fixes issues with mutable default values.


src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -340,7 +340,7 @@ def _make_new_browser(
 
     @keyword
     def create_webdriver(
-        self, driver_name: str, alias: Optional[str] = None, kwargs={}, **init_kwargs
+        self, driver_name: str, alias: Optional[str] = None, kwargs=None, **init_kwargs
```
Snooz82: You should add the type hin here:
`kwargs: Optional[Dict] = None`

src/SeleniumLibrary/keywords/browsermanagement.py
```diff
@@ -371,6 +371,8 @@ def create_webdriver(
         `Close All Browsers` keyword is used. See `Switch Browser` for an
         example.
         """
+        if kwargs is None:
+            kwargs = {}
         if not isinstance(kwargs, dict):
             raise RuntimeError("kwargs must be a dictionary.")
```
Snooz82, line 368: These two line you could removed. 
Robot Framework will fail, if you give anything that can not be converted to dict.
At least if we are not covering every bad things people could do from Python.

Snooz82 (changes requested):
Two comments. 
The type hint is important. Otherwise it will use the default as type hint.

yuriverweij (commented):
Agree with Snooz82, nothing else to add

src/SeleniumLibrary/keywords/browsermanagement.py
emanlove: For these I would prefer to keep them there. Relying on another piece code, even a single entry point type path, is something I prefer not to do.

Snooz82 (approved):
looks good to me

emanlove:
The failing rf v6.1.1 failing test is unrelated so proceeding with the merge of this fix ..

PR #1864, opened 2023-11-06
Refactor inspect-based log method finder, use for all Services

bollwyvl:
## References
- continue #1862

## Changes
- [x] hoist `log_method` finder to method
- [x] also apply `log_(path|output)` detection for Firefox, Edge, IE, and Safari
  - https://github.com/SeleniumHQ/seleniumhq.github.io/commit/c11605ec062e49b2cbf1a35ddca0bf78224cc75f

### DIFF
--- a/src/SeleniumLibrary/keywords/webdrivertools/webdrivertools.py
+++ b/src/SeleniumLibrary/keywords/webdrivertools/webdrivertools.py
@@ -302,7 +302,8 @@ class WebDriverCreator:
             return self._remote(remote_url, options=options)
         if not executable_path:
             executable_path = self._get_executable_path(webdriver.edge.service.Service)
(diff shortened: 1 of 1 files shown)


src/SeleniumLibrary/keywords/webdrivertools/webdrivertools.py
```diff
@@ -131,6 +131,16 @@ def _remote_capabilities_resolver(self, set_capabilities, default_capabilities):
             caps["browserName"] = default_capabilities["browserName"]
         return {"desired_capabilities": caps}
 
+    def _get_log_method(service_cls, service_log_path):
```
emanlove: method signature missing `self`, as in

```
def _get_log_method(self, service_cls, service_log_path):
```

emanlove:
Nick, thank you very much for this. It covers those service log paths which I missed.

PR #1865, opened 2023-11-11
Add expected conditions 1827

emanlove:
Some very early and very basic sketches of what this might become ..


src/SeleniumLibrary/keywords/expectedconditions.py
```diff
@@ -0,0 +1,25 @@
...
+from selenium.webdriver.support.wait import WebDriverWait
+from selenium.webdriver.support import expected_conditions as EC
+
+class ExpectedConditionKeywords(LibraryComponent):
+    @keyword
+    def wait_for_expected_condition(self, condition, *args):
+        wait = WebDriverWait(self.driver, 10, 0.1)
+        # import sys,pdb;pdb.Pdb(stdout=sys.__stdout__).set_trace()
+        c = getattr(EC, condition)
+        wait.until(c(*args))
```
yuriverweij: I've been looking through the expectedconditions documentation.
Depending on the expectedcondition it seems either a bool, webelement or string is returned.
Do we want to add these return values?  (probably yes)
This also adds verification possibilities.

atest/acceptance/keywords/expected_conditions.robot
```diff
@@ -0,0 +1,21 @@
+*** Settings ***
+Test Setup        Go To Page "javascript/expected_conditions.html"
+Resource          ../resource.robot
+
+*** Test Cases ***
+Wait For Expected Conditions One Argument
+    Title Should Be    Original
+    Click Element    link=delayed change title
+    Wait For Expected Condition    title_is    Delayed
```
yuriverweij, line 9: ~If we get a return value for this we can verify the keyword. Right now we're verifying the Title Change and the wait, not necessarily the expected condition itself.  (The wait will wait for max timeout, after which the change will be done regardless if the expected condition returns true or false)~

Edit:
After doing some testing I see now that the wait fails if expected condition is not met. Still, returning the result might be useful since when it succeeds a WebElement/List/ etc can be returned

src/SeleniumLibrary/keywords/expectedconditions.py
```diff
@@ -0,0 +1,25 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from SeleniumLibrary.base import LibraryComponent, keyword
+from selenium.webdriver.support.wait import WebDriverWait
+from selenium.webdriver.support import expected_conditions as EC
+
+class ExpectedConditionKeywords(LibraryComponent):
+    @keyword
+    def wait_for_expected_condition(self, condition, *args):
```
yuriverweij: some brainstorming:
wait_for_expected_condition should take the following arguments:
- condition (condition=)
- timeout  (timeout=)
- msg (msg=custom message)
- arguments  (arguments for the condition, depending on the condition this can be a String, WebElement, List)

One (to me) "strange" expectedcondition is "new_window_is_opened". Does this one implicitly add the current number of windows?

src/SeleniumLibrary/keywords/expectedconditions.py
emanlove: I lean towards Yes but I also see there are a lot of different return values and combinations. So I am not sure what that means for properly setting the return. I also wonder if it is possible to set the return values based upon which expected condition is called?

I want to recommend that we put this into the next release and get expected conditions out there. Thoughts?

atest/acceptance/keywords/expected_conditions.robot
emanlove: I see the advantage here. Not sure of a structure to do this is in. Again I would say let's do this in a follow up release if you ok with that.

src/SeleniumLibrary/keywords/expectedconditions.py
emanlove: Timeout has been added. Msg has not and that would be good to get in there. Let's make a separate ticket for that.

src/SeleniumLibrary/keywords/expectedconditions.py
yuriverweij: I think for now the "wait for" is basically a boolean already, and can be used as such. You also show this in the tests. So for now it is ok, and maybe at a later stage look at what is needed?

yuriverweij:
"Wait For Expected Condition Times out within set timeout" test fails on Firefox only, strange...

atest/acceptance/keywords/expected_conditions.robot
yuriverweij, line 1: I would like to add more tests to this (in the near future?)  Especially failing tests. Not sure if every expected condition should have a failing + passing test, since the handling of the expected conditions themselves is done (and tested I assume) by selenium itself. Maybe a failing+passing test for each type of expected condition (bool, webelement, list, etc.)?

src/SeleniumLibrary/keywords/expectedconditions.py
```diff
@@ -0,0 +1,63 @@
+# Copyright 2016-     Robot Framework Foundation
```
yuriverweij, line 1: This should be updated to 2024 I assume?

yuriverweij (commented):
Looks good to me. I added a few comments, but I don't think they should be blocking.

src/SeleniumLibrary/keywords/expectedconditions.py
emanlove: As I read that is starting in 2016 and going to the present the copyright belongs to the foundation.

atest/acceptance/keywords/expected_conditions.robot
emanlove: Agreed

src/SeleniumLibrary/keywords/expectedconditions.py
emanlove: Ok, so I think we could set the type for the wait for expected condition keyword as you are right. That is singular in focus and has a return type. I was concerned about setting the return type on the (underlying) expected condition which seems a bit problematic or complicated based upon which expected condition it was. This was something i didn't want to hold up a release on given the amount of time and my lack of seeing how this can be done with multiple different method search with their own return type.

Now noting that the wait is a type and can be added now, looking at the selenium code I am confused about the type [[L81](https://github.com/SeleniumHQ/selenium/blob/trunk/py/selenium/webdriver/support/wait.py#L81)]. It looks like a generic type "T" with no really specifics on it. I would have thought it would have been Boolean as well. I don't know  if this is generic return type because Selenium didn't have a need to specify itself. Or if i am just understand the  type hinting here. I will note there is the similar until_not method in the same file which has an addition type `Literal[True]` as well. Still this doesn't help me to understand this but at least it is different.

yuriverweij: I'm not familiar enough with typing to say this with full certainty, but to me it seems like:
"until" waits until the response is no longer False or Falsy, so for conditions where a boolean is returned this is obvious. For responses with a WebElement or anything else it's basically the same; no response is still Falsy/False, but when a response is received it becomes True/Truthy so the "until" succeeds. No specific type needed, and the return value could be anything.

emanlove: Going to leave this as is for now ..

utest/test/keywords/test_firefox_profile_parsing.py
```diff
@@ -56,8 +56,21 @@ def test_single_method(self):
...
+        if "key1" in result_attr:
+            to_str = f"{to_str} key1 {result_attr['key1']}"
+        if "key2" in result_attr:
+            to_str = f"{to_str} key2 {result_attr['key2']}"
         self.results.append(to_str)
+
+    def _get_preferences_attribute(self, result):
+        # -- temporary fix to transition selenium to v4.17.2 from v4.16.0 and prior
+        # from inspect import signature
+        # sig = signature(result)
```
yuriverweij, line 69: These comments can be removed I assume?

.github/workflows/CI.yml
```diff
@@ -11,7 +11,7 @@ jobs:
       matrix:
         python-version: [3.8, 3.11]  # 3.12, pypy-3.9
         rf-version: [5.0.1, 6.1.1, 7.0]
-        selenium-version: [4.14.0, 4.15.2, 4.16.0]  #4.17.0, 4.18.0
+        selenium-version: [4.14.0, 4.15.2, 4.16.0, 4.17.2, 4.18.1, 4.19.0]
```
yuriverweij, line 14: It's becoming quite a long list again.

Issue #1843, opened 2023-06-28
Execute JavaScript converts arguments to strings with robot==6.1


ikseek:
## Steps to reproduce the issue
```
*** Settings ***
Library    SeleniumLibrary

*** Test Cases ***
Test Print Dictionary Value
    Open Browser        browser=chrome
    &{ARGS}=            Create Dictionary  key=value
    Execute JavaScript  alert(JSON.stringify(arguments))    ARGUMENTS    ${ARGS}
    Sleep    5
```

## Expected behavior and actual behavior
When I run this robot script using robot==6.1 I see the browser window with alert message box saying:
{"0":**"{'key': 'value'}"**}

I was expecting to see a dictionary value that seleniumlibrary with robot==6.0.2 produced
{"0":**{"key":"value"}**}

Notice that currently I see a string with python representation of the dict and previously it was a dict itself.

Hint: this happens because the type hint in this line insists on WebElements and strings:
https://github.com/robotframework/SeleniumLibrary/blob/9823418feda556d48829a3137eefb962c44b14ff/src/SeleniumLibrary/keywords/javascript.py#L33
Robot 6.1 seems to use type hints for automatic type conversions.

## Environment

Browser: Chrome (latest)
Operating System: Mac OS 13.4.1
Libraries
- Robot Framework: 6.1
- Selenium: 4.9.1
- SeleniumLibrary: 6.1
- Interpreter: cpython 3.11

Snooz82:
Without testing it, but reading the type hint of that keyword and its code, i would say, that the type hint of `WebElement | str` is simply wrong. 

Python Selenium docs do say for `execute_async_script` for argument `*args`:
>  *args: Any applicable arguments for your JavaScript.

So there should not be any type hint it can be `Union[Any, str]` or so.

Ps: i am again really disappointed by the docs of Selenium itself. selenium.dev does not mention how to use args at all and even readthedocs is absolutely too lean here.
And by reading the Selenium code that is implemeting this, i must say, that it is weird too. They have a weird special handling if one of the args is a list or tuple and contains `EventFiringWebElement`.

### However: in the end they are just calling `json.dumps` on whatever you hand over to `*args`.
So if you hand over a string, it stays a string and if you hand over any json serializable object, like a dictionary in this example it will work.

Therefore the type hint must say, that args can be anything that is json-serializable. 
If SeleniumLibrary would require at least Robot Framework 5.0, you could introduce a new type hint with a custom converter.

[...]

PR #1866, opened 2023-11-17
fixed type hinting so that it is not converted to str

Snooz82:
The problem was that the type hint `str` did convert everything to string.
So a dictionary `{'key': 'value'}` does become `"{'key': 'value'}"` and this is then put to `json.dumps`
That obviously is then not a js object anymore.

Added tests as well.


Snooz82:
fixes #1843

yuriverweij (commented):
Looks good to me

PR #1885, opened 2024-02-13
updated *** Setting/Keyword/Variable *** to *** Settings/Keywords/Variables ***

yuriverweij:
Singular section header names are deprecated in RF7.0.
This is to remove warnings in log when running atests


emanlove (approved):
Thanks for fixing these Yuri!

PR #1890, opened 2024-03-10
Updated supported Python and Selenium versions

emanlove:
Within setup.py and the Github Actions workflow updated supported versions.


yuriverweij (commented):
Looks good, but selenium made some changes regarding Default_preferences on Firefoxprofile, which causes our unit tests to fail. I was looking into that, and it seems they no longer save the preferences we add to DEFAULT_PREFERENCES, but they are present in _desired_preferences. Changing that in the unit tests makes the test success again, but I'm not sure if that is the correct solution. Also that is not backwards compatible.

emanlove:
Resolved in another pull request.

PR #1891, opened 2024-04-16
Removed deprecation of `Press Key` keyword

emanlove:
Reverted keyword documentation on `Press Key`. Also added note telling about the differing underlying methods and to try the other if the one used does not work.

### DIFF
--- a/src/SeleniumLibrary/keywords/element.py
+++ b/src/SeleniumLibrary/keywords/element.py
@@ -899,7 +899,26 @@ return !element.dispatchEvent(evt);
 
     @keyword
     def press_key(self, locator: Union[WebElement, str], key: str):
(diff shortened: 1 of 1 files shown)


yuriverweij (approved):
Looks good to me. Would've been nice if we could've deprecated it, but I agree why we can't. "Press Key" could've been named "Send Keys To Element" to be more in line with Selenium, but I also see why that isn't practical since it's already part of the library for some time.

aaltat:
One can always write RoboTidy transformer to ease the transition from old keyword to new one

emanlove:
I would like to, eventually, transition the keyword, either to new name or a single keyword where one selects the underlying method used to send/press/input keys.

PR #1893, opened 2024-04-28
Fix doc to contains correct link syntax

aaltat:


aaltat:
Fixing this bug in doc
image

PR #1903, opened 2024-05-23
Specify supported Python version

jeromehuewe:
Version 6.4.0 only supports Python up to version 3.11


emanlove:
Thanks for catching this @jeromehuewe

PR #1923, opened 2024-10-07
Return base64 image in case of EMBED

hsj51:
This can be useful in case if someone wants to reuse the image i.e set test message or set suite documentation

### DIFF
--- a/src/SeleniumLibrary/keywords/screenshot.py
+++ b/src/SeleniumLibrary/keywords/screenshot.py
@@ -89,7 +89,7 @@ class ScreenshotKeywords(LibraryComponent):
         format string syntax].
 
         An absolute path to the created screenshot file is returned or if
-        ``filename``  equals to EMBED, word `EMBED` is returned.
+        ``filename``  equals to EMBED, base64 image is returned.
 
         Support for EMBED is new in SeleniumLibrary 4.2
 
@@ -126,7 +126,7 @@ class ScreenshotKeywords(LibraryComponent):
     def _capture_page_screen_to_log(self):
         screenshot_as_base64 = self.driver.get_screenshot_as_base64()
         self._embed_to_log_as_base64(screenshot_as_base64, 800)
-        return EMBED
+        return screenshot_as_base64
 
     @keyword
     def capture_element_screenshot(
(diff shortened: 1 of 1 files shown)


emanlove:
@hsj51 Could you explain a bit more the idea behind returning the string "EMBED" and give an example of how this change would be used?

hsj51:
Changes I have suggested is to return the base64 string of the image instead of the word 'EMBED'. This will enable user to reuse the image somewhere else in the report.

Below is an example use case:

Setting for EMBED
```
Library    SeleniumLibrary    screenshot_root_directory=EMBED
```
And add below lines to this function
https://github.com/robotframework/WebDemo/blob/c8d46736c8248c1ebdcd67ac77fcb5bdafdd4851/login_tests/resource.robot#L44
```
Welcome Page Should Be Open
...
    ${ss}=    Capture Page Screenshot
    Set Test Message    *HTML*Test Success<p><img src="data:image/png;base64,${ss}" width="256px">
```
Then this is what a report may look like
image

And Log file
image

This will allow an user to view the screenshots in the report itself and not search for it in the log file through a mountain of keywords.

Main goal for this change is reusability of the images. Current alternative for this is to use .png images, but then there will be multiple files to manage and sharing report become difficult.

hsj51:
@emanlove I have tried to explain my changes and a possible use cases. Please feel free to reach out for any more queries

emanlove:
Alright, I see better your use case. And appreciate you highlighting the issue of writing to file and then "there will be multiple files to manage and sharing report become difficult". I do wonder if the solution might be added functionality to rebot to do this in a post processing step? Let me think about this a bit more. I think having other opinions could also be good here. If you have an account on the Robot Framework community, could you share your idea and reasons for it within the #seleniumlibray channel. If you don't have an account I could post there as well.

adiralashiva8:
@hsj51 @emanlove We had similar requirement and applied same change in our local

_Observation_: Report size got increased from 3mb to 30mb due to embed screenshot in test message. Due to this report not loaded for long time 

We have around 2k tests and around 400 tests has embed images (failed tests)

I tried couple of approaches like flatten, remove keywords but report file size not decreased.

Thought of sharing my observation before accepting this change

emanlove:
@hsj51 and @adiralashiva8 if you both could take a look at a slightly modified version on this within https://github.com/robotframework/SeleniumLibrary/pull/1939. I wanted to prevent current users who might not need the returned string from increasing their log files by this change.

emanlove:
Fixed in #1939

PR #1924, opened 2024-10-07
Fix _find_by_data_locator

markus-leben:
Off by one in _find_by_data_locator which causes opaque failure on data values which contain a colon.

Example html:
`<div data-automation-id="foo:bar">This might be trouble</div>`

Example robot:
`Scroll Element Into View    data:automation-id:foo:bar`

Error message: 
`Provided selector (automation-id:foo:bar) is malformed. Correct format: name:value.`

I don't love the person putting colons in their automation ID, but we don't always get to choose the html we're testing.

### DIFF
--- a/src/SeleniumLibrary/locators/elementfinder.py
+++ b/src/SeleniumLibrary/locators/elementfinder.py
@@ -223,7 +223,7 @@ class ElementFinder(ContextAware):
 
     def _find_by_data_locator(self, criteria, tag, constraints, parent):
         try:
-            name, value = criteria.split(":", 2)
+            name, value = criteria.split(":", 1)
             if "" in [name, value]:
                 raise ValueError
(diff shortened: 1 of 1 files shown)


markus-leben:
On second thought it might be a good idea to also change the error raised in the try or the message given in the except. The error that gets caught by that catch is a ValueError about unpacking with string's split() method:
`ValueError: too many values to unpack (expected 2)`

But the error message from the except in _find_by_data_locator catches that ValueError and replaces it with something that's in this case more vague, and implies that the reason the ValueError is raised is because of the condition in the try, rather than an unrelated ValueError. 

I'm not sure what the ideal fix here would be though.

emanlove:
I have some unit tests to verify this change which I will submit in another PR.

PR #1925, opened 2024-11-01
Latest Versions Oct2024

emanlove:
- Fixes the cookie tests which need to be updated each year because Chrome, at least, does not allow cookies to expire further out than a year.
- Updates upload-artifacts to supported version 3. Have not moved to version 4 due to need to update artifact naming convention. (Will need to do this soon as V3 is also being deprecated)
- Updated Python version to include 3.13.0, robot framework v7.1.1, and selenium 4.24.0, 4.25.0, and 4.26.1

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -11,7 +11,7 @@ jobs:
       matrix:
         python-version: [3.8, 3.13.0]  # pypy-3.9
         rf-version: [6.1.1, 7.1.1]
-        selenium-version: [4.24.0, 4.25.0 4.26.1]
+        selenium-version: [4.24.0, 4.25.0, 4.26.1]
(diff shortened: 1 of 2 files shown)


aaltat (approved):

yuriverweij (approved):
Can we not calculate the cookie dates so this doesn't have to be updated every year?

aaltat:
> Can we not calculate the cookie dates so this doesn't have to be updated every year?

It should be possible with Robots DateTime library

emanlove:
Yeah .. we should. As I recall the one which expires in September, there is something about UTC time within the script/check that made it fail depending on where it was run. So on my local it would be one time and then in GitHub Action it was another. So knowing what time it was independent of the test became a pain. Will relook at this ..

PR #1939, opened 2025-05-26
Return screenshot as base64 string and embed into log

emanlove:
This is a slightly modified version of https://github.com/robotframework/SeleniumLibrary/pull/1923 which instead of modifying the EMBED functionality currently implemented, it allows for returning the base64 tring is one uses BASE64 instead. The reason for providing this alternative method is to prevent signifigant increase in filesize for the current users whom might not need the returned string.


aaltat (approved):

yuriverweij (approved):
Looks good to me

PR #1944, opened 2025-08-03
Latest versions jun2025

emanlove:
Various cleanup work on the internal test suite and bringing up the versions of parts of the tech stack.

### DIFF
--- a/atest/acceptance/keywords/alerts.robot
+++ b/atest/acceptance/keywords/alerts.robot
@@ -43,8 +43,7 @@ Handle Alert returns message
 
 Handle Alert with custom timeout
     Click Button    Slow alert
-    Sleep  0.1s
-    Handle Alert    timeout=1s
+    Handle Alert    timeout=2s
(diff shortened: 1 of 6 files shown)


atest/acceptance/keywords/alerts.robot
```diff
@@ -43,6 +43,7 @@ Handle Alert returns message
 
 Handle Alert with custom timeout
     Click Button    Slow alert
+    Sleep  0.1s
```
aaltat: Doesn’t this indicate problem in the library, instead of doing sleep, should library be fixed?

emanlove: Updated this. I removed the sleep but also extended the given timeout. For this test the alert was delayed 500ms and it appears that timing out at 1sec (or 500ms later) is too short to reasonable expect for alert triggered and then detected. This extension gives it a reasonable amount of time after triggering.

aaltat (approved):

Issue #1813, opened 2023-02-18
Relative Locators, iFrames, ShadowDom, RTL .. oh my!


emanlove:
I am opening up here a discussion on some fundamental design ideas around a set of locater related topics. There is a fair amount here so please read through the entire discussion. Also I understand this has both practical consequences and some deep technical discussions. I strongly feel it is important to take both into consideration and to have input from those approaching at both levels. If you come from the technical aspect please consider how the user will see and think about the solution. And if you come from the usage side don't be overwhelmed by the complexity presented here. Hopefully our solution will be clear and usable by all. Let me start with relative locators ..

## Relative Locators

Selenium 4 introduced [relative](https://www.selenium.dev/documentation/webdriver/elements/locators/) [locators](https://saucelabs.com/resources/articles/selenium-4-relative-locators). That is locators which you can say something like "find the element to the left of the image". In thinking about an implementation for SeleniumLibrary I thought about taking advantage of the existing [cascading or chained locators](https://robotframework.org/SeleniumLibrary/SeleniumLibrary.html#Locating%20elements). One design would be to put the relative locator direction within the chained strategy syntax of `>>`. For example

```
Click Element	css:.bar >leftof> xpath://a	# To find a link which is to the left of an element with class "bar"
```

I have made some really [initial test code](https://github.com/emanlove/robotframework-seleniumlibrary/blob/emanlove-relative-locators/utest/test/locators/test_elementfinder.py#L736) to try out the parsing of such locators. There is a usability/readability issue with the implementation shown in the example but I want to hold that discussion till I talk about RTL or Right-to-Left languages. Now I want to talk about frames ..

Someone recently shared an issue not solvable with the current library implementation. They wanted to drag and drop an element into an element within an iFrame. Because one needs to select a frame before interacting with an element within a frame and because the drag and drop keyword takes a "drag from" locator and then a target there isn't a way to do both. (Hopefully you are seeing the solution I am going to propose..) But if we could select a frame within a chained locator then we could overcome this short coming. My idea would be to introduce a new locator strategy `frame` which would be a "shortcut" for select frame. So in the problem outlined above one could do something like

```
Drag And Drop  css:#moveme  frame:frameOnPage >> css:#dragToTarget    # Drag the element with id moveme to the element dragToTarget within the frame named "frameOnPage"
```

One extra design consideration with frames is if one has a frame within another frame should we require them to step through the frame as in

```
[...]

Issue #120, opened 2012-09-24
Drag and drop across iframe bounds is problematic


emanlove:
The drag events are all bound within the toolbar iframe.  Example need - "We have a scenario here where we are grabbing a drag source in an iframe and dropping on a drag target in the parent window. I can 'Select Frame' to get the drag target, and Unselect Frame to return to the main window, but the Drag and Drop keyword doesn't appear to allow for that kind of context switching."

emanlove:
One solution is to drag and drop by offset. Having an source element and target element you should be able to get positions of each, using "Get Horizontal Position" and "Get Vertical Position" keywords. From there determine offset. **_Remember**_ though the click keywords uses the center of the element so you might also need size of element to properly calculate offset.

emanlove:
There is also a 'Mouse Down' keyword as well as a 'Mouse Up' keyword that might be used in conjunction to deal with the stated scenario.

aaltat:
Ed, is this still an issue or could we close this one? The frames are getting quite rare in these days, although there is one part in our application where frames are still being used.

emanlove:
I haven't tried in a while nor needed to try it. In the effort to close old issues let me say...

If anyone feels that this is still valid issue or this becomes an issue again, please send an update, reopen the ticket or create a new one.

PR #1953, opened 2025-12-14
Add "Drag And Drop Across Frames" keyword for cross-frame drag and drop support

b-vamsipunnam:
Summary

This PR adds a new keyword `Drag And Drop Across Frames` to support drag-and-drop interactions where the source and target are in different frame contexts.

Motivation

While automating post-deployment flows in a Salesforce Service Cloud setup, I encountered a limitation where drag-and-drop failed when the drop target was inside an iframe. Existing keywords do not support this scenario because WebDriver action chains lose context when switching frames mid-action.

Real-world scenario (Salesforce Service Cloud):  
Login → Setup → All Sites → Builder → Select Dashboard → Drag and drop a component into a target section (inside an iframe) → Publish

The existing `Drag And Drop` keyword fails in this case because the target resides inside an iframe.

This aligns with known limitations:
- Issue #120: cross-frame drag-and-drop challenges  
- Related discussion #1813 around iframe handling

This keyword enables reliable end-to-end automation for such scenarios.

Changes

- Added implementation in src/SeleniumLibrary/keywords/element.py
- Added acceptance tests in atest/acceptance/keywords/draganddropframe.robot
- Added test HTML page atest/resources/html/frames/draganddrop.html
- Updated unit test keyword count in utest/test/api/test_plugins.py

All acceptance tests pass locally, including the new cross-frame scenarios.

All acceptance tests pass locally (including new ones).

This addresses a long-standing limitation around cross-frame drag-and-drop. If applicable, please consider adding the ``acknowledgment`` label to #120 for release notes. Thank you!


b-vamsipunnam:
Hi @emanlove ,
I’ve updated the PR to align with the project’s keyword signature to use the new Locator type alias, and the PR is now scoped to:
- The new Drag And Drop To Frame keyword implementation
- Then updated acceptance tests

All tests are now passing in GitHub Actions. Please take another look when you have a chance. Thanks!

src/SeleniumLibrary/keywords/element.py
```diff
@@ -1293,4 +1293,41 @@ def get_css_property_value(
...
+        action = ActionChains(self.driver, duration=self.ctx.action_chain_delay)
+        action.click_and_hold(source_element).perform()
+
+        try:
+            frame_element = self.find_element(frame)
+            self.driver.switch_to.frame(frame_element)
+            target_element = self.find_element(target)
+
+            action = ActionChains(self.driver, duration=self.ctx.action_chain_delay)
+            action.move_to_element(target_element).release().perform()
```
yuriverweij: If the `find_element()` for frame or target raises an error, the mouse button is still held down. The finally switches context back but never releases. The next action in the suite inherits a held mouse button.

atest/acceptance/keywords/draganddropframe.robot
```diff
@@ -0,0 +1,47 @@
...
+Drag And Drop To Frame Hides Source Element
+    [Documentation]    Verifies that the source element becomes hidden after a successful drop.
+    Wait Until Page Contains Element    id=source    timeout=10s
+    Drag And Drop To Frame    id=source    id=target    id=previewFrame
+    Element Should Not Be Visible    id=source
+
+Standard Drag And Drop Fails When Target Is Inside Frame
+    [Documentation]    Verifies that the standard Drag And Drop keyword cannot complete this cross-frame scenario.
+    Wait Until Page Contains Element    id=source    timeout=10s
+    Run Keyword And Expect Error    *    Drag And Drop    id=source    id=target
```
yuriverweij: I would not wildcard the error that is expected. This can cause the test to "pass" when any error occurs.

atest/acceptance/keywords/draganddropframe.robot
```diff
@@ -0,0 +1,47 @@
...
+    Select Frame    id=previewFrame
+    Element Should Not Contain    id=target    Dropped Successfully!
+    Unselect Frame
+
+Drag And Drop To Frame Fails With Invalid Frame
+    [Documentation]    Verifies that the keyword fails when the frame locator is invalid.
+    Wait Until Page Contains Element    id=source    timeout=10s
+    Run Keyword And Expect Error    *    Drag And Drop To Frame
```
yuriverweij: I would not wildcard the error that is expected. This can cause the test to "pass" when any error occurs. 
*Element with locator 'id=missingFrame' not found*

atest/acceptance/keywords/draganddropframe.robot
```diff
@@ -0,0 +1,47 @@
...
+    ...    id=source    id=target    id=missingFrame
+
+Drag And Drop To Frame Fails With Invalid Target
+    [Documentation]    Verifies that the keyword fails when the target element is not found inside the iframe.
+    Wait Until Page Contains Element    id=source    timeout=10s
+    Run Keyword And Expect Error    *    Drag And Drop To Frame
```
yuriverweij: I would not wildcard the error that is expected. This can cause the test to "pass" when any error occurs.
*Element with locator 'id=missingTarget' not found*

src/SeleniumLibrary/keywords/element.py
```diff
@@ -1293,4 +1293,41 @@ def get_css_property_value(
         | ${color}= | `Get CSS Property Value` | css:button.submit | background-color |
         | ${size}=  | `Get CSS Property Value` | id:username       | font-size        |
         """
-        return self.find_element(locator).value_of_css_property(css_property)
\ No newline at end of file
+        return self.find_element(locator).value_of_css_property(css_property)
+
+    @keyword('Drag And Drop To Frame')
+    def drag_and_drop_to_frame(
+            self, locator: Locator, target: Locator, frame: Locator,
```
yuriverweij: In order to make the keyword more generic it might be worth changing some parts:

`def drag_and_drop_across_frames(
        self, locator: Locator, 
        target: Locator,
        target_frame: Locator, 
        source_frame: Optional[Locator] = None,
`

- ``locator``: Locator of the source element to drag.
- ``target``: Locator of the drop target.
- ``target_frame``: Locator of the iframe containing the target.
- ``source_frame``: Locator of the iframe containing the source, or None if the source is in the default content.

atest/resources/html/frames/draganddrop.html
yuriverweij, line 1: Test fixture does not seem to work as intended.
Right now:
When clicking on the "Drag Me!" box, and performing a mouseup (releasing the mouse button anywhere) will trigger "Dropped Successfully", regardless of actually moving/draggint the box.

The test would pass even if move_to_element(target_element) did nothing.
The test would pass even if the frame switch silently failed (as long as a mouseup fires somewhere).
The test would pass if you replaced the whole keyword body with click_and_hold().release() on the source.

The fixture needs to validate that the drop actually lands on the target inside the frame,  e.g. using real HTML5 DnD events or at least checking the mouseup coordinates hit the target's bounding box. As written, these tests give false confidence

yuriverweij (changes requested):
Thanks for this PR! it looks very promising, but there are a few things that need to be resolved.

- There is a bug in the try/except which can cause the mouse button to stay down.
- The test fixture (html file) has some problems which can give false positives 
- Suggestion: the keyword can be made a bit more generic allowing dragging from frame to frame or default context to frame. (in this case the Note can be removed from the docstring)
- Suggestion: add the following to the docstring/description `After this keyword runs, the browser context is always default content, regardless of the context before the call.`

atest/resources/html/frames/draganddrop.html
b-vamsipunnam: Hi @yuriverweij, thanks for the detailed feedback. I agree with the suggestions and I’m working on the updates. I’ll push the changes shortly.

src/SeleniumLibrary/keywords/element.py
b-vamsipunnam: @yuriverweij : I’ve updated the keyword to ensure the mouse is always released if anything fails after the click-and-hold. The `finally` block now handles the release.

atest/acceptance/keywords/draganddropframe.robot
b-vamsipunnam: @yuriverweij : I’ve updated the tests to assert the specific expected error instead of using a wildcard. Thank you.

atest/acceptance/keywords/draganddropframe.robot
b-vamsipunnam: @yuriverweij : I’ve updated the tests to assert the specific expected error instead of using a wildcard. Thank you.

atest/acceptance/keywords/draganddropframe.robot
b-vamsipunnam: @yuriverweij : @yuriverweij : I’ve updated the tests to assert the specific expected error instead of using a wildcard. Thank you.

src/SeleniumLibrary/keywords/element.py
b-vamsipunnam: @yuriverweij : I’ve updated the keyword signature and documentation to make it more generic:`locator`, `target`, `target_frame`, and optional `source_frame`.This now supports both default-content to frame and source-frame to target-frame scenarios. Thanks for the suggestion. This has now become a robust keyword.

atest/resources/html/frames/draganddrop.html
b-vamsipunnam: @yuriverweij : I updated the fixture so it only completes the drop when the mouseup coordinates are inside the target element within the target iframe. This should prevent false positives from a simple click-and-release on the source. Thanks you.

b-vamsipunnam:
Hi @yuriverweij

Thanks for the detailed feedback.

I’ve updated the fixture so the drop completes only when the mouseup occurs within the target element inside the target iframe, avoiding false positives.

The keyword now also supports an optional source frame and ensures the mouse is released on failure before resetting context.

@emanlove: If this looks good, could you please consider adding the ``acknowledgment`` label to #120 for release notes? Thank you.

atest/resources/html/frames/draganddrop.html
```diff
@@ -2,94 +2,262 @@
...
+
+<div id="defaultSource">Default Source</div>
+
+<div class="frame-wrapper">
+    <iframe id="sourceFrame"></iframe>
+    <iframe id="targetFrame"></iframe>
+</div>
+
+<div id="dragGhost"></div>
+<div id="status">Not dropped</div>
```
yuriverweij: Nitpicking/Suggestion:
`#status` isn't used in the tests, instead they assert on the `id=target` (both are updated on successful drop) 

So you could remove the `#status` entirely, or use it in the tests. 
I think it is best to keep the test fixture as small and clean as possible.

atest/acceptance/keywords/draganddropframe.robot
```diff
@@ -1,47 +1,87 @@
...
+    Wait Until Page Contains Element    id=sourceFrame    10s
+    Select Frame    id=sourceFrame
+    Wait Until Page Contains Element    id=frameSource    10s
+    Unselect Frame
+    Drag And Drop Across Frames    id=frameSource    id=target    id=targetFrame    id=sourceFrame
+    Select Frame    id=targetFrame
+    Element Should Contain    id=target    Dropped Successfully!
+    Unselect Frame
+
+Drag And Drop Across Frames Returns To Default Content
```
yuriverweij, line 27: This test doesn't really verify what it claims. 

`Element Should Be Visible    id=targetFrame` checks that the iframe element is visible, but the iframe element lives in the default content regardless of which frame context Selenium currently has selected. 
If the keyword left us stuck inside the target frame, this assertion would either still pass (because the lookup falls through) or fail for an unrelated reason. Neither outcome proves the context was restored.

Suggested fix: assert on something that's not reachable when we're actually back in default content:
`Page Should Not Contain Element    id=target`
`id=target` only exists inside `targetFrame`. If the keyword correctly returned to default content, the test passes.
It would also catch a bug in case we're accidentally in some other non-default frame.

atest/acceptance/keywords/draganddropframe.robot
```diff
@@ -1,47 +1,87 @@
...
+
+Drag And Drop Across Frames Hides Frame Source Element
+    [Documentation]    Verifies that the frame source element becomes hidden after a successful drop.
+    Wait Until Page Contains Element    id=sourceFrame    10s
+    Drag And Drop Across Frames    id=frameSource    id=target    id=targetFrame    id=sourceFrame
+    Select Frame    id=sourceFrame
+    Element Should Not Be Visible    id=frameSource
+    Unselect Frame
 
 Standard Drag And Drop Fails When Target Is Inside Frame
```
yuriverweij: I'd suggest dropping this test. It's not testing the new Drag And Drop Across Frames keyword, it's testing the standard `Drag And Drop` keyword limitations.

yuriverweij (changes requested):
On test doesn't do what it claims to do. For the rest it looks good to me. 

Maybe @emanlove still has some other suggestions/feedback and I'll let him have the final say.

atest/resources/html/frames/draganddrop.html
b-vamsipunnam: @yuriverweij : As suggested, I removed the unused #status.

atest/acceptance/keywords/draganddropframe.robot
b-vamsipunnam: @yuriverweij : as suggested, I have removed this test. Thank you.

atest/acceptance/keywords/draganddropframe.robot
b-vamsipunnam: @yuriverweij : You are right.  I have updated the test to assert on an element only available inside the iframe, which validates that the context is correctly reset to default content.

b-vamsipunnam:
Hi @yuriverweij : Thanks for all the feedback. I’ve addressed the comments and made the suggested updates. Happy to make any further changes if needed.

Thank you.

yuriverweij (approved):
Looks good to me!

emanlove (approved):

PR #1954, opened 2026-01-18
Strip whitespaces from element text should be

emanlove:
This is @jannek76 pull request with some adjustments for the changed library.


atest/acceptance/keywords/cookies.robot
yuriverweij, line 1: Could we not generate a expiry date in the future and use that instead of having to update this every year?

src/SeleniumLibrary/__init__.pyi
yuriverweij, line 1: We should create a new type for the locator, this is a lot of duplication.

```
from typing import TypeAlias, Union

Locator: TypeAlias = WebElement | str
```

yuriverweij (approved):

atest/acceptance/keywords/cookies.robot
emanlove: That would be good although I tend to push this off every year till it is that time of year. One problem is expiry dates can no longer be longer than a year. So it is not like we can set it so far into the future that it will just never expire.

yuriverweij: how about this?

```
*** Test Cases ***
Add Cookie With Python Calculation
    ${expiry} =    Evaluate    int((datetime.datetime.now() + datetime.timedelta(days=330)).timestamp())    modules=datetime
    Add Cookie    Cookie1    value1    expiry=${expiry}
```

Issue #1962, opened 2026-01-19
Chaining locators with lists does not work with Robot Framework 6.1 or above


jjaakkola-atostek:
Giving chained locator as a list of locators seems not to work since Robot Framework 6.1.0. Tested with Chrome and Firefox using test web page and test case given below. The first chained locator with >> finds the element, but latter with list does not . With Robot Framework 6.0.1 works fine. Tested also with RF 7.4.1 and not working.

I did some research in the SeleniumLibrary's code, and problem seems to be with the type annotations of the keywords. For example:

```
def click_element(
        self,
        locator: Union[WebElement, str],
        modifier: Union[bool, str] = False,
        action_chain: bool = False,
    ):
```
`locator`'s type annotation should be `Union[WebElement, str, list]´. If I chance that, my test code passes.

### Software and versions ###
SeleniumLibrary: 6.8.0
Robot Framework: 6.1 and 7.4.1
robotframework-pythonlibcore: 4.5.0
Python: 3.11.9
Chrome: 143.0.7499.171
ChromeDriver: 143.0.7499.40
Firefox: 147.0.1
Gecko Driver: 0.36.0

### Example HTML ###
```
<html>
	<head>
		<title>Locator Chain Test</title>
	</head>
	<body>
		<button id="testElement">This is a test</button>
	</body>
</html>
```

### Example test case ###
```
*** Settings ***
Library	SeleniumLibrary

*** Test Cases ***
Locator Chain Test
	Open Browser		[test_file_path]	chrome
	Click Element		tag:body >> id:testElement
	${locator_chain}	Create List	tag:body	id:testElement
	Click Element		${locator_chain}
```

emanlove:
Will take a look this morning. What version of pythonlibcore (robotframework-pythonlibcore) are you running? The latest is 4.5.0 released just this past week.

jjaakkola-atostek:
Thank you! robotframework-pythonlibcore version is 4.5.0. I added it to the description.

emanlove:
I think I am seeing the problem and yes it apopears to be the type annotation. I was temporarily confused because I know we had a test for this which was passing. Looking at the code who have just the one specific test, which should be good enough, but it was just the one keyword it uses, `Page Should Contain Element`, that has the type annotations fixed. Looks like most if not all other keywords with locators do not ..

Working on a solution now. Thanks for pointing this out @jjaakkola-atostek !

emanlove:
Fixed in #1972

PR #1963, opened 2026-01-19
#1962 fix cascading locators

emanlove:
As noted in #1962 chaining or cascading locators has been broken for a while. This PR addresses the problem.


yuriverweij (changes requested):
We're already importing typing, so this could be simplified for us.

```
from typing import TypeAlias

Locator: TypeAlias = WebElement | str | list["Locator"]
```

The we can use 
``` locator: Locator```

emanlove:
Rethinking what you are saying Yuri and see your point. Thinking we could make that in `src/SeleniumLibrary/utils/types.py` and import in the various files from there. Honestly not sure what the difference is between using this TypeAlias versus making our own type. As I mentioned to tatu in a DM types are a weak point of mine that I need to learn more ..

emanlove:
Closing in favor of solution within #1972

PR #1972, opened 2026-04-06
1962 add locator type

emanlove:
Reowrked previous solution to fix cascading locators with this using Type Alias instead of modifying the individual types at each usage of locator.

This solution's syntax requires a minimum Python version of 3.10.


src/SeleniumLibrary/keywords/tableelement.py
```diff
@@ -13,19 +13,20 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from typing import Union
+from typing import Union, List
```
aaltat: Is List used somewhere?

src/SeleniumLibrary/keywords/waiting.py
```diff
@@ -16,14 +16,15 @@
 
 import time
 from datetime import timedelta
-from typing import Optional, Union
+from typing import Optional, Union, List
```
aaltat: Is List used somewhere?

aaltat (approved):

src/SeleniumLibrary/keywords/formelement.py
```diff
@@ -15,18 +15,19 @@
...
 
 from SeleniumLibrary.base import LibraryComponent, keyword
 from SeleniumLibrary.errors import ElementNotFound
+from SeleniumLibrary.utils.types import Locator
 
 
 class FormElementKeywords(LibraryComponent):
     @keyword
-    def submit_form(self, locator: Union[WebElement, None, str] = None):
+    def submit_form(self, locator: Union[WebElement, None, str, List[Union[WebElement,str]]] = None):
```
yuriverweij: This would probably be better written as:  

`def submit_form(self, locator: Locator | None = None):`

src/SeleniumLibrary/keywords/element.py
```diff
@@ -930,7 +930,7 @@ def press_key(self, locator: Union[WebElement, str], key: str):
         element.send_keys(key)
 
     @keyword
-    def press_keys(self, locator: Union[WebElement, None, str] = None, *keys: str):
+    def press_keys(self, locator: Union[WebElement, None, str, List[Union[WebElement,str]]] = None, *keys: str):
```
yuriverweij: This would probably be better written as:

`def press_keys(self, locator: Locator | None = None):`

yuriverweij (changes requested):
Looks good! Minor change requested to keep it all in line. Should work for 3.10 compatibility

When locator can also be none would probably be better written as:
locator: Locator | None = None

src/SeleniumLibrary/keywords/tableelement.py
emanlove: 🤔 Mmm .. time to implement that linter ..

aaltat: It is there: https://github.com/robotframework/SeleniumLibrary/blob/master/tasks.py#L189. Just use it. 

Although now days, I would use ruff format and ruff check.

emanlove: Cleaned these up including a few others

src/SeleniumLibrary/keywords/formelement.py
emanlove: Modified this to use new Locator type alias

src/SeleniumLibrary/keywords/element.py
emanlove: Modified this to use new Locator type alias

yuriverweij (approved):

aaltat (approved):

Issue #1884, opened 2024-01-20
Review restriction on running ApprovalTests under Windows


emanlove:
Noting [this issue](https://github.com/approvals/ApprovalTests.Python/issues/41) which Tatu raised it looks like we should now be able to run the ApprovalTests on Windows. This would be helpful as I sometime am developing on Windows and this would help me to see issue sooner than later.

JayBazuzi:
https://github.com/approvals/ApprovalTests.Python/issues/41 is now closed. You should be able to remove https://github.com/robotframework/SeleniumLibrary/blob/c5c21ea2eb44d7e76619f96210d1f28c859f3aee/src/SeleniumLibrary/utils/types.py#L24-L25

Let us know if there's still a problem.

emanlove:
Thanks @yuriverweij for looking at this and resolving!

Issue #1787, opened 2022-10-05
Review create webdriver code and tests


emanlove:
With the recent removal of the Opera references I noticed some discrepancies and some things that just didn't look right within this code. I want to review that code now there are some significant changes. This is a placeholder reminder.

emanlove:
Also review for phantomJS, Andriod, and iPhone references as well

emanlove:
Another item we should check is the parameters for creating webdriver. I see an additional parameter `service` on the [Chrome](https://www.selenium.dev/selenium/docs/api/py/webdriver_chrome/selenium.webdriver.chrome.webdriver.html#module-selenium.webdriver.chrome.webdriver) and [Chromium WebDriver](https://www.selenium.dev/selenium/docs/api/py/webdriver_chromium/selenium.webdriver.chromium.webdriver.html#selenium.webdriver.chromium.webdriver.ChromiumDriver) classes that I am not sure we are either describing (very likely) or passing along (I think unlikely).

simonmeggle:
Hi Ed, I also notice that creating a webdriver does not work as expected (Selenium 4.7.2, SeleniumLibrary 6.0.0). 

```
# (The browser type does not really matter)
Create Webdriver  chromium
```
calls `browsermanagement > BrowserManagementKeywords > create_webdriver`

https://github.com/robotframework/SeleniumLibrary/blob/24c732b1628420ab8fd016424489ba1d63667503/src/SeleniumLibrary/keywords/browsermanagement.py#L383-L390

`creation_func` should return the browser specific webdriver class (instantiated in line 391), but all it returns is the module  <module 'selenium.webdriver.chromium'>:

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
TypeError: 'module' object is not callable
```

(its `__init__.py` is empty, no code runs)

However, while debugging, when I am assigning the Chromium Webdriver class by hand, `creation_func` starts the browser:

```
# (halted in line 391) 
from selenium.webdriver.chrome.webdriver import WebDriver
creation_func = WebDriver
creation_func(**init_kwargs)
# (chromium starts successfully)
<selenium.webdriver.chrome.webdriver.WebDriver (session="62a68ae2a4b6687ae0f76df8cc45883e")>
```

I am far from being able to judge whether my approach is correct or to decide what exactly is the error. 
But yes, there is something broken (but you already knew that before ;-) )

emanlove:
[...]

PR #1978, opened 2026-04-25
Migrate lint/format tooling to Ruff, update CI matrix, and stabilize cross-platform/firefox tests

yuriverweij:
This branch replaces legacy lint/format tooling with Ruff, updates CI and runtime compatibility targets, and cleans up platform/browser-specific test behavior so checks are consistent across environments.

Running the utests, atests, linting, and formatting should now work consistently on Windows and Unix based devices.

**Major changes**

- Replaced Flake8 and Black with Ruff for linting and formatting.
- Added/updated Invoke tasks for linting and formatting in `tasks.py`
- ❗Added/updated Ruff lint and format checks to GitHub workflow
- ❗ Configured lint/format workflow to run both format and lint steps, then fail the job if either step fails.
- ⚠️ Re-enabled Firefox testing in CI and marked one unstable mouse-over acceptance case as a known issue for Firefox in `mouse.robot`
- ⚠️ Re-enabled previously skipped Windows unit tests and removed unneeded import, see #1884 
- Updated minimum supported Python version to 3.10 and aligned GitHub workflows accordingly.
- Refactored type hints to Python 3.10+ syntax (for example, union operator style).
- Replaced collections.namedtuple usages with typing.NamedTuple in affected unit-test helpers/constructors.
- Applied import-order and formatting cleanups required by Ruff rules.
- Added SKIP_ON_WINDOWS tag for some atests that would not evaluate successfully due to Windows file paths and logging changes. 

**Documentation updates**

- Updated contributor guidance for new lint/format commands in `CONTRIBUTING.rst`

**Behavioral impact**

- No intended product/runtime behavior changes in SeleniumLibrary keywords.
- Main impact is development workflow, CI behavior, typing/style modernization, and test-platform handling.

**CI and quality checks**

- Ruff format check now runs in check mode with diff output.
-  Ruff lint check runs in the same workflow.
-  Workflow reports both step outcomes and fails if one or both fail.
- Firefox is re-enabled with known-issue handling for the unstable case.

**Notes**

- atest/ folder linting is not enabled yet, needs work. (formatting is enabled)
- The Firefox known issue should be tracked for eventual removal once root cause is resolved.
- SKIP_ON_WINDOWS tests are still valid, but produce different results on windows, or fail because of unix-oriented paths in the tests. This concerns ~12 tests and I thought the risk minimal to skip these on Windows.


.github/workflows/LintFormatCheck.yml
aaltat, line 1: Does this needs to be separate workflow? It could be just a one step in the existing workflow? Browser library does it as a part of the normal CI workflow. Saves spinning up one runner and in practice all devs are installed already in the existing workflow.

.github/workflows/CI.yml
```diff
@@ -9,8 +9,8 @@ jobs:
     continue-on-error: true
     strategy:
       matrix:
-        python-version: [3.9.23, 3.13.5, 3.14.0-rc.3]  # pypy-3.9
-        # python-version: [{earliest: 3.9}, {latest: 3.13.0}]  # pypy-3.9
+        python-version: [3.10.16, 3.13.5, 3.14.0-rc.3, pypy-3.10]
```
aaltat: May want to use released Python 3.14

.github/workflows/LintFormatCheck.yml
yuriverweij: Good point, I've now moved it into a separate job in the existing workflow

.github/workflows/CI.yml
yuriverweij: Updated the python version, also updated some other versions to latest.

tasks.py
```diff
@@ -186,11 +185,29 @@ def init_labels(ctx, username=None, password=None):
...
+def lint(ctx, fix=False):
+    """Run Ruff lint checks.
 
+    Args:
+        fix: Apply safe fixes when True. Defaults to False.
+    """
+    cmd = f"{sys.executable} -m ruff check --config pyproject.toml tasks.py src/ utest/" # atest/"
+    if fix:
+        cmd = f"{cmd} --fix"
+    ctx.run(cmd)
```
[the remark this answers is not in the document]
yuriverweij: leave for now

atest/resources/testlibs/get_driver_path.py
```diff
@@ -30,12 +31,12 @@ def get_driver_path(browser):
...
-    if ('service' in args) and ('options' in args):
+    if ("service" in args) and ("options" in args):
         # Selenium V4.20.0 or greater
         finder = driver_finder.DriverFinder(service.Service(), options.Options())
         return finder.get_driver_path()
     else:
         # Selenium v4.19.0 and prior
         finder = driver_finder.DriverFinder()
-        func = getattr(finder, 'get_path')
+        func = getattr(finder, "get_path")
```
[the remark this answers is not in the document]
yuriverweij: wont fix

atest/acceptance/keywords/textfields.robot
```diff
@@ -81,5 +81,10 @@ Attempt Clear Element Text On Non-Editable Field
...
     [Arguments]    ${alias}=${None}
-    Open Browser    ${FRONT PAGE}    ${BROWSER}    remote_url=${REMOTE_URL}
-    ...    options=add_experimental_option("prefs", {"profile.password_manager_leak_detection": False})    alias=${alias}
\ No newline at end of file
+    ${browser}=    Evaluate    "${BROWSER}".replace(" ", "").lower()
+    IF    "${browser}" in ["chrome", "googlechrome", "gc", "headlesschrome"]
+        Open Browser    ${FRONT PAGE}    ${BROWSER}    remote_url=${REMOTE_URL}
+        ...    options=add_experimental_option("prefs", {"profile.password_manager_leak_detection": False})    alias=${alias}
+    ELSE
+        Open Browser    ${FRONT PAGE}    ${BROWSER}    remote_url=${REMOTE_URL}    alias=${alias}
```
emanlove, line 89: If this test is designed to check chrome then should we be skipping it instead of running a non-test?

atest/acceptance/entry_point.robot
```diff
@@ -17,18 +17,18 @@ Entry Point Version
 
 Entry Point Translation
     ${process} =    Run Process
-    ...    python -m SeleniumLibrary.entry translation ${OUTPUT_DIR}/translation.json
+    ...    python -m SeleniumLibrary.entry translation ${OUTPUT_DIR}${/}translation.json
```
emanlove, line 20: What is the RF standard for dealing with paths? I know I had once recommended switching `/` to `${/}` but I recall someone else giving a strong argument for these to remain `/` noting that it will work for both Windows and Linux/MacOS (although I don't recall the underlying mechanisms. I thought this would fail on windows).

emanlove: Less concerned about this specific PR. I am good with changing this but trying to understand when we lint what the recommended results should be.

atest/acceptance/multiple_browsers_options.robot
```diff
@@ -8,13 +8,15 @@ Documentation     Creating test which would work on all browser is not possible.
 
 *** Test Cases ***
 Chrome Browser With Selenium Options As String
+    [Tags]    SKIP_ON_WINDOWS
```
emanlove, line 11: Why are these skip on windows?

emanlove: [An aside, I had always wanted an easy to document but clear documented explanation as to why for wa  particular test we would want to skip on a certain OS. Or if a known issues what change would trigger a reevalaution for the skip/ignore reason. That way when one came along latter one could evaklute if this was still the case.]

yuriverweij: These atests test on specific lines in the logging. Due to Windows specific stuff there are additional debug lines at the start which messes these up. So we could fix them, but this would mean the assertions will have to change quite a lot. Since this is not windows-specific functionality I figured this would not pose much of a risk if we skip these on Windows.

src/SeleniumLibrary/entry/__main__.py
```diff
@@ -39,7 +38,6 @@ def cli():
 
     See each command argument help for more details what (optional) arguments that command supports.
     """
-    pass
```
emanlove, line 42: I don't like this Linter rule. I think ity make it less clear. Does the linter accept `...` instead?

atest/acceptance/entry_point.robot
yuriverweij: I think for most paths I would (and did) not change this to `${/}`, since windows will be able to handle `/`. In this specific case it did make a difference because it's output from the `stdout` that is string compared in the test.

src/SeleniumLibrary/keywords/webdrivertools/sl_file_detector.py
```diff
@@ -33,12 +33,11 @@ def choose_file(self):
...
-            return True
-        return False
+        return bool(sl and sl._running_keyword == "choose_file")
 
     def _get_sl(self):
         libraries = BuiltIn().get_library_instance(all=True)
         for library in libraries:
             if isinstance(libraries[library], SeleniumLibrary.SeleniumLibrary):
                 return libraries[library]
+        return None
```
emanlove, line 43: Does this affect any downstream usage of this method? Or is all this doing is explicitly setting the return to `None`?

atest/acceptance/keywords/textfields.robot
yuriverweij: Not a test, A keyword that is being used bu the tests, in case of Firefox/non-chrome this gives an error, since experimental options does not exist.

src/SeleniumLibrary/keywords/webdrivertools/webdrivertools.py
```diff
@@ -40,22 +39,20 @@
...
-        "firefox": "firefox",
-        "headlessfirefox": "headless_firefox",
-        "ie": "ie",
-        "internetexplorer": "ie",
-        "edge": "edge",
-        "safari": "safari",
-    }
-
     def __init__(self, log_dir):
+        self.browser_names = {
```
emanlove: Do we want this as a class attribute (as was) or should it be an instance attribute (as changed to)?

src/SeleniumLibrary/keywords/alert.py
```diff
@@ -32,7 +31,7 @@ class AlertKeywords(LibraryComponent):
 
     @keyword
     def input_text_into_alert(
-        self, text: str, action: str = ACCEPT, timeout: Optional[timedelta] = None
+        self, text: str, action: str = ACCEPT, timeout: timedelta | None = None
```
emanlove, line 34: leaving note to self to relook at this ..

src/SeleniumLibrary/keywords/alert.py
```diff
@@ -146,7 +145,11 @@ def _wait_alert(self, timeout=None):
         wait = WebDriverWait(self.driver, timeout)
         try:
             return wait.until(EC.alert_is_present())
-        except TimeoutException:
-            raise AssertionError(f"Alert not found in {secs_to_timestr(timeout)}.")
+        except TimeoutException as original_exception:
```
emanlove, line 148: Is this a linter syntax or an AI syntax?

src/SeleniumLibrary/entry/__main__.py
yuriverweij: No, also not accepted.
I can add `# noqa: PIE790` to suppress the linting rule here.

src/SeleniumLibrary/keywords/webdrivertools/sl_file_detector.py
yuriverweij: Does not affect at all. Just explicitly returns `None`, instead of doing it anyway if the `for/if` loop does not return anything. So normally would return `None` anyway if it reached here, but difficult to debug since it's not explicit.

src/SeleniumLibrary/keywords/formelement.py
```diff
@@ -239,9 +238,7 @@ def choose_file(self, locator: Locator, file_path: str):
             self.ctx._running_keyword = None
 
     @keyword
-    def input_password(
-        self, locator: Locator, password: str, clear: bool = True
-    ):
+    def input_password(self, locator: Locator, password: str, clear: bool = True):
```
emanlove, line 241: Why does it in one instance put all the arguments into one line. And then elsewhere want one argument per line?

src/SeleniumLibrary/keywords/webdrivertools/webdrivertools.py
yuriverweij: You're right, this would make more sense as a class attribute.

src/SeleniumLibrary/keywords/alert.py
yuriverweij: Fom linter (rule B904), basically to better debug where the exception originated. In this case: instead of getting an assertion error that has something timeout in it. you'd get the explanation with it that the timeout exception caused the assertion error, preserving the full traceback

src/SeleniumLibrary/keywords/screenshot.py
```diff
@@ -35,7 +34,7 @@
 
 class ScreenshotKeywords(LibraryComponent):
     @keyword
-    def set_screenshot_directory(self, path: Union[None, str]) -> str:
+    def set_screenshot_directory(self, path: None | str) -> str:
```
emanlove, line 37: This is an example oif why i think I like the syntax `Optional[str]` over `str | None`. That is if those are reversed as here, `None | str` that would seem to have a different meaning.

Maybe these are equivalent anyways and it is just that in the Optional case it has a default value .. 🤔

src/SeleniumLibrary/keywords/formelement.py
yuriverweij: Basically comes down to 80 character line limit. If it exceeds it splits all the arguments to its own line.  (all or nothing)

src/SeleniumLibrary/keywords/window.py
```diff
@@ -239,7 +237,8 @@ def set_window_size(self, width: int, height: int, inner: bool = False):
         | `Set Window Size` | 800 | 600 | True |
         """
         if is_falsy(inner):
-            return self.driver.set_window_size(width, height)
+            self.driver.set_window_size(width, height)
```
emanlove, line 240: 🔴 This change looks like two different return values

src/SeleniumLibrary/__init__.py
```diff
@@ -689,10 +691,13 @@ def get_keyword_documentation(self, name: str) -> str:
             return self._get_intro_documentation()
         return DynamicCore.get_keyword_documentation(self, name)
 
+    class Doc(NamedTuple):
```
emanlove, line 694: 🔴 This seems like it would provide differnent functionality as a class instance. Especially seince each intance of a plug in I would expect to be different ..

src/SeleniumLibrary/keywords/screenshot.py
yuriverweij: These are 100% equivalent. `None | str`, `str | None`, and `Optional[str]` all produce the same type at runtime. Union order is irrelevant for type checking.

`Optional[str]` is literally just an alias for `Union[str, None]` and the UP007 rule converts that to `str | None`

Readability and convention say to put primary type (`str`) first before `None`, so we could update that, but no functional difference.

src/SeleniumLibrary/__init__.py
```diff
@@ -817,12 +822,15 @@ def _parse_listener(self, event_firing_webdriver):
             raise DataError(message)
         return listener
 
+    class Module(NamedTuple):
```
emanlove, line 825: 🔴 .. the same here

utest/test/api/my_lib.py
```diff
@@ -20,3 +20,6 @@ def foo(self):
     @keyword
     def bar(self, arg):
         self.info(arg)
+
```
emanlove: 🔴  ??? Is this reworking how the documented way of how libraries work in RF

utest/test/api/my_lib_args.py
```diff
@@ -21,3 +21,6 @@ def bar_2(self, arg):
     def add_cookie(self, foo, bar):
         self.info(foo)
         self.info(bar)
+
+
+my_lib_args = MyLibArgs
```
emanlove: 🔴  ??? and here. Is this reworking how the documented way of how libraries work in RF

utest/test/api/my_lib_not_inherit.py
```diff
@@ -1,10 +1,13 @@
...
+class MyLibNotInherit:
     def __init__(self, ctx):
         self.ctx = ctx
 
     @keyword
     def bar(self, arg):
         self.info(arg)
+
+
+my_lib_not_inherit = MyLibNotInherit
```
emanlove: 🔴  ??? and here. Is this reworking how the documented way of how libraries work in RF

utest/test/api/plugin_tester.py
```diff
@@ -13,3 +13,6 @@ def foo(self):
     @keyword
     def bar(self, arg):
         self.info(arg)
+
+
+plugin_tester = PluginTester
```
emanlove, line 18: 🔴  ??? and here. Is this reworking how the documented way of how libraries work in RF

utest/test/api/plugin_with_event_firing_webdriver.py
```diff
@@ -32,3 +31,6 @@ def __init__(self, ctx):
     @keyword
     def tidii(self):
         self.info("foo")
+
+
+plugin_with_event_firing_webdriver = PluginWithEventFiringWebdriver
```
emanlove, line 36: 🔴  ??? and here. Is this reworking how the documented way of how libraries work in RF

utest/test/api/test_accessing_keywod_methods.py
```diff
@@ -1,5 +1,7 @@
 import unittest
 
+import pytest
```
emanlove, line 3: Is the suggestion to switch over to pytest? Is there any advantages/disadvantages to doing so? Are we gaining anything by doing so here?

utest/test/api/test_event_firing_webdriver.py
```diff
@@ -1,6 +1,7 @@
 import os
 import unittest
 
+import pytest
```
emanlove, line 4: .. same here. It looks like there was a selective switch on a small number of utest files and just changing how asserts are made. But now signifiant gain from what I can see.

utest/test/api/test_plugin_keyword_tags.py
```diff
@@ -16,33 +17,33 @@ def setUpClass(cls):
     def test_no_plugin(self):
         sl = SeleniumLibrary()
         tags = sl.get_keyword_tags("open_browser")
-        self.assertFalse(tags)
+        assert not tags
```
emanlove, line 20: ..as noted before about inclusion of pytest.

utest/test/api/test_plugins.py
```diff
@@ -1,7 +1,8 @@
-from collections import namedtuple
 import os
 import unittest
+from typing import NamedTuple
 
+import pytest
```
emanlove, line 5: ?pytest

utest/test/keywords/IGNOREDtest_webdrivercreator.py
```diff
@@ -172,7 +172,7 @@ def test_chrome_remote_no_caps(creator):
 def test_chrome_remote_caps(creator):
     url = "http://localhost:4444/wd/hub"
     expected_webdriver = mock()
-    # capabilities = {"browserName": "chrome"}
+    capabilities = {"browserName": "chrome"}
```
emanlove, line 175: As we don't use this instead of uncommenting it maybe it should just be deleted.

utest/test/keywords/IGNOREtest_webdrivercreator_executable_path.py
```diff
@@ -247,5 +249,4 @@ def mock_file_detector(creator):
 def get_geckodriver_log():
     # return os.path.join(LOG_DIR, "geckodriver-1.log")
     # print(f"{os.getcwd()}")
-    cwd = os.getcwd()
-    return cwd
\ No newline at end of file
+    return os.getcwd()
```
emanlove, line 252: .. I kind of like the other way with these as it allows one to describe what it is with a variable name. Assuming this is a linter correction?

utest/test/keywords/test_click_modifier.py
```diff
@@ -40,24 +40,19 @@ def test_parsing_multiple_modifiers(element):
 
 
 def test_invalid_modifier(element):
-    with pytest.raises(ValueError) as error:
+    with pytest.raises(ValueError, match="'FOO' modifier "):
```
emanlove, line 43: Assuming this accomplishes the same assertions?

utest/test/keywords/test_webdrivercache.py
```diff
@@ -1,6 +1,7 @@
 import unittest
 
-from mockito import mock, verify, when, unstub
+import pytest
```
emanlove, line 3: ??pytest

src/SeleniumLibrary/entry/__main__.py
emanlove: Is it possible to exclude this rule from the linter altogether instead of having to add `# noqa: PIE790` each time? It is also possible to argue that this is a fine rule and I shloud just get over my dislike for it :)

src/SeleniumLibrary/keywords/alert.py
emanlove: Ok, this was a new one to me and one I need to learn

src/SeleniumLibrary/keywords/screenshot.py
emanlove: Somehow to me it seems a little clearer when is has the word Optional there. Replacing union with `|` make sense as that is the union operator. But Optional seems to tell me something about it. It think this is one I just need to get used to seeing it differently ..

src/SeleniumLibrary/keywords/alert.py
emanlove: closing as your note about Optional explains what I was thinking ..

src/SeleniumLibrary/entry/__main__.py
yuriverweij: yes, we can also globally add it to the exclude. I will do that

src/SeleniumLibrary/keywords/window.py
yuriverweij: is_falsy does not return a value. In this early exit because of the `if` it would return the result of `set_window_size` which is `None`, but in the normal exit there is no return value. To make it more conistent the early exit was updated to do the same (do the thing, exit early) No return value used and needed.

src/SeleniumLibrary/__init__.py
yuriverweij: The class definition only describes what the data is shaped like. The actual instances are still created per plugin inside _parse_plugin_doc(). The class definition does nothing to change the functionality .

utest/test/api/my_lib.py
yuriverweij: naming convention wants classes to have `PascalCase`  this is is only a linting fix. Using the alias at the bottom `my_lib = MyLib`  keeps the old name so nothing changes.

This is where RF kinda "breaks" python convention. RF expects classname == filename, but naming conventions ask that they are different case (pascalcase, snakecase). 

we could add a linter exception here as well, that way we can keep it as it was. `# noqa: N801`
(I would not do this exception globally, just here and the other instances where this occurs.)

utest/test/api/test_plugins.py
yuriverweij: This is more a preference and consistence thing.  `pytest.raises` gives a clearer exception than `assertRaises` and can be more flexible. In other parts of the codebase it is now also used to do the assertion on one line. I will add a comment somwhere with an example. 



(We can revert this change if you don't feel like you want this. But this is now consistently used in the entire codebase. There was a mix between `pytest.raises`  and `self.assertRaises`, with both doing the same. I opted for making it consistent and `pytest.raises` is clearer and more powerfull)

utest/test/keywords/test_click_modifier.py
yuriverweij: Yes, only less verbose and better readable (in my opionion)

utest/test/keywords/IGNOREDtest_webdrivercreator.py
yuriverweij: Agreed. But should we not delete the entire IGNOREDtest ?

utest/test/keywords/IGNOREtest_webdrivercreator_executable_path.py
yuriverweij: Yes, `RET504`, "unnecessary assignment before return" 
In this specific case I don't think the assignment adds much value, but in case of more complex returns we could argue making exceptions to the linter rule.

utest/test/keywords/test_webdrivercache.py
yuriverweij: same as above. using the `pytest.raises` consistently in the entire codebase.

src/SeleniumLibrary/keywords/webdrivertools/webdrivertools.py
yuriverweij: Fixed this.

src/SeleniumLibrary/entry/__main__.py
yuriverweij: put the `pass` back and added the linter exception globally

utest/test/api/my_lib.py
emanlove: I'm on the fence with which way to go here. My two thoughts .. (1) I am thinking for now add a linter exception. That way anyone who comes along will/should recognize this in all the other documentation the see. or (2) Let's lead and see if we can move the community forward while also providing clear understanding to the end user. Getting the filename, class name and library name all right is something I stumble on every time I write a library, noting I am writing new libraries very infrequently - less than once a year - currently.

utest/test/api/test_plugins.py
emanlove: Clearer and more powerful is a good direction to go in. Having a documented would also be helpful.

utest/test/keywords/IGNOREDtest_webdrivercreator.py
emanlove: Ahh .. IGNORED. Yeah. When I was doing selenium 4 changes so the webdriver creation there was a fair amount of changes and confussion. I think/recall at some point I gave up trying and pushed for a release. Not the best but I thought better than spinning my wheels.

Looking back it may have been #1787 which I see is actually a lot older than I thought. Maybe just leave this one? (This might depend on how strict you are. I tend to be move forward over being 100% complete)

utest/test/keywords/IGNOREtest_webdrivercreator_executable_path.py
emanlove: Good to know. Let's leave this as you have it.

utest/test/api/my_lib.py
yuriverweij: Thinking about it a bit more, I am now leaning towards adding individual linter exceptions. Since we're a RF "product", I think RF conventions are higher priority than linter conventions (as long as they don't break functionally)

utest/test/keywords/IGNOREDtest_webdrivercreator.py
yuriverweij: I think it's ok to leave this alone for now. Something we could come back to at a later time.

src/SeleniumLibrary/keywords/alert.py
yuriverweij: I think this is exactly where the linter adds value. You're somewhat forced to learn this because the linter mentions it.

utest/test/api/my_lib.py
emanlove: Agreed

emanlove (approved):

PR #1988, opened 2026-05-16
1986 set minimum selenium version

emanlove:
Setting minimum require Selenium to 4.25.0 within requirements.txt. Also bumping up the latest and previous Selenium version with the select workflow.


yuriverweij (approved):
Note: We should also update/review/refactor the CI.yml someday soon-ish. To have a more complete pipeline with more coverage when we make a release.

aaltat (commented):
Selenium 4.25 is more than year and half old. I would at least use something from 2025, perhaps not from 2026. But I would aim for something less than year old.

